PR #14, opened 2014-11-18
Support custom type (userType) decoding

clementfumey:
Hi,

I have add support for decoding custom type (or User defined type). This is probably not very clean but it's a starting point.

### DIFF
--- a/lib/encoder.js
+++ b/lib/encoder.js
@@ -18,6 +18,7 @@ var encoder = (function(){
     }
     switch(type[0]) {
       case dataTypes.custom:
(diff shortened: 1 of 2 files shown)


jorgebay:
Thanks for the pull request. The idea is to deliver UDT support with protocol v3. See the [ticket for more information](https://datastax-oss.atlassian.net/browse/NODEJS-62).

Custom type will remain encoded from / decoded to `Buffer`.

charyorde:
@clementfumey  Would this support INSERT statements?

PR #17, opened 2014-12-09
Issue 57 resolved, now collection can take keys as timestamps

guptakvgaurav:
Hi as per my observations i have tried to get rid of it. I have made some changes in hope that it may be useful for community.


jorgebay:
Thanks @guptakvgaurav , it is now implemented but I didn't merged your commit to avoid conversions for all types to Date (just string allowed), 8ebf74582c60e55b8263fb0f290cfc5fa7267a8b

PR #22, opened 2015-01-06
Enable TCP keep-alive on backend connections

jcrugzz:
As discussed in https://datastax-oss.atlassian.net/browse/NODEJS-58, we
currently see hanging backend connections to Cassandra whenever a Cassandra
node is not cleanly shutting down (OOM or kill -9).

Enabling TCP keep-alive should speed up the detection of such hanging
connections somewhat, although with the default linux settings it should still
take 10s delay (default in this patch) plus 9x 75s keepalive interval (Linux
default settings, see
http://tldp.org/HOWTO/TCP-Keepalive-HOWTO/usingkeepalive.html).

Took the excellent work done by @gwicke and just put it into a PR. @jorgebay let me know if you'd like to add anything to this, I'd just like to make sure this setting is tweakable in the next release :). In terms of defaults to extend discussion from [jira](https://datastax-oss.atlassian.net/browse/NODEJS-58), I have set it back to node's default as that has worked the best for all our use in [`http-proxy`](https://github.com/nodejitsu/node-http-proxy/blob/master/lib/http-proxy/common.js#L110-L117) which we use heavily at nodejitsu

### DIFF
--- a/lib/client-options.js
+++ b/lib/client-options.js
@@ -33,7 +33,9 @@ function defaultOptions () {
       maxConnectionsPerHost: {}
     },
     socketOptions: {
-      connectTimeout: 5000
+      connectTimeout: 5000,
+      keepAlive: true,
+      keepAliveDelay: 10000
     },
     authProvider: null,
     maxPrepared: 500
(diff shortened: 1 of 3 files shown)


jorgebay:
Looks good, I will merge it asap

PR #24, opened 2015-01-21
Added SingleNodePolicy - good for schema upgrades.

tommymessbauer:
When pushing changes to a C* cluster with multiple nodes, it is important to send the CREATE/ALTER schema changes to the same node b/c schema changes are sent with consistency ONE.  This means that sending to different nodes will result in column family id mismatch.. and that results in a failure.

https://groups.google.com/a/lists.datastax.com/forum/#!topic/nodejs-driver-user/Sghp-6N9boI
https://groups.google.com/a/lists.datastax.com/forum/#!topic/nodejs-driver-user/3YAAWgBBKfo

```
var Cassandra = require('cassandra-driver');

// this relies on base LoadBalancing class and I did not split the big load-balancing.js file.  
var SingleNodePolicy = require('cassandra-driver/lib/policies/single-node.js');

var schema_client = new Cassandra.Client({
  policies: {
    loadBalancing: new SingleNodePolicy(),
    logEmitter: function() {
        console.log(arguments);
    }
  }
});

```

If you are willing to accept this patch, then I am happy to split up lib/policies/load-balancing.js into smaller parts so that it is more modular and we can attach SingleNodePolicy to ```Cassandra.policies.loadBalancing.```


jorgebay:
I was waiting for [v1.0.3](https://github.com/datastax/nodejs-driver/releases/tag/v1.0.3) to be released to give you a full response.

On v1.0.3, there is a [new feature that allows you to configure the wait time for the schema agreement](https://datastax-oss.atlassian.net/browse/NODEJS-79). The driver queries the schema versions of the nodes. If there is a mismatch, it retries until [protocolOptions.maxSchemaAgreementWaitSeconds](http://www.datastax.com/drivers/nodejs/1.0/global.html#ClientOptions) is reached.

So now there is no need for a specific policy to handle schema changes, as the driver checks for schema versions.

tommymessbauer:
Excellent.

PR #25, opened 2015-01-23
Added "object" type to guessDataType in encoder

rjarmstrong:
Noted when doing a batch insert that there appears to be no way for an object->map to be detected, (i.e. it required you to provide a hint). There appears to be a clause missing to guess the object type. 

(probably needs more inspection to show it conforms to a map but this seems better than nothing).

### DIFF
--- a/lib/encoder.js
+++ b/lib/encoder.js
@@ -216,6 +216,9 @@ var encoder = (function(){
     else if (value === true || value === false) {
       type = dataTypes.boolean;
     }
+    else if (typeof value === 'object') {
+      type = dataTypes.map;
+    }
 
     if (type === null) {
       return null;


lib/encoder.js
```diff
@@ -216,6 +216,9 @@ var encoder = (function(){
     else if (value === true || value === false) {
       type = dataTypes.boolean;
     }
+    else if (typeof value === 'object') {
```
asilvas: `!value && typeof value === 'object'` to account for `null`

asilvas: Though technically it wont matter if it's returning if null later.

rjarmstrong:
This build is failing regardless of whether this check is in place. There is is a test for 'objects cannot be guessed' which fails regardless of whether this clause is included or not - it might be worth someone explaining why this is so?

jorgebay:
I think its not a good idea, guessing any type of object as a map could bring lots of issues. Any object instance in javascript will return typeof === 'object'. ie: `typeof Long.fromNumber(0)`, `typeof (new AnyContructor())`.

The Ecmascript type system is very limited, so if a user wants to work with most of the C* types, the recommendation is to prepare their queries. See
- http://www.datastax.com/documentation/developer/nodejs-driver/1.0/nodejs-driver/reference/nodejs2Cql3Datatypes.html?scroll=cql-to-javascript-types__encoding-data
- https://github.com/datastax/nodejs-driver#prepare-your-queries

jorgebay:
As a note, it is not possible to execute batches of prepared statements yet, but it is coming, see ticket [NODEJS-26](https://datastax-oss.atlassian.net/browse/NODEJS-26)

rjarmstrong:
Ok thanks for clearing that up - giving hints is not that elegant to use when doing a lot of queries in a batch so the prepared statements would be very much welcome.

PR #26, opened 2015-01-23
Update keepalive delay default comment to match actual default

aaron-em:
align jsdoc description of default keepalive delay with actual default in client-options.js


jorgebay:
Nice catch

PR #30, opened 2015-02-26
Fix driver misbehavior on non-uuid encoding

Shahor:
c.f. : https://datastax-oss.atlassian.net/browse/NODEJS-88

I'm not sure if the test I wrote is enough, let me know if it's not the case.

### DIFF
--- a/test/unit/encoder-tests.js
+++ b/test/unit/encoder-tests.js
@@ -206,6 +206,11 @@ describe('encoder', function () {
       encoded = typeEncoder.encode(uuid.toString(), types.dataTypes.uuid);
       assert.strictEqual(encoded.toString('hex'), uuid.getBuffer().toString('hex'));
     });
+    it('should throw when string is not an uuid', function () {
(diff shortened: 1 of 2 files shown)


test/unit/encoder-tests.js
```diff
@@ -206,6 +206,11 @@ describe('encoder', function () {
       encoded = typeEncoder.encode(uuid.toString(), types.dataTypes.uuid);
       assert.strictEqual(encoded.toString('hex'), uuid.getBuffer().toString('hex'));
     });
+    it('should throw when string is not an uuid', function () {
+      assert.throws(function () {
+        typeEncoder.encode('', types.dataTypes.uuid);
+      })
```
jorgebay: Could you limit to TypeError: `assert.throws(fn, TypeError);`

Shahor:
Yes of course!

jorgebay:
Squashed into fde640b9623224c97fd9aeea7348ae7dd044c01c and merged

PR #31, opened 2015-03-12
Setting client's connected flag to false when shutdown succeeded

razvanz:
Description on Jira: [NODEJS-94](https://datastax-oss.atlassian.net/browse/NODEJS-94)


jorgebay:
Merged, sorry for the delay!

PR #34, opened 2015-03-22
Consolidate preparation logic, repurpose RequestHandler

bauerca:
I've tried to be verbose in my commit messages, but here is another short rundown of this PR.

- Query preparation logic was previously scattered, so I thought I'd consolidate. It's now all in the Client.
- RequestHandler is more like a session. It can handle multiple requests, reusing the connection when safe, and it executes them in series, no matter when they are sent by the client (using a queue).

Carl


tolbertam:
Can one of the admins verify this patch?

jorgebay:
Closing this old pr: `RequestHandler` is thought to be the context for a request/response using 1 or multiple coordinators according to the retry mechanism and the load balancing policy.

PR #40, opened 2015-04-16
User defined types support

jorgebay:


jorgebay:
Includes: udt encoding and decoding, udt parameter hinting, lazy decoding of rows and parsing of all type names.

lib/client.js
```diff
@@ -75,11 +76,25 @@ var clientOptions = require('./client-options');
 function Client(options) {
   events.EventEmitter.call(this);
   //Unlimited amount of listeners for internal event queues by default
```
tolbertam: should this comment be moved to before line 81 since it applies to that line?

jorgebay: Yes!

lib/types/row.js
```diff
@@ -13,17 +37,24 @@ function Row(columns) {
  */
 Row.prototype.get = function (columnName) {
   if (typeof columnName === 'number') {
-    if (this.__columns && this.__columns[columnName]) {
-      columnName = this.__columns[columnName].name;
-    }
-    else {
-      throw new Error('Column not found');
-    }
+    return this._getValueFunc(columnName)();
```
tolbertam: One possible downside of this is if someone repeatedly calls get(col) on the same column it will be evaluated every time where previously it was only done once.  It doesn't seem like a common use case, but we could memoize the evaluated value so next time the function is called it can just reuse that.  The downside of this is that the storage cost is higher.  So I kind of prefer it with this change as it is more likely get(col) will only be called at most once per column.

lib/encoder.js
```diff
@@ -90,116 +145,88 @@ Encoder.prototype.encode = function (value, typeInfo) {
...
-      //noinspection JSValidateTypes
-      typeInfo = {type: typeInfo};
+      type.code = typeInfo;
     }
     else if (typeof typeInfo === 'string') {
-      typeInfo = dataTypes.getByName(typeInfo);
+      type = dataTypes.getByName(typeInfo);
     }
-    if (typeof typeInfo.type !== 'number') {
+    if (typeof typeInfo.code === 'number') {
```
tolbertam: These two if statements probably be in the else if statement above at line 157 since we know typeInfo.code !== number and type.code === number  since typeInfo is a number

tolbertam: My mistake, this is to handle all typeInfos that are not numbers, not just string.  ignore these comments :)

test/test-helper.js
```diff
@@ -336,6 +336,23 @@ var helper = {
...
+   * @param {String} testVersion Minimum version of Cassandra needed for this test
+   * @param {String} testCase Test case name
+   * @param {Function} func
+   */
+  vit: function (testVersion, testCase, func) {
+    var v = helper.getCassandraVersion().split('.').map(function (x) { return parseInt(x, 10);});
+    var currentVersion = v[0] * 10000 + v[1] * 100 + v[2];
+    v = testVersion.split('.');
+    var minimumVersion = parseFloat(v[0]) * 10000 + (parseFloat(v[1]) || 0) * 100 + (parseFloat(v[2]) || 0);
+    if (currentVersion >= minimumVersion) {
```
tolbertam: Instead of skipping the test if the version requirement is not met we could use it.skip(testCase, func).  This will show the presence of the test, but that it wasn't ran which may be useful.

jorgebay: I agree it would be nice to identify the ignored tests, similar to junit's ignore but Mocha's `it.skip()` puts the tests in a "pending state", resulting in a summary like this:
```bash
x passing
y pending
```

lib/types/row.js
jorgebay: Yes, it's a trade-off, lazy value decoding at row level allow us to better tune memory management, similar to java's `ArrayBackedRow` and C# `Row`. It will enable us to use the same network buffer to be used as row backing without the need to copy buffers as we are used to do.

jorgebay:
I updated the pull request based on [your comment](https://github.com/datastax/nodejs-driver/pull/40#discussion-diff-28622582).

About the lazy decoding of the row, I plan to dedicate some time for performance testing to provide users with real numbers when we communicate this test.

test/test-helper.js
tolbertam: Ah, yeah that doesn't seem very clear that it's being skipped then.  I agree that it is better to leave it the way you have it then.

lib/types/row.js
tolbertam: :+1: that sounds good to me, didn't occur to me at the time that the other drivers worked that way.

PR #42, opened 2015-04-29
Use prepare in trace id test for cassandra 1.2 support.

tolbertam:
Trivial change to enable prepared statements on selectNotExistent in 'should retrieve the trace id when queryTrace flag is set' (was using named parameters which is not supported in 1.2).


jorgebay:
:+1:

PR #47, opened 2015-05-19
Update to fix 'object is not a function' async race

mikeangstadt:
Updated to ensure callback is a function before it is called.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -443,7 +443,9 @@ Connection.prototype.handleResult = function (header, err, result) {
   var callback = handler.callback;
   //set the callback to null to avoid it being called when freed
   handler.callback = null;
-  callback(err, result);
+  if(callback instanceof Function){
+    callback(err, result);
+  }
 };
 
 Connection.prototype.handleNodeEvent = function (header, event) {
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -294,6 +294,10 @@ Connection.prototype.changeKeyspace = function (keyspace, callback) {
   this.toBeKeyspace = keyspace;
   var query = util.format('USE "%s"', keyspace);
   var self = this;
+  if(!this.writeQueue)
+  {
+    this.chainEvents();
+  }
   this.sendStream(
     new requests.QueryRequest(query, null, null),
     null,


jorgebay:
Do you have a way to reproduce it?

We have been hitting this issue on our integration servers from time to time while running [this test that is designed to try to create race conditions](https://github.com/datastax/nodejs-driver/blob/3748dc1b5b40012a1ce91e4bf72c0cc95d2a180f/test/integration/short/client-execute-prepared-tests.js#L58), but we can not accurately find the underlying cause...

mikeangstadt:
It occurs when you load test reads across multiple keyspaces - due to the round robin rotation through connections hitting multiple keyspaces.  If the code cycles quickly enough you'll round-trip since the .next connection is used if the keyspace is different then the last connections.

eg, fill a keyspace 'foo' and column family 'blips' with 10,000 trivial records
fill a keyspace 'fum' and column family 'bars' with 10,000 trivial records which contain references to records in 'foo'

setup an async parallel call to get a list of 1,000 bars by id from fum and hydrate their blips from foo.

* note: you'll eventually get this error.

jorgebay:
Ok, thanks for the tip, we will try that.

As a note, you should use 2 different client instances in those situations, one per keyspace.
Either that or you should "fully qualify" your table names, ie: `SELECT x, w, z FROM ks1.tbl1`.

mikeangstadt:
Thanks for the tip, hopefully since this is merely a defensive check to alleviate the issue it'll be an easy call to merge - cheers!

mikeangstadt:
ping on this @jorgebay - mind addressing this as it's a non-breaking defensive check?

jorgebay:
We are still trying to determine the underlying cause, we don't want to risk a response not being called back. I will let u know as soon as we have something.

mikeangstadt:
Understood but the only risk to a response not being called back is if the
callback isn't an executable function - which would fail regardless.

On Fri, May 22, 2015 at 2:54 AM, Jorge Bay Gondra <notifications@github.com>
wrote:

> We are still trying to determine the underlying cause, we don't want to
> risk a response not being called back. I will let u know as soon as we have
> something.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/datastax/nodejs-driver/pull/47#issuecomment-104556245>
> .
>



-- 
Mike Angstadt
www.mikeangstadt.com
@mike_angstadt <http://www.twitter.com/mike_angstadt> | LinkedIn
<http://www.linkedin.com/in/mikeangstadt>
610-914-4409

*"The noblest pleasure is the joy of understanding" - Leonardo Di Vinci*

jorgebay:
We merged #55 that fixes this issue. Even though your pull request did solved the bug, we decided to find the root cause and fix that.

PR #50, opened 2015-05-21
Support C* smallint and tinyint

jorgebay:


tolbertam:
Looks good to me! :+1:  I added some additional unit validation for encoding of smallint and tinyint.

lib/metadata.js
```diff
@@ -737,24 +738,46 @@ TableMetadata.prototype.build = function (tableRow, columnRows) {
     };
     this.columns.push(c);
     columnsKeyed[c.name] = c;
+    switch (row['type']) {
```
tolbertam: There seems to be a slight regression with 'isCompact' here.  'tbl6' in metadata-tests.js is not being listed as compact when it should (Metadata #getTable() should retrieve the metadata of a compact storaged table).  Looking into it.

lib/metadata.js
```diff
@@ -770,7 +793,10 @@ TableMetadata.prototype.build = function (tableRow, columnRows) {
       this.clusteringOrder.push(c.type.options.reversed ? 'DESC' : 'ASC');
     }
   }
-  if (!this.isCompact) {
+  if (typeof tableRow['is_dense'] !== 'undefined') {
```
tolbertam: Ah, I see the problem now, the code below says:  

>  also, compact pk, ck and val appear as is_dense false

So I think if we make this else if(!this.isCompact) that should fix it.  What do you think?

lib/metadata.js
```diff
@@ -770,7 +793,10 @@ TableMetadata.prototype.build = function (tableRow, columnRows) {
...
+    this.isCompact = tableRow['is_dense'];
+  }
+  else {
```
tolbertam: (Updated to reflect correct line)
Ah, I see the problem now, the code below says:  

>  also, compact pk, ck and val appear as is_dense false

So I think if we make this if(!this.isCompact) that should fix it.  What do you think?

jorgebay:
Fixed the `isCompact` value, added a new unit test using compact storage without clustering keys (for C* 2.0) and rebased against #57.

lib/metadata.js
```diff
@@ -770,14 +793,15 @@ TableMetadata.prototype.build = function (tableRow, columnRows) {
...
+  this.isCompact = !!tableRow['is_dense'];
   if (!this.isCompact) {
-    //is_dense column does not exist in previous versions of Cassandra
-    //also, compact pk, ck and val appear as is_dense false
-    // clusteringKeys != comparator types - 1
-    // or not composite (comparator)
-    this.isCompact =
-      this.clusteringKeys.length !== (types.length - 1) ||
-      tableRow['comparator'].indexOf(_compositeTypeName) !== 0;
+      //is_dense column does not exist in previous versions of Cassandra
```
tolbertam: is the extra two spaces here inadvertent or intentional?

lib/metadata.js
tolbertam: This has been fixed, thanks!

lib/metadata.js
jorgebay: not intentional...

jorgebay: Fixed spacing and merged

PR #51, opened 2015-05-21
Support for new C*2.2 date and time types NODEJS-118

jorgebay:


lib/types/local-date.js
```diff
@@ -0,0 +1,200 @@
...
+   * @name day
+   * @type Number
+   * @memberof module:types~LocalDate#
+   */
+  Object.defineProperties(this, {
+    'year': { enumerable: true, get: function () {
+      return self.value.getUTCFullYear();
+    }},
+    'month': { enumerable: true, get: function () {
+      return self.value.getMonth() + 1;
```
tolbertam: I think these should get getUTCMonth and getUTCDay as given Date.UTC is used to create the date.  The tests in encoder-tests.js fail if local time is offset enough from UTC.

lib/types/local-date.js
```diff
@@ -0,0 +1,200 @@
...
+    'day': { enumerable: true, get: function () {
+      return self.value.getDate();
+    }}
+  });
+}
+
+/**
+ * Creates a new instance of LocalDate using the current year, month and day from the system clock in the default time-zone.
+ */
+LocalDate.now = function () {
```
tolbertam: I think we should either assume that UTC is always used, or always use the Local time definition for defining date, otherwise this can be confusing.

lib/types/local-date.js
```diff
@@ -0,0 +1,200 @@
...
+ * @param {String} value
+ */
+LocalDate.fromString = function (value) {
+  var multiplier = 1;
+  if (value[0] === '-') {
+    value = value.substring(1);
+    multiplier = -1;
+  }
+  var parts = value.split('-');
+  return new LocalDate(multiplier * parseInt(parts[0], 10), parseInt(parts[1], 10), parseInt(parts[2], 10));
```
tolbertam: Do you think it would be worth making the fromString implementations of LocalDate and LocalTime have more stricter validation of integer parsing?  parseInt will simply ignore non-numerical values, so you can create some unexpected behavior if non-complete numeric values are provided.

lib/types/local-date.js
```diff
@@ -0,0 +1,200 @@
...
+ * </p>
+ * <p>
+ *   This class does not store or represent a time or time-zone. Instead, it is a description of the date, as used for birthdays. It cannot represent an instant on the time-line without additional information such as an offset or time-zone.
+ * </p>
+ * @param {Number} year The year.
+ * @param {Number} month Between 1 and 12 inclusive.
+ * @param {Number} day Between 1 and the number of days in the given month of the given year.
+ * @constructor
+ */
+function LocalDate(year, month, day) {
```
tolbertam: Since LocalDate and LocalTime are currently timezone independent, and the concept of 'local' implies local timezone (at least to me), should we consider renaming these?

lib/types/local-date.js
tolbertam: We could also consider getting rid of the now, utcNow, and fromDate methods altogether since it is easy enough to provide year, month, day itself.

lib/types/local-time.js
```diff
@@ -0,0 +1,306 @@
...
+ * @param {Number} milliseconds A Number from 0 to 86,399,999.
+ * @param {Number} [nanoseconds] A Number from 0 to 999,999, representing the time nanosecond portion.
+ * If not provided, it will be a monotonically increasing value.
+ * @returns {LocalTime}
+ */
+LocalTime.fromMilliseconds = function (milliseconds, nanoseconds) {
+  if (typeof nanoseconds !== 'number') {
+    //generate
+    nanoseconds = nanoGenerator++;
+    if (nanoGenerator === 1000000) {
```
tolbertam: I think if we can't guarantee a monotonic value we should just use '0' instead if nanoseconds if not specified.

test/unit/basic-tests.js
```diff
@@ -146,6 +146,131 @@ describe('types', function () {
...
+        assert.ok(decoded.equals(val));
+        assert.strictEqual(val.toString(), decoded.toString());
+      });
+    });
+    describe('#hour #minute #second #nanosecond', function () {
+      it('should get the correct parts', function () {
+        values.forEach(function (item) {
+          var val = new LocalTime(Long.fromString(item[0]));
+          var parts = item[2];
+          console.log(item[1]);
```
tolbertam: extra console.log not needed.

lib/types/local-date.js
jorgebay: Nice catch: I will fix it

lib/types/local-time.js
jorgebay: Agree! Also, now that I think about it, there is no point in generating it for the user (times are not used for timestamps or anything like it), default to `0` would make more sense.

lib/types/local-date.js
jorgebay: The names are based on [Java's LocalDate and LocalTime] (https://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html). 

I think it is meant to be local from any user perspective: "She was born on October 1st, 1980" (no matter where she lived or where she lives now, her birthday is going to remain the same).

lib/types/local-date.js
jorgebay: From the user perspective, `LocalDate` should always use local clock values (and does use local clock). The only case where utc is used / referenced is `utcNow()`, but it can be considered as a utility method. I took the concepts from [LocalDate](http://joda-time.sourceforge.net/apidocs/org/joda/time/LocalDate.html#LocalDate()).

Internally it uses UTC values to use the built-in calendar functionality, but the user doesn't have to know / understand that.

jorgebay:
Added validation for generated date (via `fromString()` or any other).
Fixed `month` and `day`.
Fixed representation for first century dates.
Removed nanoseconds generation.
Throw `TypeError` for date and time encoder errors.

lib/types/local-date.js
tolbertam: ah excellent.  Thanks for the explanation, since it is being patterned of an already existing implementation I would say this is fine then.  I was initially thrown off by Local == 'Local Timezone' but your explanation has me on the right page now.

lib/types/local-date.js
tolbertam: Ah yes, now I'm in sync.  I was getting wrapped up in the internal implementation, but it turns out to not matter if you store it in UTC and look at it in UTC.   If you look at LocalDate.value with your local timezone, there's a chance it could appear off, but I think that is a consideration one needs to take into account when using value.

tolbertam:
Changes look great! :+1:

PR #52, opened 2015-05-21
NODEJS-128 Use system.peers instead of local in getPeersSchemaVersions

tolbertam:
[NODEJS-128](https://datastax-oss.atlassian.net/browse/NODEJS-128)


jorgebay:
Moved to public branch and opened #53

PR #53, opened 2015-05-22
Fix wait for schema agreement NODEJS-128

jorgebay:


jorgebay:
I've added an additional test to try to create a race condition with schema changes.

tolbertam:
This has been working great.  No issues over a day of constant test automation runs.  I think this is safe to merge at your convenience! :+1:

PR #54, opened 2015-05-22
Unset encoding NODEJS-121

jorgebay:
- Includes unit and integration tests
- Exposed cassandra.types.unset field to set a parameter to unset
- If a flag is set, the user can use undefined as unset.


lib/client.js
```diff
@@ -60,6 +60,16 @@ var clientOptions = require('./client-options');
  *  Setting it to false will cause less overhead and the reference of the network buffer to be maintained until the row / result set are de-referenced.
  *  Default: true.
  * </p>
+ * @property {Boolean} encoding.useUndefinedAsUnset Valid for Cassandra 2.2 and above. Determines that, if a parameter is set to
```
tolbertam: Would it be worth validating on initialization of Client that if protocolVersion < 4 is used and useUndefinedAsUnset is specified as true that we fail fast?  We would probably want to do it when we detect the max allowed protocol version as well.  I think this could be useful, but I can see it not being necessary.

tolbertam:
Looks great to me! :+1:

lib/client.js
jorgebay: It would be nice to have but to do that we would have to validate at client level after connect, another validation method. Also we would still have to validate at encoding level if the user is using `unset` with a lower protocol version.

tolbertam: Ah yes, good point.  I agree that we don't need validation on client initialization then.  The documentation is explicit and if a user encounters the error it is clear what the problem is.

jorgebay:
Merged into 2.2

PR #58, opened 2015-05-27
Convert setImmediate to async.setImmediate

jkingsman:
Using async for this will let it fall through to the system setImmediate when it's defined, and shim in the functionality when it's not (falls back to `nextTick` (mainly for node <0.9), which falls back to `setInterval(..., 0)` when all else fails).

I know this isn't a super common issue, but it can make things nicer for older node versions and browserify, and being more supportive/having fallbacks seems like a good principle (plus async is already included, and it's a transparent fallthrough for 99% of cases, meaning there's minimal performance hit).

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -374,7 +374,7 @@ Client.prototype.shutdown = function (callback) {
  */
 Client.prototype.waitForSchemaAgreement = function (connection, callback) {
   if (this.hosts.length === 1) {
-    return setImmediate(callback);
+    return async.setImmediate(callback);
   }
   var self = this;
   var start = new Date();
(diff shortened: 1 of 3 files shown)


jorgebay:
Thanks for the pull request. It makes sense for node versions under 0.10 but [the Node.js driver supports node 0.10 or above](https://github.com/datastax/nodejs-driver/blob/master/package.json#L42).

We use node.js 0.10 api extensively (specially [streams2](http://blog.nodejs.org/2012/12/20/streams2/)) and it is been out for more than 2 years, so I think it is not an option.

jkingsman:
Okay, that makes sense. I bumped into the issue while browserifying the driver (no browser but IE has setImmediate), and the async fallback handled it well, but I know that's a practically non-existent usage case.

PR #59, opened 2015-05-27
Custom payload encoding and decoding

jorgebay:


lib/requests.js
```diff
@@ -149,6 +158,12 @@ QueryRequest.prototype.write = function (encoder) {
   //v3: <query>
   //      <consistency><flags>[<n>[name_1]<value_1>...[name_n]<value_n>][<result_page_size>][<paging_state>][<serial_consistency>][<timestamp>]
   var frameWriter = new FrameWriter(types.opcodes.query);
+  var headerFlags = this.options.traceQuery ? types.frameFlags.tracing : 0;
```
tolbertam: One thing I noticed that both the python and java drivers supports is that it uses Custom Payload returned from an associated Prepared Statement Response as the Custom Payload set on requests using that Prepared Statement if the User did not provide their own payload (see the algorithm defined [here](https://github.com/datastax/java-driver/blob/f6b5809d880dba65257298963c95da8387e9b9c7/features/custom_payloads/README.md#custom-payloads-and-prepared-statements)).

The way I think this could be supported here is whenever emitting a prepared statement you would also emit the customPayload with the id and metadata and pass it on to ExecuteRequest from Client._executeAsPrepared.

Do you think this is needed for the nodejs-driver?

tolbertam: Another possible thing I noticed is that you currently can't get at the prepared statement custom payload if you wanted to act on it in some way in the client.  I'm not sure if there is a use case for wanting to look at that data, but thought I'd point that out.

jorgebay: I was wondering the same the thing when I implemented it. There is no way to deal with prepared statement data in the Node.js driver, it is handled by the driver and abstracted away.

I think it would make sense in case we add a method to `prepareAll()` or something like that, that will enable users to prepare statements before the first execution, manually (most likely on app start), but not for now.

tolbertam: Makes sense :+1:.  I think supporting custom payloads on the query itself is sufficient for now.  I logged [NODEJS-134](https://datastax-oss.atlassian.net/browse/NODEJS-134) for future support.

tolbertam:
Looks good to me :+1:

PR #60, opened 2015-05-27
Parse client address in query trace

jorgebay:


tolbertam:
Looks great to me!  :+1:

PR #63, opened 2015-05-29
Use prepared meta partition keys for routing key

jorgebay:
Refactored `setQueryOptions()` to include the partition keys from the prepared metadata.

I've added a new `long` test to check that there are no networks hops when using TokenAwarePolicy using query trace events.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -719,8 +719,8 @@ Client.prototype._setQueryOptions = function (options, params, meta, callback) {
           //The data is not there, maybe it is being recreated
           return next();
         }
-        options.routingIndexes = tableInfo.partitionKeys.map(function (key) {
-          return meta.columnsByName[key];
+        options.routingIndexes = tableInfo.partitionKeys.map(function (c) {
+          return meta.columnsByName[c.name];
         });
         //Skip parsing metadata next time
         meta.partitionKeys = options.routingIndexes;
--- a/lib/client.js
+++ b/lib/client.js
@@ -690,44 +690,80 @@ Client.prototype._getEncoder = function () {
  * @private
  */
 Client.prototype._setQueryOptions = function (options, params, meta, callback) {
-  var paramsInfo;
-  var self = this;
-  function adaptPageStateAndRouting(keys, params) {
-    if (typeof options.pageState === 'string') {
-      //pageState can be a hex string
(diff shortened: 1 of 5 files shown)


lib/client.js
```diff
@@ -689,44 +689,80 @@ Client.prototype._getEncoder = function () {
...
+        //it is filled by the user
+        //or it is not prepared
+        return next();
+      }
+      if (util.isArray(meta.partitionKeys)) {
+        //the partition keys are provided as part of the metadata
+        options.routingIndexes = meta.partitionKeys;
+        return next();
+      }
+      self.metadata.getTable(meta.keyspace, meta.table, function (err, tableInfo) {
```
tolbertam: Woah this is pretty cool, is this new (from what I can tell it is)?   This would only apply to proto version < 4 right? ( since meta.partitionKeys will be present).

test/integration/long/load-balancing-tests.js
```diff
@@ -218,4 +217,44 @@ describe('TokenAwarePolicy', function () {
...
+        async.timesSeries(10, function (n, timesNext) {
+          var id = (n % 10) + 1;
+          var query = util.format('INSERT INTO %s (id, name) VALUES (?, ?)', table);
+          client.execute(query, [id, id], { traceQuery: true, prepare: true}, function (err, result) {
+            assert.ifError(err);
+            var coordinator = result.info.queriedHost;
+            var traceId = result.info.traceId;
+            client.metadata.getTrace(traceId, function (err, trace) {
+              assert.ifError(err);
+              trace.events.forEach(function (event) {
```
tolbertam: If I understand this correctly this is just making sure that all events share the same host, thus proving that the data was retrievable by the coordinator therefore the correct routing was used?

tolbertam:
Looks good to me :+1:

test/integration/long/load-balancing-tests.js
jorgebay: Correct, as replication factor is 1, there is only 1 valid coordinator per partition key.
When there is a hop between the coordinator and the partition, it is visible in the event sources.

lib/client.js
jorgebay: When using protocol version 4+, index of parameters that form the partition key is loaded into `meta.partitionKeys`.
For lower protocol versions, the table metadata is used.

PR #64, opened 2015-06-02
User-defined functions metadata support

jorgebay:


lib/metadata.js
```diff
@@ -898,4 +1099,115 @@ function adaptKeyTypes(typesString, encoder) {
...
 
+/**
+ * Creates a new SchemaFunction.
+ * @class
+ * @classdesc Describes a CQL function.
+ * @param {Encoder} encoder
+ * @param {Row} row
+ * @constructor
+ */
+function SchemaFunction(encoder, row) {
```
tolbertam: It could be helpful to also have a field for the associated keyspace for both SchemaFunction and SchemaAggregate as that may be required depending on whether or not the Client has a set keyspace.

tolbertam:
Looks good to me!  :+1:

lib/metadata.js
jorgebay: Sounds good, I will add it.

jorgebay:
Added the `keyspaceName` by rebase editing and merged.

PR #65, opened 2015-06-02
Update LocalDate to handle dates > ±100000000 days from epoch

tolbertam:
* Changes LocalDate constructor to interpret 'year' as days from epoch instead of millis from epoch.
* LocalDate now keeps internal state 'value' as number representing days from epoch.
* Removed year, month, day properties.  Added 'date' property which returns date representation of LocalDate (or Invalid Date if not possible).
* Updated toString implementation to render as yyyy-mm-dd if can be represented as date, otherwise days from epoch.
* Updated tests to test for different dates.

This currently fails for 0000-01-01 as it is off by a day (thus tests will fail).  Will add a fix for that.
Will also improve documentation, validate 'date' object in tests.

### DIFF
--- a/lib/types/local-date.js
+++ b/lib/types/local-date.js
@@ -12,10 +12,6 @@ var millisecondsPerDay = 86400000;
  */
 var dateCenter = Math.pow(2,31);
 /**
- * @private
- */
-var twentiethCentury = new Date('1899-12-31') - new Date('0000-01-01');
-/**
  *
  * Creates a new instance of LocalDate.
  * @class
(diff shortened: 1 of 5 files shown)


tolbertam:
Updated so it now properly handles 0000-01-01.  In addition, fromString will now also support parsing since epoch days format (i.e. passing in -1 returns a LocalDate representing 12-31-1969).

jorgebay:
I think it makes sense to have `year`, `month` and `day` properties, similar to Java 8 `LocalDate`, as originally implemented, and have `toDate()` as a conversion method, if needed.

I think we should fix from the original implementation:
- Fix the `toString()` and `fromString()` methods to support dates the outside the Ecmascript valid Date values, returning the value in days (this way it will used as valid object maps keys and as json from and to cassandra).
- Explicitly return [NaN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN) as year / month / day property values for dates that are outside the range.

Document the whole thing.

lib/types/local-date.js
```diff
@@ -26,58 +22,81 @@ var twentiethCentury = new Date('1899-12-31') - new Date('0000-01-01');
...
    * A number between 1 and the number of days in the given month of the given year (28, 29, 30, 31).
+   * May return NaN if cannot be represented as a Date.
    * @name day
    * @type Number
    * @memberof module:types~LocalDate#
    */
   Object.defineProperties(this, {
     'year': { enumerable: true, get: function () {
-      return self.value.getUTCFullYear();
+      return isNaN(self.date.getTime()) ? NaN : self.date.getUTCFullYear();
```
jorgebay: I think this is unnecessary, `self.value.getUTCFullYear()` already returns `NaN` if `getTime()` is `NaN`.

lib/types/local-date.js
```diff
@@ -110,17 +129,27 @@ LocalDate.fromDate = function (date) {
...
+    if (value[0] === '-') {
+      value = value.substring(1);
+      multiplier = -1;
+    }
+    var parts = value.split('-');
+    return new LocalDate(multiplier * parseInt(parts[0], 10), parseInt(parts[1], 10), parseInt(parts[2], 10));
+  } else if(value.match(/^-?\d+$/)) {
+    // Parse as days since epoch.
+    //noinspection JSCheckFunctionSignatures
+    return new LocalDate(parseInt(value));
```
jorgebay: The result will be the same if you remove this `else if` block.
`parseInt(value)` will yield NaN if it can not be parsed, causing the `LocalDate` constructor to throw an error (as expected).

tolbertam: The main reason for that check is if a string that parses to an Int is passed in that we may not want to consider legal.  For example in basic-tests.js 1880-1 would parse to '1880' in parseInt, do we want that behavior?

lib/types/local-date.js
tolbertam: Ah, you are right!  I'll change it back.

lib/types/local-date.js
```diff
@@ -26,58 +22,81 @@ var twentiethCentury = new Date('1899-12-31') - new Date('0000-01-01');
...
   }
   else if (typeof month === 'undefined' && typeof day === 'undefined') {
     if (typeof year === 'number') {
-      //in milliseconds since unix epoch
-      this.value = new Date(year);
+      //in days since epoch.
+      if(year < -2147483648 || year > 2147483647) {
+        throw new Error('You must provide a valid value for days since epoch (-2147483648 <= value <= 2147483647).')
+      }
+      this.value = year;
```
jorgebay: Can we create the `value` field only if date is `NaN`? We could reduce an allocation and we can access the `date.getTime()` to get the same result.

This applies to the previous `if`block also.

lib/types/local-date.js
jorgebay: Makes sense :+1:

lib/types/local-date.js
tolbertam: My only concern there is if a user wants to use value directly rather than Date.  Though that seems less likely, what do you think?

lib/types/local-date.js
jorgebay: A minor thing, we should specify the radix in `parseInt(value, 10)`, otherwise it will try to guess it (thanks ES!).

tolbertam: Ah, ES can be full of surprises!  I'll make it explicit.

lib/types/local-date.js
jorgebay: I would try to avoid exposing it for now and reduce the allocation whenever possible.
We could even make it not enumerable initially:

```javascript
Object.defineProperty(this, 'value', { enumerable: false, value: null });
```

But as long it is undocumented, it is fine by me.

tolbertam: Sounds good to me :+1:.  In that case it'll only ever be used in the toString() implementation which makes sense.

tolbertam: While making this change one thing I noticed is that this adds complexity to both compareTo and toBuffer in that we'll need to convert the date if it is appropriate, otherwise use value.  Is that ok?  I'm fine implementing it that way but just wanted to verify that was ok.

jorgebay: For me is ok, we will be optimizing for the most common use case (dates within ES range) and fallback for others, without the extra allocation for all.

tolbertam: Sounds good.  It ended up being more simple than I had originally thought, pushing up a change shortly for this.

jorgebay:
The IDE warnings that I get are related to `isNaN()` calls, according to specification `isNaN(date)` is a valid call but according to WebStorm Ecmascript metadata (go to declaration), `isNaN` expects a Number.

You can add 2 pending `//noinspection JSCheckFunctionSignatures` or call isNaN over the `getTime()` value: `isNaN(this.date.getTime())`, as this.date is always defined.

Besides that, it looks good to me: squash the commits so I can merge them.

tolbertam:
Sounds good.  I'll correct the isNaN checks around .getTime() and squash, thanks!

tolbertam:
corrected and squashed.

PR #67, opened 2015-06-03
Use retry policy defined in the query options NODEJS-139

jorgebay:


tolbertam:
looks good to me! :+1:

PR #68, opened 2015-06-05
Fix: Use pool connection length to check before switching to a keyspace

jorgebay:
After connecting the control connection, use pool connection length to check before switching to a keyspace.


asilvas:
+1

jorgebay:
Thanks @asilvas for the feedback! 

I've been created a unit test to cover it (failed and now passes), but we haven't been able to reproduce it in an integration test. If you can figure a way we can repro, let us know.

PR #69, opened 2015-06-05
Custom payload using bytes instead of short bytes NODEJS-123

jorgebay:
For upcoming Cassandra 2.2 beta or rc version, for the [CASSANDRA-9515](https://issues.apache.org/jira/browse/CASSANDRA-9515) fix.

To run the test, include the env variable: `TEST_CASSANDRA_DIR`, for example:
```bash
export TEST_CASSANDRA_DIR=/path/to/cassandra-2.2/src/
```

We should not merge this until the C* version is released AND ci uses the latest C* 2.2.

### DIFF
--- a/lib/readers.js
+++ b/lib/readers.js
@@ -163,7 +163,7 @@ FrameReader.prototype.readBytesMap = function () {
   }
   var map = {};
   for (var i = 0; i < length; i++) {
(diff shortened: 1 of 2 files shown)


tolbertam:
looks good to me :+1: on merging when the RC is released

PR #71, opened 2015-06-15
Defer removal of error listener until connection is ready.  NODEJS-146

tolbertam:
This change defers the removal of any error handlers until the startup request has be successfully processed and the connection is ready.  Should ensure that there is always an 'error' handler on a connection, which prevents an ```Unhandled 'error' event``` error.   

Included test that reproduces the error condition if ```this.netClient.removeAllListeners('error');``` is added back to line 70.

[NODEJS-146](https://datastax-oss.atlassian.net/browse/NODEJS-146)

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -67,7 +67,6 @@ Connection.prototype.log = utils.log;
  */
 Connection.prototype.bindSocketListeners = function() {
   //Remove listeners that were used for connecting
-  this.netClient.removeAllListeners('error');
   this.netClient.removeAllListeners('connect');
   this.netClient.removeAllListeners('timeout');
   var self = this;
@@ -193,6 +192,8 @@ Connection.prototype.connectionReady = function (callback) {
   this.emit('connected');
   this.connected = true;
   this.connecting = false;
+  // Remove existing error handlers as the connection is now ready.
+  this.netClient.removeAllListeners('error');
   this.netClient.on('error', this.handleSocketError.bind(this));
   callback();
 };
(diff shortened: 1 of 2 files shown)


jcrugzz:
This LGTM :+1:. I've definitely been bitten by unhandled error messages.

jorgebay:
+1 Makes sense. I've introduced a minor change, remove all the listeners at once as adding and removing listeners is a sync op and there is only 1 thread.

If its OK for you, I will squash and merge to master and 2.2.

tolbertam:
@jorgebay Change looks good to me!  :+1:

jcrugzz:
@jorgebay could we get a patch release with this fix ASAP? Currently certain failure cases are crashing the process with ambiguous unhandlable errors and we are trying to get this driver in use in production. Would really appreciate it!

jorgebay:
We are running endurance tests against master, we will release is it as soon they are finished.

Are you sure you are hitting this bug? This bug happens when a node continues to allow incoming connections but it is unresponsive (after connecting but before the startup messages).

Would you like to post a JIRA with more info (trace / topology / etc) for us to take a look?

jcrugzz:
@jorgebay i think i realized it was another bug specifically. Was just getting an arbitrary stream error on the current released version and then using the master branch I was able to get the real error which was an incorrect value being passed for a `TimeUuid` since the types changed the semantics of how we execute `where` based queries with `>, >=, <, <=` on time series columns.

jorgebay:
@jcrugzz You can check [the tickets included per release on JIRA](https://datastax-oss.atlassian.net/projects/NODEJS?selectedItem=com.atlassian.jira.jira-projects-plugin:release-page&status=all).
Before releasing, we run duration and endurance tests that take more than a day to complete.

PR #72, opened 2015-06-26
Option to warmup pool on connect

jorgebay:
- Provide option to open all connections at startup NODEJS-138
- connect() callbacks in error after shutdown NODEJS-111


jorgebay:
Using 2.1 as base branch: #74

PR #73, opened 2015-06-26
Whitelist load balancing policy

jorgebay:


jorgebay:
Using 2.1 as base branch: #75

PR #75, opened 2015-06-29
Whitelist load balancing policy (2.1)

jorgebay:


tolbertam:
looks good to me :+1:

PR #76, opened 2015-06-29
Expose events host up, down, add, remove

jorgebay:
It is based on #74 branch because these 2 pull requests affect the Client connect() method.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -140,6 +140,35 @@ function Client(options) {
 util.inherits(Client, events.EventEmitter);
 
 /**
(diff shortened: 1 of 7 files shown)


tolbertam:
Looks good to me! :+1:. I added a test for NODEJS-138 and also adjusted one of the tests to account for inconsistent up/down notification behavior in 1.2.x (similar exceptions in [java-driver](https://github.com/datastax/java-driver/blob/2.0/driver-core/src/test/java/com/datastax/driver/core/StateListenerTest.java))

PR #77, opened 2015-07-01
Timer-based read timeout

jorgebay:
Moves to the next host on timeout based on a setting
Defuncts connection if a threshold is passed, based on a setting

NODEJS-114

### DIFF
--- a/lib/client-options.js
+++ b/lib/client-options.js
@@ -22,7 +22,8 @@ function defaultOptions () {
     queryOptions: {
       consistency: types.consistencies.one,
       fetchSize: 5000,
-      prepare: false
+      prepare: false,
+      retryOnTimeout: true
     },
     protocolOptions: {
       port: 9042,
@@ -34,8 +35,10 @@ function defaultOptions () {
     },
     socketOptions: {
       connectTimeout: 5000,
+      defunctReadTimeoutThreshold: 64,
       keepAlive: true,
       keepAliveDelay: 0,
+      readTimeout: 0,
       tcpNoDelay: true
     },
     authProvider: null,
@@ -87,7 +90,13 @@ function extend(baseOptions, userOptions) {
     throw new TypeError('queryOptions not defined in options');
   }
   if (!options.protocolOptions) {
-    throw new TypeError('protocolOptions');
+    throw new TypeError('protocolOptions not defined in options');
+  }
+  if (!options.socketOptions) {
+    throw new TypeError('socketOptions not defined in options');
+  }
+  if (typeof options.socketOptions.readTimeout !== 'number') {
+    throw new TypeError('socketOptions.readTimeout must be a Number');
   }
   if (!options.logEmitter) {
     options.logEmitter = function () {};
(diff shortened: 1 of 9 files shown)


lib/client-options.js
```diff
@@ -34,8 +35,10 @@ function defaultOptions () {
     },
     socketOptions: {
       connectTimeout: 5000,
+      defunctReadTimeoutThreshold: 64,
       keepAlive: true,
       keepAliveDelay: 0,
+      readTimeout: 0,
```
tolbertam: Is the plan that in a future release this would be defaulted to a set value?

lib/client-options.js
```diff
@@ -34,8 +35,10 @@ function defaultOptions () {
     },
     socketOptions: {
       connectTimeout: 5000,
+      defunctReadTimeoutThreshold: 64,
```
tolbertam: I really like this capability.   What do you think about having some kind of 'heartbeat' timeout threshold?   In the java-driver if any heartbeat options message times out, [the connection is defuncted](https://github.com/datastax/java-driver/blob/2.1.6/driver-core/src/main/java/com/datastax/driver/core/Connection.java#L1094-L1102).  Part of me thinks this may not be necessary since the nodejs driver shares the control connection in the pool, so each connection should always stay pretty active and hit the current defunctReadTimeoutThreshold pretty quick.

lib/client-options.js
```diff
@@ -87,7 +90,13 @@ function extend(baseOptions, userOptions) {
     throw new TypeError('queryOptions not defined in options');
   }
   if (!options.protocolOptions) {
-    throw new TypeError('protocolOptions');
+    throw new TypeError('protocolOptions not defined in options');
+  }
+  if (!options.socketOptions) {
```
tolbertam: Not really any likelihood of backwards compatibility issues from this since values should be present from defaults right?

lib/connection.js
```diff
@@ -520,17 +527,62 @@ Connection.prototype.handleRow = function (header, row, meta, rowLength, traceId
...
+  var self = this;
+  //clear callback closures
+  handler.callback = function () {
+    //if replies, remove from timedOutQueries
+    self.timedOutHandlers--;
+  };
+  if (handler.options && handler.options.rowCallback) {
+    handler.options.rowCallback = function noop() {};
   }
+  var message = util.format('The host %s did not replied before timeout %d ms', this.endPoint, this.options.socketOptions.readTimeout);
```
tolbertam: should change to 'did not reply'

lib/client-options.js
jorgebay: I think we should default it to 12000, if possible in the next minor version. I created a ticket for it [NODEJS-152](https://datastax-oss.atlassian.net/browse/NODEJS-152).

lib/client-options.js
jorgebay: If not specified, it sets the defaults, this is just a validation for our default settings.

tolbertam:
looks good to me! :+1:

PR #78, opened 2015-07-02
NODEJS-151 Keep separate index for each plan in Round Robin policies.

tolbertam:
An attempt to prevent separate query plans from affecting each other by maintaining a separate index for each generated query plan based off of the policy-based index.   This applies to both DCAwareRoundRobinPolicy and RoundRobinPolicy.

### DIFF
--- a/lib/policies/load-balancing.js
+++ b/lib/policies/load-balancing.js
@@ -175,7 +175,7 @@ DCAwareRoundRobinPolicy.prototype.newQueryPlan = function (keyspace, queryOption
   if (self.index >= utils.maxInt) {
     self.index = 0;
   }
-  var planRemoteIndex = undefined;
+  var planRemoteIndex;
 
   callback(null, {
     next: function () {
--- a/lib/policies/load-balancing.js
+++ b/lib/policies/load-balancing.js
@@ -70,17 +70,19 @@ RoundRobinPolicy.prototype.newQueryPlan = function (keyspace, queryOptions, call
   var hosts = this.hosts.values();
   var self = this;
   var counter = 0;
+
+  var planIndex = self.index % hosts.length;
+  self.index += 1;
+  if (self.index >= utils.maxInt) {
+    self.index = 0;
+  }
+
   callback(null, {
     next: function () {
       if (++counter > hosts.length) {
         return {done: true};
       }
-      self.index += 1;
-      //overflow protection
-      if (self.index >= utils.maxInt) {
-        self.index = 0;
-      }
-      return {value: hosts[self.index % hosts.length], done: false};
+      return {value: hosts[planIndex++ % hosts.length], done: false};
     }
   });
 };
@@ -167,6 +169,14 @@ DCAwareRoundRobinPolicy.prototype.newQueryPlan = function (keyspace, queryOption
   var self = this;
   var counter = 0;
   var remoteCounter = 0;
+
+  var planIndex = self.index % hosts.length;
+  self.index += 1;
+  if (self.index >= utils.maxInt) {
+    self.index = 0;
+  }
+  var planRemoteIndex = undefined;
+
   callback(null, {
     next: function () {
       var local = this.nextLocal();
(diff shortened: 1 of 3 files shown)


tolbertam:
Updated `'should yield host in a round robin manner when consuming'` to add random variance between when the next iterator.next() is called on a query plan.  With the previous implementation of this test, the same plan (`['B', 'C', 'E', 'F', 'A']`) would have been emitted every time which passed the test.   By adding a random timeout variance this will reproduce the error where the same host is in the query plan multiple times while other hosts are omitted.   With a fixed variance using setTimeout the query plans are generated like `['A','A','A','A','A']` ... `['B','B','B','B','B']` and so on. 

To reproduce against 2.1 branch:
```shell
git cherry-pick 78398b5
mocha test/unit/load-balancing-tests.js -g 'should yield host in a round robin manner when consuming'
```

This should produce an error like:

> Uncaught AssertionError: B appears 3 times in B,F,B,B,A.  Expected only once.

One possible remaining weakness is that this will always produce the same query plans, so if you had 5 hosts, you can only ever have these 5 permutations of plans:

```  
  [ 'A', 'B', 'C', 'E', 'F' ]
  [ 'B', 'C', 'E', 'F', 'A' ]
  [ 'C', 'E', 'F', 'A', 'B' ]
  [ 'E', 'F', 'A', 'B', 'C' ]
  [ 'F', 'A', 'B', 'C', 'E' ] 
```

What we could do is on generation of a query plan generate a randomized host list.   I noticed that none of the other drivers do this though, with the only exception being in the Token Aware Policies of  the [java-driver](https://github.com/datastax/java-driver/blob/2.1.6/driver-core/src/main/java/com/datastax/driver/core/policies/TokenAwarePolicy.java#L134-L137) and [ruby driver](https://github.com/datastax/ruby-driver/blob/master/lib/cassandra/load_balancing/policies/token_aware.rb#L129-L130) where it provides the option to shuffle replicas.

test/unit/load-balancing-tests.js
```diff
@@ -51,33 +51,66 @@ describe('RoundRobinPolicy', function () {
...
+      permutations.push(permutation);
+    }
+    var originalHosts = createHostMap(hostList);
+    var times = 30;
     policy.init(null, originalHosts, function () {
       async.times(times, function (n, next) {
         policy.newQueryPlan(null, null, function (err, iterator) {
           assert.equal(err, null);
-          for (var i = 0; i < originalHosts.length; i++) {
+          async.timesSeries(originalHosts.length, function (planN, iteratorNext) {
```
tolbertam: I should and will also add a similar test for DCAware as well.

lib/policies/load-balancing.js
```diff
@@ -167,6 +169,14 @@ DCAwareRoundRobinPolicy.prototype.newQueryPlan = function (keyspace, queryOption
   var self = this;
   var counter = 0;
   var remoteCounter = 0;
+
```
tolbertam: One small issue here is that we can generate some duplicate consecutive query plans if hosts is organized like:

```
aDC1, bDC2, cDC3, dDC1, eDC2, fDC3, gDC1
```

the plans returned would be:

```
aDC1, dDC1, gDC1
dDC1, gDC1, aDC1
dDC1, gDC1, aDC1
dDC1, gDC1, aDC1
gDC1, aDC1, dDC1
gDC1, aDC1, dDC1
gDC1, aDC1, dDC1
```

So it wouldn't be true round robin..will try to find a solution.

tolbertam: Hrmm, I don't imagine this being easy without filtering out local and remote hosts into separate arrays at the beginning of the query plan or doing some refactoring of load balancing policies so DC hosts can be maintained separately from one another.  My concern about creating arrays for each query plan can be computationally expensive when you have a large number of hosts.

Thoughts @jorgebay?

tolbertam:
Pushed commits that demonstrate that DCAwareRoundRobinPolicy can return the same query plan consecutively.   If we adjust that behavior the tests should start passing.

jorgebay:
I used the host map events (available thanks to #76) to cache an array of local nodes and remote nodes. 

Using those cached arrays, the iterator logic is straightforward.

PR #79, opened 2015-07-07
Reuse query plan on retries

jorgebay:


tolbertam:
looks good to me :+1:

PR #80, opened 2015-07-08
NODEJS-157 Subscribe to down event earlier on init of ControlConnection 

tolbertam:
For [NODEJS-157](https://datastax-oss.atlassian.net/browse/NODEJS-157), simply moving the subscribe on host down right after the connection is acquired will handle the case where an error is encountered during refreshHosts fixes the issue where the keyspace set up for the client no longer exists.


jorgebay:
:+1:  Looks good to me. 
It would be nice to include a test for it, I think I can create a unit test but not sure how to reproduce it in an integration test.

PR #81, opened 2015-07-09
Make client failover tests resilient by allowing duplicate down events.

tolbertam:
Allow duplicate down events as we can (rarely) get multiple down events on a control connection.

Also submits all tests concurrently to make it more likely to catch a request occurring while a host is going down.  I originally made the switch to timesLimit but that apparently reduces the likelihood of a request hitting the node as it's coming down.

Opened as PR to see how jenkins handles this (works great locally).   If all is well should be good to merge.


tolbertam:
Seems to be behaving well, passes across 9 different configurations (though previous code worked about 90% of the time, so will give this a few more runs).  It should be good to merge if the change looks fine.

PR #82, opened 2015-09-01
NODEJS-162 Add host as coordinator attribute to error object.

stanlemon:
See https://datastax-oss.atlassian.net/browse/NODEJS-162

### DIFF
--- a/lib/request-handler.js
+++ b/lib/request-handler.js
@@ -291,6 +291,7 @@ RequestHandler.prototype.prepareOnConnection = function (queries, callbacksArray
 RequestHandler.prototype.handleError = function (err, callback) {
   //Add the error to the tried hosts
   this.triedHosts[this.host.address] = err;
+  err['coordinator'] = this.host.address;
   var decisionInfo = { decision: retry.RetryPolicy.retryDecision.rethrow };
   //noinspection JSUnresolvedVariable
   if (err && err.isServerUnhealthy) {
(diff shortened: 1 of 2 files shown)


stanlemon:
@jorgebay Do you have any recommendations on where to add a test for this change?  I looked around in test/unit/request-handler-tests.js and figured that was probably the best place but wasn't sure if I should add to an existing test or write a new one.

jorgebay:
Yes, unit test is a good way to go.

Integration tests requires you to install [ccm](https://github.com/pcmanus/ccm) which is a great tool to create Cassandra clusters with multiple nodes on a single machine, but it takes a little more time to set up.

lib/request-handler.js
```diff
@@ -336,6 +336,7 @@ RequestHandler.prototype.handleError = function (err, callback) {
   utils.fixStack(this.stackContainer.stack, err);
   if (this.request instanceof requests.QueryRequest || this.request instanceof requests.ExecuteRequest) {
     err['query'] = this.request.query;
+    err['coordinator'] = this.host.address;
```
jorgebay: It can be done out of the `if` block, that way is always for all requests.

stanlemon:
@jorgebay Added a test. What do you think about moving the addition of the coordinator attribute up? I would like to get this info on read timeouts too (my current problem) and that wouldn't get tacked on if this was scoped just to the block on 338.

jorgebay:
Makes sense.

BTW: If you are getting server read/write timeouts, you can define what to do in the retry policy.

Can you squash your commits into a single one?

stanlemon:
@jorgebay squashed

PR #83, opened 2015-09-14
Allow socket close event to be emitted before connection close calls back

jorgebay:
https://datastax-oss.atlassian.net/browse/NODEJS-165

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -604,7 +604,7 @@ Connection.prototype.close = function (callback) {
   }
   if (!this.connected) {
     this.netClient.destroy();
-    callback();
+    setImmediate(callback);
     return;
   }
   var self = this;
@@ -612,7 +612,7 @@ Connection.prototype.close = function (callback) {
     if (hadError) {
       self.log('info', 'The socket closed with a transmission error');
     }
-    callback();
+    setImmediate(callback);
   });
   this.netClient.end();
   this.streamHandlers = {};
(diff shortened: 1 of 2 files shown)


tolbertam:
Looks good to me, reproduced issue on windows with 2.1.9 against master and can't reproduce problem on this branch :+1:

deepakkb:
Hello Jorge,
I installed nodejs cassandra driver 2.2.2 and having this issue. Do you have any suggestions?.

jorgebay:
Could you post the error and stack trace @deepakkb ?

deepakkb:
When my NODEJS code in Windows calls cassandra server with a simple CSQL like "select * from table1" , following is the response I get

"Error: No host could be resolved,info=Represents an error when a query cannot be performed because no host is available or could be reached by the driver.,innerErrors=null".

However I can do telnet to the Cassandra server using the IP address default port 9042 and can do ping the server. So there is no connectivity issue between my machine and the Cassandra server. This is an existing server which the .NET program can communicate successfully but NODEJS cannot. Yesterday after reading the post at https://datastax-oss.atlassian.net/browse/NODEJS-165, I downloaded Cassandra driver 2.2.2 but no luck. Should I install a older version?

jorgebay:
Which are the parameters you are using to create the `Client` instance?
Can you post the [driver logs](https://github.com/datastax/nodejs-driver#logging)?

Using v2.2.2 of the driver is OK.

PR #88, opened 2015-09-15
Fix: HostMap cache when modifying an item

jorgebay:


tolbertam:
looks good to me :+1:

PR #90, opened 2015-09-17
Expose materialized views metadata

jorgebay:
It is chained with NODEJS-163 as it requires C* 3.0 table and keyspace metadata parsing.


test/integration/short/client-execute-prepared-tests.js
```diff
@@ -707,6 +709,47 @@ describe('Client', function () {
         ], done);
       });
     });
+    describe('with materialized views', function () {
```
tolbertam: can use vdescribe to restrict to 3.0

jorgebay:
Fixed `vdescribe()`.
Merged into 3.0 branch.

PR #91, opened 2015-09-18
Frame coalescing

jorgebay:
Process all frames in the write queue and flush it into a single buffer once a threshold is reached.


jorgebay:
This pull request targeted master branch, closing in favor of #108

PR #92, opened 2015-09-21
Process changes in system_schema.columns metadata

jorgebay:


jorgebay:
Targets C* 3.0 rc1 or above

PR #94, opened 2015-09-22
Parse crc_check_chance column from 'tables' and 'views' metadata tables

jorgebay:


tolbertam:
This is covered by #97 so closing (please reopen if disagree).

PR #95, opened 2015-09-23
Add buffer for non-streaming rows messages

jorgebay:


jorgebay:
Close in favor of #109

PR #97, opened 2015-10-06
Include missing options to table and view metadata

jorgebay:


tolbertam:
Made a small change (feel free to squash) and added another test.   Looks good to me :+1:.

PR #98, opened 2015-10-07
New index metadata API

jorgebay:


tolbertam:
Added some more tests.  Looks good to me other than the issue with parsing CompositeTypes. :+1:

jorgebay:
Nice! +1 on the custom index unit test.

Just posted #99 for the `ColumnToCollectionType` issue. If we have to release 3.0 beta1 before merging master with the fix, we can either leave it as a known issue or cherry pick it.

tolbertam:
Awesome, thanks!  I'll take a look at #99 this morning and get back to you.  It'd be really nice to get that into 3.0 as well.

PR #99, opened 2015-10-08
Support collections in parse key types with composites

jorgebay:


tolbertam:
Looks good to me! :+1:.   I also added some tests for retrieving data using secondary indexes with prepared statements.   This would produce an error for map indexes prior to this fix like the tests in #98, but thought it'd be good to ensure prepared queries would work as well with secondary indexes.

jorgebay:
Good idea

PR #100, opened 2015-10-09
Fix setting the routing key for hardcoded parameters in prepared query

jorgebay:


tolbertam:
looks good to me :+1:, I removed part of the last commit that added `@debug` to `should support hardcoded parameters that are part of the routing key`

PR #103, opened 2015-10-28
Change default consistency level and use undefined as unset by default

jorgebay:


lib/client-options.js
```diff
@@ -20,7 +20,7 @@ function defaultOptions () {
       addressResolution: new addressResolution.AddressTranslator()
     },
     queryOptions: {
-      consistency: types.consistencies.one,
+      consistency: types.consistencies.localQuorum,
```
asilvas: This should probably be noted in the notes since it is a breaking change, and/or associated with a major version bump.

tolbertam:
Some tests are failing with the change to make useDefinedAsUnset the default because this exposed a bug (opened as [NODEJS-190](https://datastax-oss.atlassian.net/browse/NODEJS-190)).  I've made an attempt to fix that in #104.

lib/client-options.js
```diff
@@ -44,7 +44,8 @@ function defaultOptions () {
     authProvider: null,
     maxPrepared: 500,
     encoding: {
-      copyBuffer: true
+      copyBuffer: true,
+      useUndefinedAsUnset: true
```
tolbertam: This will be a problem for those using < 2.2 right?  Since we have the protocolVersion what do you think about adding a check to encode in that we only set a value to unset if protocolVersion >= 4?

tolbertam:
Looks good to me :+1:

lib/client-options.js
tolbertam: :+1: on making explicit in release notes.  This change will be incorporated into the first RC or GA release of driver 3.0 so it will be associated with a major version bump.

lib/client-options.js
jorgebay: `useUndefinedAsUnset` only affects protocol v4 and above, when using unset or `undefined` for protocol version < 4, [it will throw a TypeError](https://github.com/datastax/nodejs-driver/blob/9f83ddbf411f2d568f6b9c780d1fef71ffd1a3cb/lib/encoder.js#L179).

jorgebay: You are right, `useUndefinedAsUnset` affected protocol versions < 4.
The default behaviour in driver v2.1 (before that setting was introduced) was that `undefined` was encoded as `null`. I've restored the same behaviour by [checking protocol version](https://github.com/datastax/nodejs-driver/pull/103/files#diff-b364beb377c8302c49a6988ecb79bc26R170).

tolbertam: Awesome, that looks perfect, thanks!

jorgebay:
Changed tests to use RF 1 instead of using a 3 node cluster.

PR #104, opened 2015-10-29
NODEJS-190 Don't apply useUndefinedAsUnset to UDT and Tuple fields.

tolbertam:
For [NODEJS-190](https://datastax-oss.atlassian.net/browse/NODEJS-190).

Currently if you have an undefined field in a Tuple or UDT, encodeUdt/encodeTuple would attempt to contact a list of Buffers with undefined values passed in as `{ unset: true }` rather than `<Buffer FF FF FF FE>` (-2).  Instead of setting undefined values as unset for Tuple and UDT fields, this checks a new `nested` flag to determine whether or not the value to encode is nested in one of those types.

One possible downside of this is that a user would not be able to use `cassandra.types.unset` in a UDT or Tuple field value.  We could fix that by handling unsets in `encodeUdt` and `encodeTuple` but wasn't sure if that was worth it.

### DIFF
--- a/lib/encoder.js
+++ b/lib/encoder.js
@@ -160,14 +160,15 @@ Encoder.prototype.decode = function (buffer, type) {
 /**
  * @param value
  * @param {{code: number, info: *|Object}|String|Number} [typeInfo]
+ * @param nested Whether or not the value is nested in a type such as a UDT or tuple.
  * @returns {Buffer}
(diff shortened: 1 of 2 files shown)


jorgebay:
Published #104 that uses a different approach and validates for null/unset in collection values.

tolbertam:
Sounds good, I'll close this one, thanks!

PR #105, opened 2015-10-29
Support unset for udts and validate collections with null/unset values

jorgebay:
Chained with #103

### DIFF
--- a/lib/encoder.js
+++ b/lib/encoder.js
@@ -56,6 +56,7 @@ var singleTypeNamesLength = Object.keys(singleTypeNames).reduce(function (previo
   return current.length > previous ? current.length : previous;
 }, 0);
 var nullValueBuffer = new Buffer([255, 255, 255, 255]);
+var unsetValueBuffer = new Buffer([255, 255, 255, 254]);
 
 /**
  * Encodes and decodes from a type to Cassandra bytes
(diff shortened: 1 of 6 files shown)


lib/encoder.js
```diff
@@ -906,6 +911,9 @@ Encoder.prototype.encodeMap = function (value, subtypes) {
     //include item
```
tolbertam: From what I can tell a key cannot be undefined or null right?  Can't seem to produce a scenario where it is, so makes sense not to have to handle it.

test/integration/short/client-execute-prepared-tests.js
```diff
@@ -530,6 +530,90 @@ describe('Client', function () {
...
+          var map = {};
+          map[tid] = 1;
+          map[types.TimeUuid.now()] = null;
+          var query = util.format('INSERT INTO %s (id1, id2, map_sample) VALUES (?, ?, ?)', commonTable);
+          client.execute(query, [Uuid.random(), tid, map], { prepare: true}, function (err) {
+            helper.assertInstanceOf(err, TypeError);
+            next();
+          });
+        },
+        function testMapWithNull(next) {
```
tolbertam: This should be `testMapWithUnset` right?

jorgebay: right

lib/encoder.js
jorgebay: Object can not contain null keys but looking into ES6 `Map`, it looks like it can contain a null/undefined key :open_mouth: , this is valid:

```javascript
var m = new Map();
m.set(1, 1);
m.set(null, 2);
```
Will fix

tolbertam:
Looks great to me, everything running cleanly in ci as well :+1:

lib/encoder.js
tolbertam: Good catch!  Did not even consider that.

PR #106, opened 2015-11-02
expose encoder in index.js

apollocatlin:
This will greatly help usage of the encoder since at present we need to create a hard link to a it as dependency of a dependency in node_modules. This should be exposed at the top level for consumption even when npm dedupe is used.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,5 @@
 exports.Client = require('./lib/client.js');
+exports.encoder = require('./lib/encoder.js');
 exports.types = require('./lib/types');
 exports.errors = require('./lib/errors.js');
 exports.policies = require('./lib/policies');


jorgebay:
What is the usage you are planning to give to the exposed `Encoder` prototype?

Is it to provide custom codec methods for certain types?

sbisbee:
Hey @jorgebay - Below are the two reasons we did this starting in 2014.

  - We were running into a ton of hinting bugs. Not sure if this would still be an issue with the current driver/server version combo we're using today, but I'm not inclined to mess with this code again since explicit typing is both easier and safer here.
  - Our batch writing code inspects the buffers so that we can chunk based on batch size. Ex., take a batch of writes that are targeted at a single partition, then sub-batch by 5120 bytes (server warn threshold). The math and therefore performance is a lot better when you work with the bytes that are actually being allocated and sent to the server, instead of having to hand roll our own math based on JS types.

Much appreciated,
Sam

jorgebay:
I understand, I remember it was a pain to deal with batch parameters types in the past.
This was [fixed in v2.0 of the driver](https://datastax-oss.atlassian.net/browse/NODEJS-26): remove all the parameter hints, set the prepare flag and that's it, just works :tm: .

If you want to send the raw data (`Buffer`) using the driver, for some types this is already allowed: `custom`, `bigint`, `blob`, `decimal`, `uuid`, `timeuuid` and `varint`.

I think it should allow all types to be input as raw bytes: I'll create a ticket for it.

jorgebay:
Exposing `Encoder` prototype would force to maintain it as part of the API and introduce breaking changes if serialization is changed in future versions.

I've created [NODEJS-192](https://datastax-oss.atlassian.net/browse/NODEJS-192) to support `Buffer` as parameter value for any type. I think this should help you in what you are trying to achieve.

sbisbee:
Hey @jorgebay - Thanks for the link to the batch ticket, I'll give it a whirl this weekend. Might reduce some lines of code for us.

I'm not sure about the Buffer bit though since I would still have to implement the mapping that's inside of `Encoder.encode()` to map ECMA to C* types, which seems worse than exposing an API.

jorgebay:
We could expose custom codecs, [similar to the java driver's custom codecs](https://datastax.github.io/java-driver/features/custom_codecs/) per datatype but exposing the `Encoder` itself I don't think its a good idea, as it is designed to be internal and could constraint future changes.
I think to count the size of the batches, you would have to do it mostly manually, as besides the parameter size itself, you have to consider the other parts of the [protocol](https://git-wip-us.apache.org/repos/asf?p=cassandra.git;a=blob_plain;f=doc/native_protocol_v4.spec;hb=refs/heads/cassandra-2.2): 
- 16 bytes for queryid
- 2 bytes for the length of all parameters.
- 2 bytes extra per parameter to identify each parameter length.

I'll close this one, feel free to [create a ticket](https://datastax-oss.atlassian.net/browse/NODEJS/issues) if you think we can improve the API some other way.

sbisbee:
Created ticket: https://datastax-oss.atlassian.net/browse/NODEJS-195

PR #110, opened 2015-11-06
A better Client.stream() implementation to solve NODEJS-194

yeiniel:
NODEJS-194 is caused by the fact that the Client.stream() API is based on the Client.eachRow() API and therefore back pressure cant be handled. This pull request provide a Client.stream() API implementation based on Client.execute().


jorgebay:
Looks good but #89 already addresses this issue.

PR #111, opened 2015-11-06
Manual paging via nextPage() and client.stream() throttling

jorgebay:


tolbertam:
looks good to me :+1:

PR #113, opened 2015-11-12
Metadata: Parse udt names with quoted identifiers

jorgebay:


tolbertam:
I added a slight change that handles the case where double quotes are used within a quoted identifier which is valid (although hopefully no one uses it :)).  Hope that change looks ok.

PR looks good to me!  :+1:

PR #114, opened 2015-11-13
BigDecimal.fromString() throws a TypeError when there is a conversion error

jorgebay:


tolbertam:
looks good to me! :+1:   The error is now properly surfaced as a TypeError instead of an Error and the host is thus not considered unhealthy.

**Before:**
```
{ [Error: All host(s) tried for query failed. First host tried, 127.0.0.1:9042: Error: Invalid null or undefined value. See innerErrors.]
  innerErrors: { '127.0.0.1:9042': { [Error: Invalid null or undefined value] isServerUnhealthy: true, coordinator: '127.0.0.1:9042' } },
  info: 'Represents an error when a query cannot be performed because no host is available or could be reached by the driver.',
  message: 'All host(s) tried for query failed. First host tried, 127.0.0.1:9042: Error: Invalid null or undefined value. See innerErrors.' }
```

**After:**
```
{ [TypeError: Invalid null or undefined value]
  coordinator: '127.0.0.1:9042',
  query: 'insert into bd_test (k, v) values (0, ?)' }
```

PR #115, opened 2015-11-17
Schedule reconnection using setTimeout()

jorgebay:
`HostConnectionPool` reconnects using `setTimeout()`.

When all the hosts go down, the `ControlConnection` listens to `Host`'s `up` event instead of actively trying to reconnect.

### DIFF
--- a/lib/control-connection.js
+++ b/lib/control-connection.js
@@ -12,6 +12,7 @@ var HostMap = require('./host').HostMap;
 var Metadata = require('./metadata');
 var requests = require('./requests');
 var utils = require('./utils');
+var types = require('./types');
 
 var selectPeers = "SELECT peer,data_center,rack,tokens,rpc_address,release_version FROM system.peers";
 var selectLocal = "SELECT * FROM system.local WHERE key='local'";
@@ -37,6 +38,16 @@ function ControlConnection(options) {
   this.metadata = new Metadata(this.options, this);
   this.addressTranslator = this.options.policies.addressResolution;
   this.initialized = false;
+  /**
+   * Host used by the control connection
+   * @type {Host|null}
+   */
+  this.host = null;
+  /**
+   * Connection used to retrieve metadata and subscribed to events
+   * @type {Connection|null}
+   */
+  this.connection = null;
 }
 
 util.inherits(ControlConnection, events.EventEmitter);
@@ -85,7 +96,7 @@ ControlConnection.prototype.init = function (callback) {
     },
     this.getConnection.bind(this),
     function tryInitOnConnection(next) {
-      self.initOnConnection(true, next);
+      self.refreshOnConnection(true, next);
     }
   ], function (err) {
     self.initialized = !err;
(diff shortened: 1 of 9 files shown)


lib/control-connection.js
```diff
@@ -141,7 +245,7 @@ ControlConnection.prototype.getConnection = function (callback) {
  * it's the first time that trying to retrieve host information
  * @param {Function} callback
  */
-ControlConnection.prototype.initOnConnection = function (firstTime, callback) {
+ControlConnection.prototype.refreshOnConnection = function (firstTime, callback) {
   var c = this.connection;
   var self = this;
   self.log('info', 'Connection acquired, refreshing nodes list');
```
tolbertam: I noticed that subscribeHostEvents subscribes to 'down' events on a host, but the listener is never removed when the host goes down.  Is there something that prevents receiving down events for an old host?  I think as it would work right now, the hostDownHandler could get called for a host that was previously the control connection going down.  It seems like in this case we would detect that a connection is still set and log that the control connection is down and it's attempting to get one, but it wouldn't actually replace the connection (fortunately).   I'll test this out.

lib/host.js
```diff
@@ -127,6 +131,11 @@ Host.prototype.canBeConsideredAsUp = function () {
 Host.prototype.setDistance = function (distance) {
   distance = distance || types.distance.local;
   this.pool.coreConnectionsLength = this.options.pooling.coreConnectionsPerHost[distance];
+  //if (distance === types.distance.ignored && this.pool.connections.length > 0) {
```
tolbertam: Was this intentional?

lib/host.js
```diff
@@ -305,7 +335,47 @@ HostConnectionPool.prototype.forceShutdown = function () {
...
+  this.reconnectionTimeout = null;
+  var c = this._createConnection();
+  var self = this;
+  c.open(function (err) {
+    if (err) {
+      self.log('info', util.format('Reconnection attempt to host %s failed', self.address), err);
+      self.host.setDown();
+      return c.close();
+    }
+    self.host.setUp();
```
tolbertam: It might be a nice optimization somewhere around here to schedule `_maybeCreatePool` so the pool pre-emptively starts growing to core connections, this way we may not need to wait as long for the pool to grow to core when borrowing a connection after a reconnection.   I think this would help the scenario @gwicke [described](https://datastax-oss.atlassian.net/browse/NODEJS-155?focusedCommentId=23416&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-23416).

tolbertam: Initially in `_maybeCreatePool` we could choose to invoke the callback immediately if the pool is being created and at least 1 core connection exists we invoke the callback, that way `borrowConnection` could in theory proceed granted the host is up and there was a successful reconnection.

```javascript
   if (!this.creating && this.connections.length >= this.coreConnectionsLength) {
     return callback();	
   }
```

lib/host.js
```diff
@@ -251,14 +267,16 @@ HostConnectionPool.prototype._maybeCreatePool = function (callback) {
       return connections.length < self.coreConnectionsLength;
     },
     function iterator(next) {
-      var c = new Connection(self.address, self.protocolVersion, self.options);
-      //Relay the event idleRequestError
-      c.on('idleRequestError', function (err) {
-        //The pool will emit the event
-        self.emit('idleRequestError', err);
-      });
+      var c = self._createConnection();
```
tolbertam: core connections are currently created one at a time, we could optimize this by creating them all at once, although I think in the general case the number of core connections will be pretty low.

lib/control-connection.js
tolbertam: I couldn't seem to reproduce this, so perhaps this is not a problem (though I'd like to understand why)

test/integration/short/client-pool-tests.js
```diff
@@ -707,6 +697,47 @@ describe('Client', function () {
...
+          async.timesSeries(10, function (n, timesNext) {
+            client.execute(helper.queries.basic, function (err) {
+              //it should fail
+              assert.ok(err);
+              timesNext();
+            });
+          }, next);
+        },
+        function startNode3(next) {
+          helper.ccmHelper.startNode(3, helper.wait(5000, next));
```
tolbertam: Unfortunately 5 seconds doesn't seem to be enough time for our jenkins environment in some cases.  I've also seen this test fail because nodes remove not stopped and removed in previous tests.  I'll see if there's something we can do like listen for 'on' events for a host and then timeout after a prohibitively long time (like 30 seconds?).

lib/control-connection.js
tolbertam: oh geez, just realized we're doing a 'once' instead of 'on'.  Well that explains it :)

test/integration/short/client-pool-tests.js
tolbertam: Ahh i guess I misunderstood this, listening for 'up' events is not really good enough because the host is marked up and then the control connection is set.

lib/host.js
```diff
@@ -305,7 +335,47 @@ HostConnectionPool.prototype.forceShutdown = function () {
...
+};
+
+HostConnectionPool.prototype._attemptReconnection = function () {
+  this.reconnectionTimeout = null;
+  var c = this._createConnection();
+  var self = this;
+  c.open(function (err) {
+    if (err) {
+      self.log('info', util.format('Reconnection attempt to host %s failed', self.address), err);
+      self.host.setDown();
```
tolbertam: I've noticed that each time a reconnect attempt fails, the Host gets marked DOWN when it shouldn't be.  I think this is because canBeConsideredAsUp returns true since `self.unhealthyAt` is not being set.  Maybe it should be set here?

tolbertam: Ah, it's not that simple, since we still need to schedule another reconnection.  We need to tune the logic such that the 'down' event is not emitted in this case but we still attempt to a reconnect.

lib/host.js
```diff
@@ -63,11 +66,11 @@ Host.prototype.setDown = function() {
     //This host is not down, the pool is being shutdown
     return;
   }
-  this.log('warning', 'Setting host ' + this.address + ' as DOWN');
-  this.unhealthyAt = new Date().getTime();
   this.reconnectionDelay = this.reconnectionSchedule.next().value;
+  this.log('warning', util.format('Host %s considered as DOWN. Reconnection delay %dms.', this.address, this.reconnectionDelay));
+  this.unhealthyAt = new Date().getTime();
```
tolbertam: I think using `new Date().getTime()` here and in `canBeConsideredAsUp()` could be a problem since this uses the system clock instead of the realtime clock, so clock skew / time changes could mess things up.  Rarely i've seen that a reconnect is not tried for a host after it fails to connect, i haven't been able to reproduce it viably or figure out why, but I think it may be because of clock skew.   

It is either this or a reconnect is taking less than 1ms, which causes the check in hasTimePassed() to return false since it's a 'greater than' check instead of 'greater than equals'.  I've done some testing locally and i've observed cases where a reconnect is scheduled and then it fails 1001ms after it was submitted (where it's reconnect timer is 1000).

This will probably only ever be a problem in environments where the cassandra node is on the same system as the client.  I think in any case though that is setDown() is called as a result of a failed reconnection we should always schedule another reconnect, unless the host is actually up (reschedule if `unhealthyAt != 0` and failed reconnect maybe?)

tolbertam: So it looks like the problem was my guess that the reconnect was taking less than 1ms, reproduced this in the jenkins environment:

```
04:26:20 info HostConnectionPool Reconnection attempt to host 127.0.0.2:9042 failed { [Error: connect ECONNREFUSED 127.0.0.2:9042]
04:26:20   code: 'ECONNREFUSED',
04:26:20   errno: 'ECONNREFUSED',
04:26:20   syscall: 'connect',
04:26:20   address: '127.0.0.2',
04:26:20   port: 9042 }
04:26:20 Duration is less than or the same as reconnection delay! 1000 1000
04:26:20 << Duration is 1000 >>
04:26:20 Stop trying because we can't be considered as up.
```

A fix could be to either change the check in `hasTimePassed` to '>=' or to always force schedule a reconnect after a failed reconnect.

tolbertam:
I have an experimental branch [origin/NODEJS-155-test](https://github.com/datastax/nodejs-driver/tree/NODEJS-155-test) with both some fixes and test reliability changes that are now creating clean builds (so far, configured it to run repeatedly overnight to see how it looks in the morning).  It includes the following changes:

* Only emit a down even if not previously down in setDown.
* Change the duration check to >= reconnectDelay in hasTimePassed.
* Only refresh hosts if there is an active control connection.

I'm not sure if these are the proper fixes but they seem to resolve the problems I was observing.

lib/host.js
jorgebay: I think using `canBeConsideredAsUp()` makes no sense now as it should be only one reconnection attempt scheduled and is not being considered by the lbp in the mean time.
`is it up? -> set it down`
I'll replace it by calling `isUp()` instead. `canBeConsideredAsUp()` should remain there as it is part of the API.

lib/host.js
jorgebay: I don't know why I commented it out, it makes sense: a custom load balancing policy may determine that a host that was being used now should be ignored.

lib/host.js
jorgebay: I wanted to make the first connection and, if it succeeds, try to open the rest in parallel. Then I remembered that it is not common to use more than 1 connection per host, as generally there are multiple (1 per core) nodejs processes running in the same machine...

lib/host.js
jorgebay: I think what @gwicke described was related to using the user call to make reconnection attempts (in their case to dead nodes).

As you suggest, after a succeeded reconnection attempt we can call `_maybeCreatePool()` to start growing additional core connections (if any).

About using the first connection if its already "growing", it also makes sense, the C# driver does the same thing.

jorgebay:
I will pull them in this branch.

jorgebay:
I pulled your changes from NODEJS-155-test branch, CI is looking stable. I've changed the failover tests to use `async.timesLimit()` to slow them down a little.

jorgebay:
It looks like a second CI run surfaced an issue: `connection` field in `ControlConnection` can be `null`, causing an error while trying to get the metadata or encoding the routing key...
I'll push a fix soon.

jorgebay:
I published the fix and CI run is in better shape.

tolbertam:
looking good and continues to run great in CI.  :+1:

PR #116, opened 2015-11-18
Introduce a setting to capture stack trace before query execution

jorgebay:


tolbertam:
Looks good to me!  :+1: also added some unit tests to validate that the new options is used.

PR #117, opened 2015-11-18
Expose Encoder prototype

jorgebay:


tolbertam:
:+1:, looks good to me.  Just had a few suggestions on the jsdoc:

* One thing that might be nice would be to define `@property` for each data type in types.dataTypes so they show up in the jsdoc like what was done for types.consistencies [[link(http://docs.datastax.com/en/drivers/nodejs/3.0/module-types.html)].
* decode has a parameter named 'type', while encode has 'typeInfo' which seems to support more types.  Is there any reason why decode couldn't work the same as encode?  It would also be good to link to dataTypes in decode as well.

jorgebay:
I've added jsdoc for `dataTypes` as suggested.

Both `encode()` and `decode()` methods accept a type composed by `{ code: Number, info: Object|String|Array }` which is the "correct" representation of a type but `encode()` supports more formats (like the string representation of the data type) for guessing/hinting.

PR #118, opened 2015-11-19
Use closures instead of Function.prototype.bind() for common execution path

jorgebay:
Chained with NODEJS-155 #115


tolbertam:
The change to use closures instead of bind looks good to me. :+1:.   I'm adding comments about the timeout changes in #115

PR #119, opened 2015-11-19
Support parsing Custom Types in C* 3.0.

tolbertam:
For [NODEJS-202](https://datastax-oss.atlassian.net/browse/NODEJS-202)

- parseTypeName will use parseFqTypeName when typeName starts with a single quote (').
- parseFqTypeName will now fallback on returning a Custom Type if it cannot detect the type.
- Added test for custom type metadata.

TODO:

- Test UDT with custom type.
- Unit tests.


jorgebay:
Agree

tolbertam:
removing [wip] tag as I think I've finished adding tests.

jorgebay:
lgtm :+1:

PR #121, opened 2015-11-28
Missing return statement when calling callback

ibillboard:
Hi, there are bugs in error handling in Load balancing policies. The method `newQueryPlan` calls a callback with an error when the property `this.hosts` doesn't exists, but the method doesn't end and it continues to process the rest of the code and it eventually crashed because of TypeError: Cannot call method 'values' of null. I've added the return statement and unit tests. 

Ah, sorry for the misleading commit message. It is a commit message of another commit.

### DIFF
--- a/lib/policies/load-balancing.js
+++ b/lib/policies/load-balancing.js
@@ -65,7 +65,7 @@ util.inherits(RoundRobinPolicy, LoadBalancingPolicy);
  */
 RoundRobinPolicy.prototype.newQueryPlan = function (keyspace, queryOptions, callback) {
   if (!this.hosts) {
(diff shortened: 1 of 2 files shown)


jorgebay:
Nice catch! I'll merge it as soon as you sign the CLA.

I've created the JIRA ticket [NODEJS-203](https://datastax-oss.atlassian.net/browse/NODEJS-203) for it.
If you don't want to amend the commit message, we can do it for you.

jorgebay:
Landed in 84d56709429b8daca3dfb9371eaeb1e0b225b67d

Thanks!

havrlant:
You're welcome :)

PR #125, opened 2016-01-20
Fix #NODEJS-217 - Add encoding option for tuples encoded/decoded as Array

razvanz:
Add `tuple` encoding option for encoding/decoding tuple type as Array

[JIRA link](https://datastax-oss.atlassian.net/browse/NODEJS-217)


jorgebay:
I replied in the ticket https://datastax-oss.atlassian.net/browse/NODEJS-217

PR #127, opened 2016-02-02
Consider putting nodejs-driver types into a new repository

ezsper:
Would be nice to create a CQL builder which supports the same type classes as the driver, and make this CQL builder only dependent of the types repository, not the driver itself. So the CQL builder could be used along with the driver, but not necessarily.

### DIFF
--- a/lib/auth/plain-text-auth-provider.js
+++ b/lib/auth/plain-text-auth-provider.js
@@ -7,6 +7,7 @@ var Authenticator = provider.Authenticator;
  * Creates a new instance of the Authenticator provider
  * @classdesc Provides plain text [Authenticator]{@link module:auth~Authenticator} instances to be used when
  * connecting to a host.
+ * @extends module:auth~AuthProvider
  * @example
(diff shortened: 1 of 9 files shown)


jorgebay:
I think having the types exposed by the driver in a separate package is a good idea, specially considering the small amount of types defined by the ecmascript standard.
We are still discussing internally if it can be done (support / commitment / ...) and how.

For now, if you only want to use the types you can still depend on the driver package, using a wide range of versions, knowing that npm will flat the dependencies on the user side.

For example (dependencies)
```
user-package
|__ your package
|       |__ cassandra-driver (any version > 2.0)
|__ cassandra-driver (v3.x)
```
Will get flatten to:
```
user-package
|__ your package
|__ cassandra-driver (v3.x)
```

BTW: Did you open the pull request to avoid going to the JIRA issue tracker, right?
If so, can I close it?

ezsper:
It's good that you're discussing it. Yes, you can close it.

PR #128, opened 2016-03-14
Read timeout at statement level

jorgebay:
`readTimeout`:

> When defined, it overrides the default read timeout (`socketOptions.readTimeout`) in milliseconds for this execution per coordinator.
> 
> Suitable for statements for which the coordinator may allow a longer server-side timeout, for example aggregation queries.
> 
> A value of `0` disables client side read timeout for the execution. Default: `undefined`.

Also alphabetized and added more info on QueryOptions jsdocs.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -98,43 +98,60 @@ var clientOptions = require('./client-options');
 /**
  * Query options
  * @typedef {Object} QueryOptions
- * @property {Number} consistency Consistency level. Default: localOne.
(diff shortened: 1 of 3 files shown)


tolbertam:
looks good :+1:

tolbertam:
Have a few tests i'm working on adding here (mostly for readTimeout = 0 on queryOptions).  Will push shortly!

tolbertam:
CI looking good, think this one is ready :+1:

PR #129, opened 2016-03-14
Generate monotonically increasing timeuuids for monotonically increasing dates

SebastianMax:
We are having a problem with the current TimeUuid implementation of the nodejs driver: From time to time we generate two TimeUuids using the same date (especially when importing events from a different system). Most of the times, these will have the correct order, i.e. they have the same date component and increasing ticks.
However, every 10000 times these two TimeUuids are created when the internal tick counter flips from 9999 to 0. Then, they will have the same date component but the ticks decrease, and they will be stored in the wrong order.
The following pull requests adds a unit test for this problem and fixes it by remembering the time for which the last tick was created, such that it resets the _ticks not only if it reaches 10000 but also if time advances.

### DIFF
--- a/lib/types/time-uuid.js
+++ b/lib/types/time-uuid.js
@@ -24,8 +24,10 @@ var _ticksInMs = 10000;
 var _ticks = 0;
 /**
  * Remember the last time when a timeuuid tick was generated
+ * @private
+ * @type {number}
  */
-var _lastTickTime = 0;
+var _lastTimestamp = 0;
 /**
  * Creates a new instance of Uuid based on the parameters provided according to rfc4122.
  * If any of the arguments is not provided, it will be randomly generated, except for the date that will use the current date.
@@ -208,9 +210,9 @@ function getNodeId(nodeId) {
 function getTicks(time, ticks) {
   if (typeof ticks !== 'number'|| ticks >= _ticksInMs) {
     _ticks++;
-    if (_ticks >= _ticksInMs || time !== _lastTickTime) {
+    if (_ticks >= _ticksInMs || time > _lastTimestamp) {
       _ticks = 0;
-      _lastTickTime = time;
+      _lastTimestamp = time;
     }
     ticks = _ticks;
   }
--- a/lib/types/time-uuid.js
+++ b/lib/types/time-uuid.js
@@ -23,6 +23,10 @@ var _ticksInMs = 10000;
  */
 var _ticks = 0;
 /**
+ * Remember the last time when a timeuuid tick was generated
+ */
+var _lastTickTime = 0;
+/**
  * Creates a new instance of Uuid based on the parameters provided according to rfc4122.
  * If any of the arguments is not provided, it will be randomly generated, except for the date that will use the current date.
  * @class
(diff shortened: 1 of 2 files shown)


tolbertam:
thank you for your contribution.  This makes sense to me.  Isn't this not technically monotonic though since the tick gets reset after 10k increments in the millisecond?   Don't think it needs to be fixed here but we could do [what the java driver does](https://github.com/datastax/java-driver/blob/3.0.0/driver-core/src/main/java/com/datastax/driver/core/utils/UUIDs.java#L237-L270) and defer until the next millisecond

SebastianMax:
Hi, thanks for the reply. Of course you're right, when generating >10k timeuuids in a millisecond it (still) generates colliding uuids; I like the approach of the java driver in this edge case.
How do you want to proceed with the pull request?

jorgebay:
I think doing busy spin as the java driver does is not an option as we have a single core.

The current implementation does not guarantee that `timeuuidA` < `timeuuidB` when both were generated using the system date and `timeuuidA` was generated first, as @SebastianMax explains. We must guarantee that the statement is true up to 10K generated timeuuids.

The patch only considers timeuuid generation using the system date (`TimeUuid.now()`) and not timeuuids generated using a fixed date (millis). If we start using a fixed date A and a fixed date B, alternatively, the ticks generated will always be 0.

I propose 2 different ticks instances: 
- One `ticks` instance for when using a generated date, which we reset to 0 when the milliseconds of the current date changes.
- Another `ticks` instance used for fixed dates, that only gets reset when `value >= _ticksInMs`

That way we guarantee the correct ordering of generated timeuuids based on the current date (up to 10K). 
For timeuuid generated using fixed dates, we document that the ticks portion is not guaranteed to be greater than the previous generated using the same fixed date, recommending users to specify the ticks portion when that is not desirable. I think its the best solution, given that we allow users to just specify the milliseconds portion of a timeuuid, without requiring the ticks.

SebastianMax:
Ah, I've overseen this case, thanks! I've added a failing test covering the case of the colliding `timeuuids` if three `timeuuids` are generated with two fixed dates A and B, A < B. Of course, the generated ticks are always 0 and thus the test fails to generate 3 different `timeuuids`.

However, using 2 different ticks instances will not fix the problem, as also for fixed dates the non-monotonically increasing `timeuuids` may be produced - this is exactly what I'm doing in the first submitted unit test.

I'd propose the fix in the second patch, which simply resets _ticks only if the used time is greater than the last time.

jorgebay:
I think we shouldn't have allowed users to specify the Date (ms since unix epoch) without specifying the 100-nanoseconds (ticks) portion but I think we have to do our best to supported until removed from the API (breaking).

In this case, I think it translates into making a single method that generates the complete timestamp (with 100-nanoseconds precision and resetting the counter once the millis portion changed). Note that the java driver has a single `getCurrentTimestamp()` method.
When the users provide the `Date` but not the ticks portion, we can do a "lets try our best" approach and have another counter, that only gets resetted once ticks >= 10000.

Otherwise, 2 different calls made alternatively in 2 different parts of an application using fixed dates (but not ticks) will gets the ticks portion 0 most of the time, that is worst than the current behaviour.

Would you like to change your patch?

SebastianMax:
I've change the patch accordingly - the TimeUuid.now() uses a different counter than the _ticks which is used if the Date is supplied at TimeUuid.fromDate().

jorgebay:
Looks good, I will squash the commits and merge.

jorgebay:
Landed in 68c2c73286166b65b264460a78637b816e2dd167 and 1c42abbe8581c2936381b61aa241c1f447e30857.

Thanks!

SebastianMax:
Great! Do you also increase the version number to 3.0.2?

jorgebay:
There are other [tickets planned for v3.0.2](https://datastax-oss.atlassian.net/browse/NODEJS-228?filter=-4&jql=project%20%3D%20NODEJS%20AND%20fixVersion%20in%20(3.0.2)%20ORDER%20BY%20updated%20) (#131 #130 #128) and also, before each release, we ran endurance tests against the new version. But we should release v3.0.2 soon.

PR #130, opened 2016-03-17
Improve handling of parsing errors

jorgebay:
Improves error handling of stream deserialization.

Also:
- Avoid buffer region copies on `Protocol` stream.
- Use protocol version from first response message for protocol negotiation (instead of decreasing the protocol version one by one).

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -78,8 +78,8 @@ Connection.prototype.bindSocketListeners = function() {
     self.clearAndInvokePending();
   });
 
-  var protocol = new streams.Protocol({objectMode: true}, this.protocolVersion);
-  this.parser = new streams.Parser({objectMode: true}, this.encoder);
+  this.protocol = new streams.Protocol({ objectMode: true });
+  this.parser = new streams.Parser({ objectMode: true }, this.encoder);
   var resultEmitter = new streams.ResultEmitter({objectMode: true});
   resultEmitter.on('result', this.handleResult.bind(this));
   resultEmitter.on('row', this.handleRow.bind(this));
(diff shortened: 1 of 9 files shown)


tolbertam:
Looks good so far!   Running a perf test out in ec2 to see how this fares against previous.  Will post numbers this afternoon.

tolbertam:
Getting a nice ~4% throughput bump with this PR :).  (6896.27r/s vs. 7137.96 r/s on a c4.xlarge).  Will do some more benchmarks, but this looks promising.

jorgebay:
Not much bit is something :/

tolbertam:
I just took a look at the memory utilization and it does appear this uses slightly less heap.  `heapUsed` from `process.memoryUsage()` averages 55.04mb vs. 60.34, with rss pretty close in both cases (~129-133mb) for this particular test.  That could all be within some margin of error though, i'l see if there's better way to capture memory/gc metrics to see if there is any tangible benefit there.

jorgebay:
Rebased against master

tolbertam:
It seems that any performance gains here are marginal with what i've tried so far.  Any suggestions for other scenarios i should try?  I've tried the following:

1. Insert throughput w/ variable concurrent requests up to 500.
2. Read small page size / default page size (5000) / large page size (20000).

That being said, this appears to be functioning well and behaving as expected, so this gets my :+1:  for merging.

jorgebay:
The refactor affects the read path of the driver, it should affect the read performance, regardless if the results are VOID/ROWS. I'm getting the same results on my end: there isn't a noticeable performance gain in NODEJS-235.

I'm happy with the patch, a refactoring of streams module was long overdue (it was basically the same code from jorgebay/node-cassandra-cql), error handling on the read path is refined and the module code is more readable (at least in my POV).

We should allocate some time to tweak the throughput benchmarks, as currently they are biased against the async library. Also, after NODEJS-240, the CPU profiling results will become more meaningful.

As an additional premise, it seems that the low hanging fruits regarding performance improvements are on the write path.

PR #131, opened 2016-03-17
Connection: use an specific handler property for rowCallback

jorgebay:
Fixes NODEJS-238.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -429,6 +429,7 @@ Connection.prototype.getWriteCallback = function (request, options, callback) {
     }
     self.streamHandlers[request.streamId] = {
       callback: callback,
+      rowCallback: options && options.rowCallback,
       options: options,
       timeout: timeout
     };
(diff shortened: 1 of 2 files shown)


lib/connection.js
```diff
@@ -548,11 +549,13 @@ Connection.prototype.handleRow = function (header, row, meta, rowLength, flags)
...
     handler.timeout = null;
   }
-  handler.rowIndex = handler.rowIndex || 0;
-  var rowCallback = handler.options && handler.options.rowCallback;
-  if (rowCallback) {
-    rowCallback(handler.rowIndex++, row, rowLength);
+  if (!handler.rowCallback) {
+    return this.invokeCallback(
+      handler,
+      new errors.DriverInternalError('RowCallback not found for streaming frame handler'));
```
tolbertam: Is this technically even possible?  I've tried using eachRow without a row callback and the logic in parseCommonArgs is exercised.   Still good to have this here i suppose :)

tolbertam:
looks good to me :+1:

lib/connection.js
jorgebay: Good point, its not possible using the public API, that is the reason it is a `DriverInternalError`.

PR #132, opened 2016-03-21
support decoding timestamp as numeric value

eladkarakuli:
Add support to decode the timestamp type as its numeric value.
This option is required, when communicating with client side with the numeric timestamp value.

### DIFF
--- a/lib/client-options.js
+++ b/lib/client-options.js
@@ -47,7 +47,8 @@ function defaultOptions () {
     maxPrepared: 500,
     encoding: {
       copyBuffer: true,
-      useUndefinedAsUnset: true
+      useUndefinedAsUnset: true,
+      useNumericTimestamp: false
     }
   });
 }
--- a/lib/client.js
+++ b/lib/client.js
@@ -93,6 +93,11 @@ var clientOptions = require('./client-options');
  * <p>
  *   Default: true.
  * </p>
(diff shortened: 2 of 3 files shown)


jorgebay:
You get the same behaviour by using [Date.prototype.getTime()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime) client side.

eladkarakuli:
Indeed. But what if there's a need to communicate the timestamp as epoch value?
I think same approach as with map and set, you could say that instead of decoding a set using by a switch with encodingOptions your could use new Set() in your code.

jorgebay:
We are looking to [provide 1 to 1 type mapping between CQL and Javascript](http://docs.datastax.com/en/developer/nodejs-driver/3.0/nodejs-driver/reference/nodejs2Cql3Datatypes.html), `Date` is one of the few builtin objects in Ecmascript and represents the same information (single moment in time) as the cql counterpart.

If you need to communicate the value to third party, you should adapt the value (ie: [providing a replacer function to the JSON.stringify()](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#The_replacer_parameter) and using `getTime()` method) but that should be outside your data layer.

You could also use the [experimental API for overriding encoding and decoding behaviour](https://datastax-oss.atlassian.net/browse/NODEJS-195), even though I would advice you against as the value is properly represented in the driver and the API is experimental and likely to be changed in future versions.

PR #133, opened 2016-03-21
Consider racks on NetworkTopologyStrategy

jorgebay:


lib/metadata/schema-parser.js
```diff
@@ -1009,37 +1009,24 @@ function getTokenToReplicaSimpleMapper(replicationFactor) {
...
-      return true;
-    }
-    //For each token
-    //Get an Array of tokens
-    //Checking that there aren't more tokens per dc than specified by the replication factors
     var replicas = {};
     for (var i = 0; i < ring.length; i++) {
       var token = ring[i];
       var key = tokenizer.stringify(token);
       var tokenReplicas = [];
```
tolbertam: If you are using vnodes, a replica can have multiple tokens.  In the case where not enough replicas are found to meet CL, it's possible tokenToReplicas will be populated with the same replicas multiple times.  In this case, we should skip repeated replicas and move on to the next replica.

tolbertam: have a test that exhibits this that i'll add shortly.

tolbertam: Hrmm, I think there may also be a problem with multiple datacenters, or how I set up my test.   When I call getReplicas() i'm only getting the number of replicas == RF in 1 dc, not combined.  Will see if I can figure out if its something I'm doing wrong here or if its the code.

jorgebay: Maybe its easier to repro with a unit test? Ive added some to unit/metadata-tests.js but probably Ive missed something...

tolbertam: That is what I have done, but it's not quite behaving as I like, diving into a debugger to see what I'm missing.  It is likely i'm doing something wrong.  I'm attempting to simulate 100 hosts with 256 vnodes each and then ensure that the replicas returned are as expected, this isn't in the best shape but here's what i currently got:

```javascript

    it('should run expediently with many replicas and 0 nodes in DC.', function (done) {
      var cc = getControlConnectionForRows([{
        'keyspace_name': 'dummy',
        'strategy_class': 'NetworkTopologyStrategy',
        'strategy_options': '{"dc1": "3", "dc2": "3"}'
      }]);
      var options = clientOptions.extend({}, helper.baseOptions);
      var metadata = new Metadata(options, cc);
      metadata.tokenizer = getTokenizer();
      var racksDc1 = new utils.HashSet();
      racksDc1.add('dc1_r1');
      var racksDc2 = new utils.HashSet();
      racksDc2.add('dc2_r1');
      metadata.datacenters = {
        'dc1': { hostLength: 100, racks: racksDc1 },
        'dc2': { hostLength: 0, racks: racksDc2 }};

      metadata.ring = [];
      // create ring and replicas
      metadata.primaryReplicas = {};
      for (var r = 0; r < 100; r++) {
        var h = new Host(r.toString(), 2, options);
        if (r == 0) {
          h.datacenter = 'dc1';
          h.rack = 'dc1_r1';
        } else {
          h.datacenter = 'dc2';
          h.rack = 'dc2_r1';
        }
        // 256 vnodes per replica.
        for (var v = 0; v < 256; v++) {
          var token = (v * 256) + r;
          metadata.ring.push(token);
          metadata.primaryReplicas[token.toString()] = h;
        }
      }
      // sort the ring so the tokens are in order (this is done in metadata buildTokens, but it accounts for
      // partitioner which we don't need to use here).
      metadata.ring.sort(function (a, b) { return a - b });

      metadata.log = helper.noop;
      // Get the replicas of 0.  Since token 0 is a replica in DC1 and DC1 only has 0, it should return 0 and then
      // fallback on 1 and 2.
      metadata.refreshKeyspaces(function () {
        var replicas = metadata.getReplicas('dummy', new Buffer([0]));
        assert.ok(replicas);
        assert.deepEqual(replicas.map(getAddress), ['0', '0', '0']);
        // Get the replicas on
        metadata.refreshKeyspaces(function () {
          var replicas = metadata.getReplicas('dummy', new Buffer([7000]));
          assert.ok(replicas);
          assert.deepEqual(replicas.map(getAddress), ['2', '3', '4']);
          done();
        });
      });
```

I see in your tests that you are getting the right amount of replicas back, so it's incredibly likely i'm missing something :)

tolbertam: actually just about everything about the test (the description, the replica placement, etc.) is currently wrong.  What I get for hastily copy-pasting but should demonstrate what I am attempting.

tolbertam: Aha, so now it's behaving correctly now that i've adjusted the host lengths.   The only thing I need to resolve now is the # of replicas I get back.   I no longer get duplicate replicas, which is encouraging :)

lib/metadata/schema-parser.js
```diff
@@ -1071,6 +1078,33 @@ function getTokenToReplicaNetworkMapper(replicationFactors) {
...
+  var keys = Object.keys(replicationFactors);
+  for (var i = 0; i < keys.length; i++) {
+    var dcName = keys[i];
+    var dc = datacenters[dcName];
+    if (!dc) {
+      // A DC is included in the RF but the DC does not exist in the topology
+      continue;
+    }
+    var rf = Math.min(parseInt(replicationFactors[dcName], 10), dc.hostLength);
+    if (replicasByDc[dcName] < rf) {
```
tolbertam: I think this explains the bug I'm seeing.  It's possible that replicasByDc[dcName] does not exist at this point if the datacenter has not been encountered yet.   It only gets populated when a host is encountered while parsing replicas.

tolbertam: This seems to fix it:

```javascript
    var rf = Math.min(parseInt(replicationFactors[dcName], 10), dc.hostLength);
    if (!replicasByDc[dcName] || replicasByDc[dcName] < rf) {
      return false;
    }
```

I'll push it up with my tests but feel free to revert if there is a better solution or i'm wrong about my assumptions :)

tolbertam: This is not perfect since it needs to consider the case where there are 0 replicas belonging to a DC.  Checking hostLength for 0 explicitly should short circuit that.

tolbertam:
Made a small change as described [here](https://github.com/datastax/nodejs-driver/pull/133#discussion_r57947351) which prevents skipping of including replicas from a DC if no host in that DC has been encountered yet.

There are still a few possible weaknesses with the implementation, but based on how C* places tokens I don't think we need to worry about it:

1. The algorithm is still n^2, but really the only case where you can even approach this being a problem is if replicas in a DC are placed at the very end of a token range, meaning all tokens leading up to it need to be evaluated first.   This is not a problem given that DC placement is not typically done this way, and there should really only be a dramatic amount of tokens if using vnodes, in which case tokens will be evenly distributed.   The java driver has this same issue.   We could optimize this if we kept a map of DC:Host and then evaluate each Dc's nodes at a time, but I think this is pretty complex and not worth it.
2. It is technically possible, but not should not be seen in practice to have repeated replicas returned for a given token.   I was able to reproduce this by creating a token range with vnodes where the vnodes were all consecutive, which in general wouldn't happen, i.e.:

    ```javascript
          // create ring with 100 replicas and 256 vnodes each.
          for (var r = 0; r < 100; r++) {
            var h = new Host(r.toString(), 2, options);
            h.datacenter = 'dc1';
            h.rack = 'dc1_r1';
            // 256 vnodes per replica.
            for (var v = 0; v < 256; v++) {
              var token = (r * 256) + v;
              metadata.ring.push(token);
              metadata.primaryReplicas[token.toString()] = h;
            }
          }
    ```

All that said, this ticket looks great to me! :+1:

lib/metadata/schema-parser.js
```diff
@@ -1098,7 +1098,7 @@ function isDoneForToken(replicationFactors, datacenters, replicasByDc) {
       continue;
     }
     var rf = Math.min(parseInt(replicationFactors[dcName], 10), dc.hostLength);
-    if (replicasByDc[dcName] < rf) {
+    if (rf != 0 && (!replicasByDc[dcName] || replicasByDc[dcName] < rf)) {
```
jorgebay: `!=` operator should be avoided as it performs conversions. Its best to evaluate `rf !== 0` or `!rf`.

jorgebay:
About 1: I initially went for that approach when rewriting this, it felt more natural to write it, but it required the additional map of maps. This approach is safe, as long as `isDoneForToken()` works as expected: I will add more tests to cover that function specifically.

About 2: Now that you bring it up, it could theoretically happen. I understand that this wont happen with virtual nodes, as consecutive tokens are not an option; and for manual tokens, you just choose 1 token.

Your change looks good, I'll add some tests to cover `isDoneForToken()` using rewire and merge it.

lib/metadata/schema-parser.js
tolbertam: ah!  good catch

PR #134, opened 2016-03-21
Performance: Use neo-async instead of async

gwicke:
The node cassandra driver uses https://github.com/caolan/async for much of its
control flow. As a result, the `async` library shows up prominently in cpu
profiles, typically using most cpu time of all driver components.

The neo-async library (https://github.com/suguru03/neo-async) promises a
light-weight drop-in replacement, with lower memory and cpu consumption. After
seeing promising results in bluebird's micro-benchmark suite, I was curious
whether those results would translate to real-world applications.

To test this, I replaced `async` with `neo-async`, and ran a single-threaded
web server benchmark:

- using https://github.com/wikimedia/restbase: `node server.js -c
  config.example.wikimedia.yaml -n 1`
- Each request involves two simple Cassandra read queries.
- Test run: `wrk -H'Accept-encoding: gzip' --latency -d60
  http://localhost:7231/en.wikipedia.org/v1/page/html/FOO`
- First run for warm-up, second run for actual measurement.

Results:

library | throughput | memory
---------|----------------|--------------
async     | 670 req/s | 320m RSS
neo-async | 788 req/s | 222m RSS

So, it looks like neo-async does indeed improve performance significantly in
this benchmark.

Out of an abundance of caution, this patch pulls in a fixed version of
neo-async. This way, new versions of neo-async can be tested with the driver
before being officially pulled in.

### DIFF
--- a/examples/basic/basic-execute-flow.js
+++ b/examples/basic/basic-execute-flow.js
@@ -1,7 +1,7 @@
 "use strict";
 //replace by    require('cassandra-driver');
 var cassandra = require('../../');
-var async = require('async');
+var async = require('neo-async');
 var assert = require('assert');
 
 var client = new cassandra.Client({ contactPoints: ['127.0.0.1']});
--- a/examples/metadata/metadata-table.js
+++ b/examples/metadata/metadata-table.js
@@ -1,7 +1,7 @@
 "use strict";
 //replace by    require('cassandra-driver');
 var cassandra = require('../../');
-var async = require('async');
+var async = require('neo-async');
 var assert = require('assert');
 
 var client = new cassandra.Client({ contactPoints: ['127.0.0.1']});
--- a/examples/runner.js
+++ b/examples/runner.js
@@ -1,6 +1,6 @@
 "use strict";
 
-var async = require('async');
+var async = require('neo-async');
 var exec = require('child_process').exec;
 var fs = require('fs');
 var path = require('path');
--- a/examples/tracing/retrieve-query-trace.js
+++ b/examples/tracing/retrieve-query-trace.js
@@ -1,7 +1,7 @@
 "use strict";
 //replace by    require('cassandra-driver');
 var cassandra = require('../../');
-var async = require('async');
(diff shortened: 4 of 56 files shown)


jorgebay:
I was sceptic at first, but I ran a [basic throughput test](https://github.com/riptano/nodejs-driver-sut/blob/master/test/throughput-test.js) and WOW! On my local dev machine, throughput when from 9038 ops/sec against master to 12135 ops/sec. It is a huge difference that can not be ignored (that continues to show that micro optimizations under v8 make a huge impact).

My only concern is that `neo-async` is not as widely used as `async` (26 K  downloads per month vs 26M / month) and it's 1 maintainer vs 2 maintainers, but its only community support so it should not make much difference, also the `async`/`neo-async` functionality is very limited in scope.
I read the code for `neo-async` `series()` and `waterfall()` and contains simple javascript optimizations

In the core driver, we use:

- `async.series()`
- `async.waterfall()` (that can easily be replaced by `series()` calls).
- `async.each()`, `async.eachSeries()`, `async.eachLimit()`.
- `async.whilst()`

We could also deliver our own methods async methods, optimized for our usage, removing the external dependency.

jorgebay:
I've created a JIRA ticket to track this: https://datastax-oss.atlassian.net/browse/NODEJS-240

gwicke:
We have been running this branch in production at @wikimedia for over a week now, and have not encountered any issues related to neo-async.

jorgebay:
Thanks for the heads up.

jorgebay:
I've replaced "async" dependency with our own async control flow methods in #138, removing the external dependency (which is always good news!).

Thanks for bringing this up. After removing "async", we were able to identify [NODEJS-247](https://datastax-oss.atlassian.net/browse/NODEJS-247) (#139), [NODEJS-248](https://datastax-oss.atlassian.net/browse/NODEJS-248) and [NODEJS-249](https://datastax-oss.atlassian.net/browse/NODEJS-249), as they are now easier to spot.

PR #135, opened 2016-03-30
Doc features, faq and getting started

jorgebay:


doc/features/address-resolution/README.md
```diff
@@ -0,0 +1,66 @@
+# Address resolution
+
+The driver auto-detects new Cassandra nodes when they are added to the cluster by means of server-side push
+notifications and checking the system tables.
+
+For each node, the address the driver receives the address set as `rpc_address` in the node's cassandra.yaml file. In
```
tolbertam: Should we mention the nuance about `broadcast_rpc_address` here in C* 2.1+, or is that probably too much info?

doc/features/address-resolution/README.md
```diff
@@ -0,0 +1,66 @@
...
+var addressResolution = cassandra.policies.addressResolution;
+var client = new Client({
+   contactPoints: ['1.2.3.4'], 
+   policies: { 
+      addressResolution: new addressResolution.EC2MultiRegionTranslator() 
+   }
+});
+```
+
+The `Client class performs a reverse DNS lookup of the origin address to find the domain name of the target instance.
```
tolbertam: Missing a backtick here after client.

doc/coding-rules/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+Preparing and executing statements in the driver does not require two chained asynchronous calls. You can set the
+`prepare` flag in the query options and the driver handles the rest.
+
+```javascript
+const query = 'SELECT id, name FROM users WHERE id = ?';
+client.execute(query, [ id ], { prepare: true }, callback);
+```
+
+## BATCH statements
```
jorgebay: Agree

jorgebay:
I will include jsdocs for `keyspace` in the `ClientOptions` and move the most of the `Client` jsdoc to `@classdesc`.

PR #136, opened 2016-03-30
Appveyor enablement

tolbertam:
Adds appveyor support to enable CI on windows.  Had to make a few tweaks to tests to make them run well on windows, but didn't take much.  Currently runs all short+unit tests.

Some things to determine:

* Do we want to run short tests all the time?  We have the bandwidth to do it.  Takes 30-35 minutes to run through all short tests.
* What combinations to test against?  Currently set up to test with nodejs driver 5.X and C* 3.0.4.  C* 2.2+ run reliably enough on windows.  Might be able to find a way to run against more combinations on master only on a weekly basis or something like that.

https://ci.appveyor.com/project/datastax/nodejs-driver

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -1,8 +1,8 @@
 nodejs:
   - "0.10"
   - "0.12"
-  - "4.3"
-  - "5.6"
+  - "4"
+  - "5"
 os:
   - ubuntu/trusty64
 cassandra:
--- a/build.yaml
+++ b/build.yaml
@@ -1,8 +1,8 @@
 nodejs:
   - "0.10"
   - "0.12"
-  - "iojs-3"
-  - "4.0"
+  - "4.3"
+  - "5.6"
 os:
   - ubuntu/trusty64
 cassandra:
@@ -19,6 +19,7 @@ build:
       CCM_PATH=$HOME/ccm
       JUNIT_REPORT_STACK=1
       JUNIT_REPORT_PATH=.
+      multi=mocha-jenkins-reporter=-
   - npm: install
   - npm: run-script ci
     graceful: true
--- a/build.yaml
+++ b/build.yaml
@@ -1,8 +1,8 @@
 nodejs:
   - "0.10"
   - "0.12"
-  - "4"
-  - "5"
+  - "4.3"
+  - "5.6"
 os:
   - ubuntu/trusty64
 cassandra:
--- a/README.md
+++ b/README.md
@@ -9,7 +9,8 @@ Enterprise (3.1+) using exclusively Cassandra's binary protocol and Cassandra Qu
 $ npm install cassandra-driver
 ```
 
-[![Build Status](https://travis-ci.org/datastax/nodejs-driver.svg?branch=master)](https://travis-ci.org/datastax/nodejs-driver)
+[![Build Status](https://travis-ci.org/datastax/nodejs-driver.svg?branch=master)](https://travis-ci.org/datastax/nodejs-driver) [![Build status](https://ci.appveyor.com/api/projects/status/m21t2tfdpmkjex1l/branch/master?svg=true)](https://ci.appveyor.com/project/datastax/nodejs-driver/branch/master)
+
 
 ## Features
 
--- a/appveyor.ps1
+++ b/appveyor.ps1
@@ -1,0 +1,61 @@
+$env:JAVA_HOME="C:\Program Files\Java\jdk1.8.0"
+$env:PYTHON="C:\Python27-x64"
+$env:PATH="$($env:PYTHON);$($env:PYTHON)\Scripts;$($env:JAVA_HOME)\bin;$($env:PATH)"
+$env:CCM_PATH="C:\Users\appveyor\ccm"
+
+# Install Ant and Maven
+$packages = "ant", "maven"
+Start-Process cinst -ArgumentList @("-y","$packages") -Wait -NoNewWindow
+# Workaround for ccm, link ant.exe -> ant.bat
+If (!(Test-Path C:\ProgramData\chocolatey\bin\ant.bat)) {
+  cmd /c mklink C:\ProgramData\chocolatey\bin\ant.bat C:\ProgramData\chocolatey\bin\ant.exe
+}
+
+# Install Java Cryptographic Extensions, needed for SSL.
+$target = "$($env:JAVA_HOME)\jre\lib\security"
+# If this file doesn't exist we know JCE hasn't been installed.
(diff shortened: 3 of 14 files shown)


tolbertam:
Looks like iojs doesn't like the deasync module (dependency of mocha-appveyor-reporter which is a devDependency).  Will remove from travis matrix (since we had discussed doing that anyways).

jorgebay:
> - Do we want to run short tests all the time? We have the bandwidth to do it. Takes 30-35 minutes to run through all short tests.

Its OK to have unit tests on all branches and unit + short on master and version branches, but if we have the bandwidth and it does not affect the rest of builds, its good to have short always.

> - What combinations to test against? Currently set up to test with nodejs driver 5.X and C* 3.0.4. C* 2.2+ run reliably enough on windows. Might be able to find a way to run against more combinations on master only on a weekly basis or something like that.

You mean nodejs driver under Node.js 5.X? We should run it against the LTS (0.10, 0.12, 4.x) if possible...

jorgebay:
If having Node.js 0.10, 0.12, 4.x on Windows is too much (now that I think about it, it seems too much), having the latest LTS is good enough.

tolbertam:
> You mean nodejs driver under Node.js 5.X? We should run it against the LTS (0.10, 0.12, 4.x) if possible...

Sounds good ,I also meant the C* version.  If we test against 0.10, 0.12, 4.x, 5.x every time and also C* versions (maybe 2.2.X, 3.0.X and 3.X for now) we'd have a pretty big matrix.   We have 5 concurrent workers shared on our account, but it probably won't be problem initially since we'll be one of the first drivers coming online.  We can eventually set up nightly builds of everything off master that builds and tests everything (appveyor has support for that).

tolbertam:
There's a few problems I need to work through, primarily:

* `ControlConnection #init() should subscribe to TOPOLOGY_CHANGE add events and refresh ring info` 
and `ControlConnection #init() should subscribe to SCHEMA_CHANGE events and refresh keyspace information` always seem to fail.   It could be a timing thing or maybe not.  
* `Client failover should failover after a node goes down` intermittently fails, this happens in jenkins too, i'll fix it.
* [`should handle multiple prepares in parallel`](https://ci.appveyor.com/project/DataStax/nodejs-driver/build/6.appveyor-enablement) sometimes fails and causes everything to tear down.  The stack trace looks like it might be something not handled appropriately in `mocha-appveyor-reporter` but that doesn't make much sense.  Will debug.

I think if I can wrap my head around those three things and if the rest of the PR looks good, it would be nice to get this in around the 3.0.2 timeframe so we can get a good windows run through against master.  If we still have intermittent problems I can always rebase against master and test with this branch.

build.yaml
```diff
@@ -1,8 +1,8 @@
 nodejs:
   - "0.10"
   - "0.12"
-  - "iojs-3"
-  - "4.0"
+  - "4.3"
```
jorgebay: In nvm you can just use "4" and it will install latest of the major version (4.4.1) , I don't know in nodeenv, though

tolbertam: Nice, that would be nice to use so we don't have to constantly update.  I'll see if it's something we can handle in our jenkins env with nodeenv.

jorgebay:
I think running the unit test on Windows against all active LTS releases (0.10, 0.12, 4.x) is necessary but we can limit the size of the build matrix on appveyor by using a single Cassandra version (3.0.x or 3.x).

jorgebay:
> "should failover after a node goes down" intermittently fails

It is fixed as part of #138.

tolbertam:
> I think running the unit test on Windows against all active LTS releases (0.10, 0.12, 4.x) is necessary but we can limit the size of the build matrix on appveyor by using a single Cassandra version (3.0.x or 3.x).

Sounds good, agree that is a good way to go to get us started.

> It is fixed as part of #138.

Nice!  I'll rebase and see how it goes.  I plan on revisiting this soon.

tolbertam:
This is looking pretty good, just two intermittently failing tests I'd like to nail down before we merge since it'd be nice for this to be relatively rock solid since it marks PRs as failing otherwise:

* `Client #batch(queries, {prepare: 1}, callback) should handle multiple prepares in parallel` - Sometimes fails with `Error: Operation failed - received 1 responses and 2 failures` which is peculiar, will see if I can reproduce locally to understand why C* is generating failures.
* `TimeUuid now() should use current date` - Happened once, timestamps differed by ~15ms, may have just been a VM pause.  We could leave this as is, or implement a retry or something.

jorgebay:
`"Operation failed - received ..."` is a read failure (protocol v4 and above) that occurs on the SELECT statement to assert the result.
I think the size of the test is too much for the hardware: 3 nodes on the same (spinning?) disks, reading 4000 rows on quorum, read timeouts and read failures are likely to happen on modest hardware.

I think we can limit the size of the batch to 120 size each (100 in parallel, plus some more after all queries have been prepared).

tolbertam:
That makes sense.  Although was wondering if maybe we were hitting some kind of bizarre `AssertionError` on the C* side.  I may create an offshoot branch just for running that test and enabling logs somehow to see if we can understand what is going on there.

> I think we can limit the size of the batch to 120 size each (100 in parallel, plus some more after all queries have been prepared).

Will also give that a try.  Thanks!

tolbertam:
Only thing i see in the logs that sticks out is:

> INFO  [ScheduledTasks:1] 2016-04-25 18:08:45,578 MessagingService.java:983 - MUTATION messages were dropped in last 5000 ms: 5 for internal timeout and 0 for cross node timeout
> INFO  [ScheduledTasks:1] 2016-04-25 18:08:45,578 MessagingService.java:983 - BATCH_STORE messages were dropped in last 5000 ms: 3 for internal timeout and 0 for cross node timeout

I'm going to do as you suggested and reduce the size of the batches.

tolbertam:
Removed [wip] tag, I think this is good to go.

build.yaml
jorgebay: I've tried to use v4 and v5 (without minor) on build.yaml file https://github.com/datastax/nodejs-driver/pull/136/commits/a5fc268804d098db0059c13bc7b67560d4c5b11f but it seems that needs settings update on the Jenkins side. I've restored it to use major/minor runtime versions.

tolbertam: Unfortunately we need to update the job creator code to allow those runtimes to be used (would be nice to make that dynamic, i'll look into that).  In near term i'll add those runtimes to job creator so they can be used here.

PR #138, opened 2016-04-14
Replace async dependency

jorgebay:
Created control flow methods and functional utilities in the utils module adapted to our needs:
- `series()` method invokes the final callback (if provided) with the last result.
- `times()`, `timesSeries()` and `timesLimit()` don't accumulate the result of each async execution.

Included micro benchmarks of each method against "async" and "neo-async" modules, where it beats both of them: https://github.com/riptano/nodejs-driver-sut/tree/master/test/micro

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1,7 +1,6 @@
 "use strict";
 var events = require('events');
 var util = require('util');
-var async = require('async');
 
 var utils = require('./utils.js');
 var errors = require('./errors.js');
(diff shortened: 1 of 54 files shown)


jorgebay:
Also fixed reliability of connection pooling test "should failover after a node goes down". Jenkins is looking good.

jorgebay:
CI runs clean, "async" was a crosscutting dependency and by merging this patch into master, we can begin using the new methods. 

We can add more tests to each `utils` function.

PR #140, opened 2016-04-18
NODEJS-250 Log timeout value used onTimeout.

tolbertam:
For [NODEJS-250](https://datastax-oss.atlassian.net/browse/NODEJS-250).


jorgebay:
I've included a test to check that the correct value appears in the logs

tolbertam:
Test looks great, thanks!  👍

PR #141, opened 2016-04-21
Support IPv6 addresses with embedded IPv4 addresses

jorgebay:


test/unit/inet-address-tests.js
```diff
@@ -94,6 +94,19 @@ describe('InetAddress', function () {
...
       assert.strictEqual(val.toString(), '10.11.12.13');
     });
+    it('should parse IPv6 Addresses with Embedded IPv4 Addresses', function () {
+      [
+        ['0:0:0:0:0:FFFF:129.144.52.38',    '00000000000000000000ffff81903426'],
+        ['0:0:0:0:0:0:129.144.52.38',       '00000000000000000000000081903426'],
+        ['::ffff:129.144.52.38',            '00000000000000000000ffff81903426'],
+        ['::ffff:254.255.52.32',            '00000000000000000000fffffeff3420'],
+        ['::13.1.68.3',                     '0000000000000000000000000d014403'],
+        ['fe::13.1.68.3',                   '00fe000000000000000000000d014403']
```
sbisbee: Hey @jorgebay - To my knowledge this isn't valid. `::ffff:` prefix is valid per RFC while the `::` prefix is deprecated but commonly supported. Non-RFC docu: https://en.m.wikipedia.org/wiki/IPv6#IPv4-mapped_IPv6_addresses

lib/types/inet-address.js
```diff
@@ -37,42 +37,55 @@ InetAddress.fromString = function (value) {
...
   }
-  var pattern = /^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$|^[\da-f:]+$/i;
-  if (!pattern.test(value)) {
-    throw new TypeError('Value could not be parsed as InetAddress: ' + value);
-  }
-  var parts = value.split('.');
-  if (parts.length === 4) {
-    //IPv4
+  var ipv4Pattern = /^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$/;
+  var ipv6Pattern = /^[\da-f:.]+$/i;
```
tomc603: This regex will not match v6 addresses with %interface postfix.
Also, and this may be by intent given the 4-in-6 tests below, but this pattern will match v4 addresses.

jorgebay: I'm not sure if we should support interface suffix, or at least not in this patch.
About the ipv6 regex, it will also match ipv4 to handle embedded IPv4 addresses in IPv6 strings, but IPv4 addressed will be handled by `ipv4Pattern.test()` below.

test/unit/inet-address-tests.js
jorgebay: Even though `::` is deprecated its supported in [Java (C*)](http://guava-libraries.googlecode.com/svn/tags/release09/javadoc/com/google/common/net/InetAddresses.html), [other runtimes](http://ideone.com/ndrEDb) and part of the rfc.

Do you see a problem in supporting it or do you recommend that we specifically disallow it?

jorgebay:
Thanks @sbisbee and @tomc603 for taking the time to review the patch, keep it coming!

tolbertam:
Looks good to me 👍 .  One thing I was wondering though was whether or not it was important that the `toString()` method prints the IPV4 part in the case of an embedded address?   It currently doesn't do that.  i.e.:  `::ffff:129.144.52.38` becomes `::ffff:8190:3426`.

I also found a small bug in the toString() method (#144) which is not a regression while debugging this.

jorgebay:
@tolbertam I was not sure either... I see that there is no agreement accross runtimes:
- In Java: `java.net.InetAddress` prints an ipv4 string if its something like: `::ffff:129.144.52.38`, but an ipv6 representation for `::129.144.52.38` (`0:0:0:0:0:0:8190:3426`).
- In .NET: `IPAddress` class maintains the embedded ipv4 notation when calling `ToString()`, if the instance was created that way.
- Python: Behaves the same way as the current implementation of the driver, representing it using ipv6 notation (`::ffff:8190:3426`).

tolbertam:
As there is no agreement between drivers, I think it is reasonable to keep it the way it is (logging the compressed ipv6 address). 👍

test/unit/inet-address-tests.js
sbisbee: @jorgebay Sorry, didn't see the follow up messages but did see the merge. Weird.

Anyway, I would explicitly disallow the `::` case since it's been out of spec for a long time. Since this is new code and not supporting backward behavioral compatibility it seems like a good time to make "stable" choices like this.

If we were following the Java driver for guidance then we'd actually convert mapped addresses with both suffixes to IPv4 according to the (perceived?) original intent of the RFC. That seems even worse to me though because now I've written IPv6 and gotten IPv4 back.

sbisbee:
@jorgebay @tolbertam Two things for `toString()`...

(1) It'd be easier to debug applications if the serialized form looks like what my code sent into the driver. Ex., if I'm trying to trace out a data issue (maybe I set my column type to Int instead of Inet by mistake) then I want to see my representation of the data for as long as possible.

(2) This also gets messy for test frameworks. If my test code writes `::ffff:10.10.10.10` as a String to a table, performs a query, and then invokes `toString()` on the returned value, it won't actually be what I wrote in. I'd prefer to avoid adding an additional column to my tables to distinguish between IPv4 vs IPv6 vs IPv4-mapped-to-IPv6, using it to perform my own serialization of the returned Inet type, since my application and its users care about the serialized String form.

test/unit/inet-address-tests.js
jorgebay: ok, sounds good, its a new implementation and there is no need to implement a deprecated spec.
[RFC 4291 (that obsoletes 3513) states](https://tools.ietf.org/html/rfc4291#section-2.5.5):

> The "IPv4-Compatible IPv6 address" is now deprecated because the
   current IPv6 transition mechanisms no longer use these addresses.

Defining IPv4-mapped as:
```
   |                80 bits               | 16 |      32 bits        |
   +--------------------------------------+--------------------------+
   |0000..............................0000|FFFF|    IPv4 address     |
   +--------------------------------------+----+---------------------+
```

We should explicitly validate anything different. Thanks for insisting on this!

jorgebay:
@sbisbee I think the driver should provide an specified way to represent an inet address as a string and it does: for ipv4, it is traditional ipv4 notation and for a ipv6, it's the ipv6 compressed syntax, ie: `2016::a:1`.

The inet value is sent to C* and retrieved as a byte array, as the rest of the datatypes, in this case of length 4 or 16. We are not able to store how to display it, so in a later time is not possible to determine how it was input.

We can add a third case to the string representation of the `InetAddress` class: when the binary value is `00000000000000000000FFFF` for the first 96 bits, we could represent it as a IPv4-mapped IPv6. What do you think?

In the meantime, I'll open a new pull request for explicitly validating IPv4-mapped IPv6.

jorgebay:
Added validation for IPv4-mapped IPv6 on #152.

PR #143, opened 2016-04-25
Retry policy decision: use current or next host

jorgebay:
Includes `RetryPolicy` jsdoc improvements (NODEJS-225) and new `#onRequestError()` method (NODEJS-251).

### DIFF
--- a/lib/policies/retry.js
+++ b/lib/policies/retry.js
@@ -8,43 +8,55 @@ function RetryPolicy() {
 
 }
 
+//noinspection JSUnusedLocalSymbols
 /**
  * Determines what to do when the drivers gets an UnavailableException from a Cassandra node
- * @param {{request: object, handler: RequestHandler, nbRetry: Number}} requestInfo
- * @param consistency
- * @param {Number} required
- * @param {Number} alive
+ * @param {{request: object, nbRetry: Number}} requestInfo
+ * @param {Number} consistency The [consistency]{@link module:types~consistencies} level of the query that triggered
+ * the exception.
+ * @param {Number} required The number of replicas whose response is required to achieve the
+ * required [consistency]{@link module:types~consistencies}.
+ * @param {Number} alive The number of replicas that were known to be alive when the request had been processed
+ * (since an unavailable exception has been triggered, there will be alive &lt; required)
  * @returns {DecisionInfo}
  */
 RetryPolicy.prototype.onUnavailable = function (requestInfo, consistency, required, alive) {
-  return this.rethrowResult();
+  if (requestInfo.nbRetry > 0) {
+    return this.rethrowResult();
+  }
+  return this.retryResult(undefined, false);
 };
 
 /**
  * Determines what to do when the drivers gets an ReadTimeoutException from a Cassandra node
- * @param {{request: object, handler: RequestHandler, nbRetry: Number}} requestInfo
- * @param consistency
- * @param {Number} received
- * @param blockFor
- * @param {Boolean} isDataPresent
+ * @param {{request: object, nbRetry: Number}} requestInfo
+ * @param {Number} consistency The [consistency]{@link module:types~consistencies} level of the query that triggered
+ * the exception.
+ * @param {Number} received The number of nodes having answered the request.
+ * @param {Number} blockFor The number of replicas whose response is required to achieve the
+ * required [consistency]{@link module:types~consistencies}.
+ * @param {Boolean} isDataPresent When <code>false</code>, it means the replica that was asked for data has not responded.
  * @returns {DecisionInfo}
  */
 RetryPolicy.prototype.onReadTimeout = function (requestInfo, consistency, received, blockFor, isDataPresent) {
   if (requestInfo.nbRetry > 0) {
     return this.rethrowResult();
   }
-  return ((received >= blockFor && !this.isDataPresent) ?
+  return ((received >= blockFor && !isDataPresent) ?
     this.retryResult() :
     this.rethrowResult());
 };
 
 /**
  * Determines what to do when the drivers gets an WriteTimeoutException from a Cassandra node
- * @param {{request: object, handler: RequestHandler, nbRetry: Number}} requestInfo
- * @param consistency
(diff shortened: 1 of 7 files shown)


lib/policies/retry.js
```diff
@@ -56,22 +65,30 @@ RetryPolicy.prototype.onWriteTimeout = function (requestInfo, consistency, recei
 };
 
 /**
- * @returns {{decision: Number}}
+ * Returns a {@link DecisionInfo} to retry the request with the given [consistency]{@link module:types~consistencies}.
+ * @param {Number} [consistency] When specified, it retries the request with the given consistency.
+ * @param {Boolean} [useCurrentHost] When specified, determines if the retry should be made using the same coordinator.
```
tolbertam: Would be good to indicate the default is to not try the next host.

lib/request-handler.js
```diff
@@ -323,24 +323,35 @@ RequestHandler.prototype.handleError = function (err, callback) {
...
   }
-  if (decisionInfo && decisionInfo.decision === retry.RetryPolicy.retryDecision.retry) {
-    return this.retry(callback);
+  if (decisionInfo) {
+    if (decisionInfo.decision === retry.RetryPolicy.retryDecision.retry) {
+      return this.retry(decisionInfo.consistency, decisionInfo.useCurrentHost, callback);
+    }
+    if (decisionInfo.decision === retry.RetryPolicy.retryDecision.ignore) {
+      //Return an empty response
+      return callback(
```
tolbertam: Was there not logic for ignore before, guess never noticed :)

lib/request-handler.js
```diff
@@ -351,12 +362,25 @@ RequestHandler.prototype.handleError = function (err, callback) {
...
+  }
   if (this.retryHandler) {
+    //custom retry handler (not a QueryRequest / ExecuteRequest / BatchRequest)
     return this.retryHandler(callback);
   }
+  if (useCurrentHost !== false) {
+    // retry on the current host by default
+    return this.sendOnConnection(this.request, this.requestOptions, callback);
+  }
   this.send(this.request, this.requestOptions, callback);
```
tolbertam: Shouldn't this `this.send` not be called if useCurrentHost is true?

tolbertam:
To be consistent with the java driver, should we update RetryPolicy to retry once on next host on unavailable by default?  ([java driver code that does this](https://github.com/datastax/java-driver/blob/3.0.1/driver-core/src/main/java/com/datastax/driver/core/policies/DefaultRetryPolicy.java#L108-L113)).  This was done as part of [JAVA-709](https://datastax-oss.atlassian.net/browse/JAVA-709) to address the possible case where a node gets partitioned from the rest of the cluster, but the client can still talk to it (seems unlikely, but it can happen).

jorgebay:
> Should we update RetryPolicy to retry once on next host on unavailable by default?

I wasn't aware of that change, I'll fix it!

Also, I think that it makes sense to include [NODEJS-251](https://datastax-oss.atlassian.net/browse/NODEJS-251) in this patch, to group all the upcoming changes related to the `RetryPolicy`. What do you think?

tolbertam:
> Also, I think that it makes sense to include NODEJS-251 in this patch, to group all the upcoming changes related to the RetryPolicy. What do you think?

I agree, this is a good place to put that. 👍

lib/request-handler.js
jorgebay: I'm not sure if I understand the question, `#sendOnConnection ()` uses the current connection and `#send()` uses the next host in the query plan.

tolbertam: Ah, when I made the comment I must not have noticed the `return` in the sendOnConnection call, in that case this is fine.

lib/request-handler.js
```diff
@@ -295,71 +300,109 @@ RequestHandler.prototype.prepareOnConnection = function (queries, callbacksArray
...
+/**
+ * @returns {{decision, useCurrentHost, consistency}}
+ */
+RequestHandler.prototype.getDecision = function (err) {
+  var requestInfo = {
+    request: this.request,
+    handler: this,
+    nbRetry: this.retryCount
+  };
+  //TODO: isServerUnhealthy is going to be called isSocketError as part of NODEJS-106 and should not call setDown()
```
jorgebay: We should revisit this pull request once NODEJS-106 is merged, as it affects how socket error flag is named and the behaviour when there is a socket error.

jorgebay:
- Included `onRequestError()` method in the retry policy.
- Added the flag `wasRequestWritten` when we suspect that it could have been written in the wire (and mutation could be applied). We retry on next host if we know for a fact that the request was not written and rely on the retry policy otherwise. 
- Rebased against master (after #145 was merged).
- The retry policy behaviour is supposed to be the same as the `DefaultRetryPolicy` in the java driver 3.0+.

tolbertam:
I wrote some tests to validate this using [stubbed cassandra](http://www.scassandra.org/).  Since the test requires standing stubbed cassandra clusters I put it on a [separate branch](https://github.com/tolbertam/nodejs-driver/tree/NODEJS-105-stubbed) with the eventual goal to integrate it into our test automation.  A run down of what is validated:

```
  RetryPolicy
    #onUnavailable()
      ✓ should retry on the next host the first time (50ms)
      ✓ should rethrow the following times
    #onRequestError()
      ✓ should retry on the next host until successful (2033ms)
      ✓ should rethrow when retryOnTimeout is false (1019ms)
      ✓ should retry on the next host until successful if connection closed on request
      ✓ should retry on the next host until successful if server error
```

Some of these tests are modeled after the unit tests (just at an integration level).  Would appreciate any early feedback on that branch as I intend to build up a stubbed suite for this and also #145 among other things.

The only thing I haven't been able to figure out how to exercise is when wasRequestWritten != true (which you have a unit test for via `should retry when there was a socket error and mutation was not applied`)

All that being said, this looks good to me. 👍

jorgebay:
Big +1 to tests with stubbed cassandra, I've commented on some minor improvements on the commit.

PR #144, opened 2016-05-03
Fix case in InetAddress.toString() that truncates last group.

tolbertam:
For [NODEJS-255](https://datastax-oss.atlassian.net/browse/NODEJS-255)

### DIFF
--- a/lib/types/inet-address.js
+++ b/lib/types/inet-address.js
@@ -147,12 +147,16 @@ InetAddress.prototype.toString = function (encoding) {
       if (start < 0) {
         start = i;
       }
+
+      // at the end of the buffer, make a final call to checkLongest.
+      if(i === this.buffer.length - 2) {
+        checkLongest(i+2);
+      }
       continue;
     }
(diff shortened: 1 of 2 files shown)


lib/types/inet-address.js
```diff
@@ -147,12 +147,16 @@ InetAddress.prototype.toString = function (encoding) {
       if (start < 0) {
         start = i;
       }
+
+      // at the end of the buffer, make a final call to checkLongest.
+      if(i == this.buffer.length - 2) {
```
jorgebay: Use strict equality `===` :)

jorgebay:
Nice catch!
Minor nit about the operator, otherwise looks good to me.

lib/types/inet-address.js
tolbertam: d'oh!  Keep making that mistake...will fix shortly.

tolbertam: Fixed and squashed, thanks for pointing that out :)

PR #147, opened 2016-05-10
Use all IPv4 and IPv6 addresses from dns records

jorgebay:


tolbertam:
This works great 👍.  I set up dnsmasq with a hostname that had multiple A and AAAA records.   One thing I was worried about was whether or not the hosts would be retained if they weren't in `system.peers`, that does seem to be the case which is good (way its always been).  It looks like records not found in `system.peers` are discarded as desired.  This is important for the case of 'localhost' particularly since its typical for localhost to resolve to 127.0.0.1 and ::1 and less common for C* to be bound on the ipv6 interface.

In my case, dnsmasq always randomizes the A / AAAA records it returns, but the AAAA records are always tried last since they are fetched after the A records, but that's not really a big deal.  It would be nice to randomize the order the contact points are tried ([NODEJS-104](https://datastax-oss.atlassian.net/browse/NODEJS-104)) as to put less burden on one particular contact point in the case the DNS server always returns in the same order and a user has a ton of clients.

PR #148, opened 2016-05-11
Avoid using Object.defineProperty() in ResultSet constructor

jorgebay:


tolbertam:
Looks great.  This shows a very noticeable performance improvement (~9-9.5k writes/sec vs. ~7-7.5k writes/sec in a throughput test).  Surprising at first glance but there are a lot of references on the web about defineProperty being awfully slow, so avoiding it where can looks like a great approach. 👍

PR #149, opened 2016-05-13
Debounce protocol events

jorgebay:
Debounce protocol events using a sliding delay. Allow disabling metadata retrieval.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -22,7 +22,11 @@ var clientOptions = require('./client-options');
  *  Only one contact point is required (the driver will retrieve the address of the other nodes automatically),
  *  but it is usually a good idea to provide more than one contact point, because if that single contact point is unavailable, the driver will not be able to initialize correctly.
  * </p>
- * @property {String} keyspace
+ * @property {String} keyspace The logged keyspace for all the connections created within the {@link Client} instance.
+ * @property {Number} refreshSchemaDelay The default window size in milliseconds used to debounce node list and schema
+ * refresh metadata requests. Default: 1000.
+ * @property {Boolean} isMetadataSyncEnabled Determines whether client-side schema metadata retrieval and update is
+ * enabled. Default: true.
  * @property {Object} policies
  * @property {LoadBalancingPolicy} policies.loadBalancing The load balancing policy instance to be used to determine the coordinator per query.
  * @property {RetryPolicy} policies.retry The retry policy.
@@ -170,13 +174,14 @@ var clientOptions = require('./client-options');
  * since it is always possible to use a single session with fully qualified table name
  * in queries.
  * </p>
+ * @extends EventEmitter
  * @param {ClientOptions} options The options for this instance.
  * @constructor
  */
 function Client(options) {
   events.EventEmitter.call(this);
   this.options = clientOptions.extend({ logEmitter: this.emit.bind(this) }, options);
-  this.controlConnection = new ControlConnection(this.options);
+  Object.defineProperty(this, 'controlConnection', { value: new ControlConnection(this.options), enumerable: false, writable: true });
   //Unlimited amount of listeners for internal event queues by default
   this.setMaxListeners(0);
   this.connected = false;
@@ -482,25 +487,30 @@ Client.prototype.log = utils.log;
  * @param {Function} [callback]
  */
 Client.prototype.shutdown = function (callback) {
-  //Go through all the host and shut down their pools
-  this.log('info', 'Shutting down');
-  if (!callback) {
-    callback = function() {};
+  var self = this;
+  function doShutdown() {
+    self.connected = false;
+    self.isShuttingDown = true;
+    var hosts = self.hosts.values();
+    self.controlConnection.shutdown();
+    // go through all the host and shut down their pools
+    utils.each(hosts, function (h, next) {
+      h.shutdown(false, next);
+    }, callback);
   }
-  if (!this.hosts) {
-    //not initialized
+  this.log('info', 'Shutting down');
+  callback = callback || utils.noop;
+  if (!this.hosts || !this.connected) {
+    // not initialized
     this.connected = false;
     return callback();
   }
-  this.isShuttingDown = true;
-  var self = this;
-  var hosts = this.hosts.values();
-  utils.each(hosts, function (h, next) {
-    h.shutdown(false, next);
-  }, function(err) {
-    self.connected = !!err;
-    callback(err);
-  });
+  if (this.connecting) {
+    this.log('warning', 'Shutting down while connecting');
+    // wait until finish connecting for easier troubleshooting
+    return this.once('connected', doShutdown);
+  }
+  doShutdown();
 };
 
 /**
@@ -509,7 +519,7 @@ Client.prototype.shutdown = function (callback) {
  * @param {Function} callback
  * @ignore
  */
-Client.prototype.waitForSchemaAgreement = function (connection, callback) {
+Client.prototype._waitForSchemaAgreement = function (connection, callback) {
   if (this.hosts.length === 1) {
     return setImmediate(callback);
   }
@@ -550,6 +560,29 @@ Client.prototype.waitForSchemaAgreement = function (connection, callback) {
       setTimeout(next, 500);
     });
   }, callback);
+};
+
+/**
+ * Waits for schema agreements and schedules schema metadata refresh.
+ * @param {Connection} connection
+ * @param event
+ * @param {Function} callback
+ * @ignore
+ * @internal
+ */
+Client.prototype.handleSchemaAgreementAndRefresh = function (connection, event, callback) {
+  var self = this;
+  this._waitForSchemaAgreement(connection, function agreementCb(err) {
+    if (err) {
+      //we issue a warning but we continue with the normal flow
+      self.log('warning', 'There was an error while waiting for the schema agreement between nodes', err);
+    }
+    if (!self.options.isMetadataSyncEnabled) {
+      return callback();
+    }
+    // schedule metadata refresh immediately and the callback will be invoked once it was refreshed
+    self.controlConnection.handleSchemaChange(event, true, callback);
+  });
 };
 
 /**
--- a/lib/client.js
+++ b/lib/client.js
@@ -26,7 +26,12 @@ var clientOptions = require('./client-options');
  * @property {Number} refreshSchemaDelay The default window size in milliseconds used to debounce node list and schema
  * refresh metadata requests. Default: 1000.
  * @property {Boolean} isMetadataSyncEnabled Determines whether client-side schema metadata retrieval and update is
- * enabled. Default: true.
+ * enabled.
+ * <p>Setting this value to <code>false</code> will cause keyspace information not to be automatically loaded, affecting
+ * replica calculation per token in the different keyspaces. When disabling metadata synchronization, use
+ * [Metadata.refreshKeyspaces()]{@link module:metadata~Metadata#refreshKeyspaces} to keep keyspace information up to
+ * date or token-awareness will not work correctly.</p>
+ * Default: <code>true</code>.
  * @property {Object} policies
  * @property {LoadBalancingPolicy} policies.loadBalancing The load balancing policy instance to be used to determine the coordinator per query.
  * @property {RetryPolicy} policies.retry The retry policy.
--- a/lib/control-connection.js
+++ b/lib/control-connection.js
@@ -8,6 +8,7 @@ var errors = require('./errors');
 var Host = require('./host').Host;
 var HostMap = require('./host').HostMap;
 var Metadata = require('./metadata');
+var EventDebouncer = require('./metadata/event-debouncer');
 var requests = require('./requests');
 var utils = require('./utils');
 var types = require('./types');
@@ -24,6 +25,7 @@ var schemaChangeTypes = {
 /**
  * Represents a connection used by the driver to receive events and to check the status of the cluster
  * @param {Object} options
+ * @extends EventEmitter
  * @constructor
  */
 function ControlConnection(options) {
@@ -54,6 +56,7 @@ function ControlConnection(options) {
    * @type {Encoder|null}
    */
   this.encoder = null;
+  this.debouncer = new EventDebouncer(options.refreshSchemaDelay, this.log.bind(this));
 }
 
 util.inherits(ControlConnection, events.EventEmitter);
@@ -272,20 +275,6 @@ ControlConnection.prototype.refreshOnConnection = function (firstTime, callback)
       self.host.once('down', self.hostDownHandler.bind(self));
       self.refreshHosts(firstTime, next);
     },
-    function getKeyspaces(next) {
-      //to acquire metadata we need to specify the cassandra version
-      self.metadata.setCassandraVersion(self.host.getCassandraVersion());
-      self.metadata.refreshKeyspaces(next);
-    },
-    function buildTokenPerReplica(next) {
-      if (self.metadata.tokenizer) {
-        self.metadata.buildTokens(self.hosts);
-      }
-      else {
-        self.log('warning', 'Tokenizer could not be determined');
-      }
-      next();
-    },
     function subscribeConnectionEvents(next) {
       c.on('nodeTopologyChange', self.nodeTopologyChangeHandler.bind(self));
       c.on('nodeStatusChange', self.nodeStatusChangeHandler.bind(self));
@@ -305,7 +294,7 @@ ControlConnection.prototype.refreshOnConnection = function (firstTime, callback)
 };
 
 /**
- * Gets the info from local and peer metadata column families
+ * Gets the info from local and peer metadata, reloads the keyspaces metadata and rebuilds tokens.
  * @param {Boolean} newNodesUp
  * @param {Function} [callback]
  */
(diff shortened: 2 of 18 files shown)


lib/client.js
```diff
@@ -22,7 +22,11 @@ var clientOptions = require('./client-options');
  *  Only one contact point is required (the driver will retrieve the address of the other nodes automatically),
  *  but it is usually a good idea to provide more than one contact point, because if that single contact point is unavailable, the driver will not be able to initialize correctly.
  * </p>
- * @property {String} keyspace
+ * @property {String} keyspace The logged keyspace for all the connections created within the {@link Client} instance.
+ * @property {Number} refreshSchemaDelay The default window size in milliseconds used to debounce node list and schema
+ * refresh metadata requests. Default: 1000.
+ * @property {Boolean} isMetadataEnabled Determines whether client-side schema metadata retrieval and update is enabled.
```
tolbertam: Oh neat, so if you have this disabled, you can still request schema metadata on demand?  That is really cool.

lib/client.js
```diff
@@ -509,7 +519,7 @@ Client.prototype.shutdown = function (callback) {
  * @param {Function} callback
  * @ignore
  */
-Client.prototype.waitForSchemaAgreement = function (connection, callback) {
+Client.prototype._waitForSchemaAgreement = function (connection, callback) {
```
tolbertam: Isn't this a breaking change?

lib/control-connection.js
```diff
@@ -429,40 +428,93 @@ ControlConnection.prototype.nodeStatusChangeHandler = function (event) {
...
+  var self = this;
+  var handler, cqlObject;
+  if (event.isKeyspace) {
+    if (event.schemaChangeType === schemaChangeTypes.dropped) {
+      handler = function removeKeyspace() {
+        // if on the same event queue there is a creation, this handler is not going to be executed
+        // it is safe to remove the keyspace metadata
+        delete self.metadata.keyspaces[event.keyspace];
+      };
+      return this.scheduleObjectRefresh(handler, event.keyspace, null);
```
tolbertam: Just curious if there is a difference between doing scheduleObjectRefresh and scheduleKeyspaceRefresh below?  Might be some small detail i'm missing.

lib/client.js
jorgebay: It was marked as `@ignore` so it was not part of the public API.

lib/control-connection.js
tolbertam: Ah i see, scheduleObjectRefresh won't trigger a refresh of the keyspace which isn't needed here since it's being deleted.

lib/metadata/event-debouncer.js
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * Adds a new event to the queue and moves the delay.
+ * @param {{ handler: Function, all: boolean|undefined, keyspace: String|undefined,
+ *  cqlObject: String|undefined, callback: Function|undefined }} event
+ */
+EventDebouncer.prototype.eventReceived = function (event) {
+  event.callback = event.callback || utils.noop;
+  this._queue = this._queue || { callbacks: [], keyspaces: {} };
+  //TODO: Warn when size is too large
```
tolbertam: In the java-driver there is a differentiation between max pending and max queue size, and the max pending is configurable.  I'm not sure if we need that level of configurability, but I agree that a warning on max queue depth is probably a good idea.

tolbertam: Another difference here is that a single event debouncer is shared between all kinds of events (schema refreshes and host refreshes).  I like this, keeps things much simpler, but now you can't have different interval configurations between the two kind of events, which isn't that important (I haven't heard of cases of people configuring those differently yet).

The java driver also debounces node refresh requests, which are triggered as result of a `STATUS_CHANGE` event.  This prevent a burst of UP/DOWN/UP from marking up down and up again, and instead just up.  Do you think there is any value in debouncing that?  It seems less important that node list / schema refresh.  Also the java driver refreshes the node info from peers on these events and the node driver does not (from what I can tell).

test/unit/event-debouncer-tests.js
```diff
@@ -0,0 +1,157 @@
+"use strict";
```
tolbertam: very nice tests

lib/client.js
tolbertam: Hrmm, it looks like metadata is still fetched on initialization and establishment of a control connection via `ControlConnection#refreshHosts`

tolbertam: With respect to my comment about 'request schema metadata on demand' i don't think that will work because the driver will currently only fetch the table if it knows the keyspace is there, which is wouldn't since the keyspace wouldn't be known (although right now it will since the metadata is grabbed on control connection).

Also it currently wouldn't be possible to know if a table is stale in the local metadata or not, so that approach wouldn't work either.

I think the best route would be for `isMetadataEnabled` to never fetch schema metadata (which means not building token map either).

test/integration/short/control-connection-tests.js
```diff
@@ -30,13 +30,16 @@ describe('ControlConnection', function () {
...
       var otherClient = new Client(helper.baseOptions);
       utils.series([
         cc.init.bind(cc),
-        helper.toTask(otherClient.execute, otherClient, "CREATE KEYSPACE sample_change_1 WITH replication = {'class': 'SimpleStrategy', 'replication_factor' : 3}"),
+        helper.toTask(otherClient.execute, otherClient, "CREATE KEYSPACE sample_change_1 WITH replication = " +
+          "{'class': 'SimpleStrategy', 'replication_factor' : 3}"),
         function (next) {
-          setTimeout(next, 500);
+          helper.setIntervalUntil(function () {
+            return cc.metadata.keyspaces['sample_change_1'];
```
tolbertam: This is pretty smart, much better than sleeping :)

lib/client.js
```diff
@@ -553,6 +563,29 @@ Client.prototype.waitForSchemaAgreement = function (connection, callback) {
...
+Client.prototype.handleSchemaChange = function (connection, event, callback) {
+  var self = this;
+  this._waitForSchemaAgreement(connection, function (err) {
+    if (err) {
+      //we issue a warning but we continue with the normal flow
+      self.log('warning', 'There was an error while waiting for the schema agreement between nodes', err);
+    }
+    if (self.options.isMetadataEnabled) {
+      // schedule metadata refresh but do not wait for it to finish (no callback)
+      self.controlConnection.handleSchemaChange(event);
```
tolbertam: Two possible issues here:

1. if the schema does not agree, it will still call `handleSchemaChange` which will be invoked on the control connection which will debounce and fetch the schema on the control connection.  Its possible that the control connection has a stale view of the data as the schema change wasn't fully propagated to it yet.
2. The callback is invoked because the metadata is updated via handleSchemaChange, so if someone goes to grab the metadata immediately after their request completes they may not get the most recent data yet, as it hasn't been fetched yet (still in the debouncer).

lib/client.js
tolbertam: > It was marked as @ignore so it was not part of the public API.

oh hah, there is is literally 2 lines above :D

lib/client.js
tolbertam: Was just looking at the java driver to see what it does in this case (query response type is a schema change). It **does** wait for the updated schema to be fetched, but it is debounced, which delays the response by the debouncer interval ([JAVA-1120](https://datastax-oss.atlassian.net/browse/JAVA-1120)).  In this case the debouncing isn't necessary because the client was the one making the change, and it's very likely/good practice to make schema changes from one client.  With debouncing it would delay each schema change by 1 second, which is not ideal if making a bunch of schema changes, so I think it would be nice if we put that detection in here as well so it does not debounce if the client was the one that made the change.

lib/client.js
jorgebay: As you point out, it looks like the `isMetadataEnabled` flag implementation is incorrect.

When automatic metadata retrieval is not enabled the driver: 

1. Should NOT retrieve keyspaces metadata.
1. Should NOT cache results of calls to get metadata of a schema object,  ie: `Metadata#getTable()` should always get the latest info.
1. Should get topology information (on init and up to date based on topology events).
1. Should build tokens: [tokens are retrieved and primary replicas][1] are populated but keyspace-specific replicas per token are [lazily calculated when the first call to `getReplicas()` for a given keyspace is made][2]. This could still be valid with metadata retrieval is disabled.

`Metadata#refreshKeyspaces()` is a public method so users can call it manually if they want to get/refresh the keyspaces information.

Let me know if I you agree with the above reasoning. In the meantime, I'll push a new commit to try it out.

[1]: https://github.com/datastax/nodejs-driver/blob/74a48f7bf7f786041615eea6bb70e1ae494c5b9a/lib/metadata/index.js#L90
[2]: https://github.com/datastax/nodejs-driver/blob/74a48f7bf7f786041615eea6bb70e1ae494c5b9a/lib/metadata/index.js#L202

lib/control-connection.js
jorgebay: Additionally, if there is succession of events in the following order: "DROPPED" -> "CREATED" for a given keyspace, we make sure that `scheduleKeyspaceRefresh()` (caused by the "CREATED") will supersede the `delete` call.

lib/metadata/event-debouncer.js
jorgebay: After NODEJS-106, `STATUS_CHANGE` events have no significant effect in the driver.

I think sharing a "single" debouncer helps setting a hierarchy: if the hosts + keyspaces are metadata is refreshed, there is no point in refreshing a single keyspace / table.

lib/client.js
jorgebay: I see... Nice catch!

The solution would be to queue it up with the rest of schema changes and process the queue immediately.

jorgebay:
I've added 2 commits to fix:

- Retrieve metadata (process queue) immediately after a schema change response.
- Check `isMetadataEnabled` flag to refresh keyspaces after refreshing topology info. Avoid caching results of user metadata queries when `isMetadataEnabled` is false, according to [this thread](https://github.com/datastax/nodejs-driver/pull/149#discussion_r63838465).

lib/client.js
tolbertam: I like what you've outlined.  With those conditions, schema metadata will not be touched unless the user explicitly goes out and requests it, nice.   I'll experiment with it.

tolbertam: On another note, `isMetadataEnabled` could probably be renamed to better represent what it's doing, something like `isMetadataSyncEnabled` or something like that, although i'm not too worried about it.

lib/metadata/event-debouncer.js
tolbertam: > After NODEJS-106, STATUS_CHANGE events have no significant effect in the driver.

That's a good point, in fact the same applies to the java driver as well now, so that might be a good consideration there as well.

> I think sharing a "single" debouncer helps setting a hierarchy: if the hosts + keyspaces are metadata is refreshed, there is no point in refreshing a single keyspace / table.

I agree, makes sense and it is a lot easier to conceptualize vs. having separate debouncers on their own timers.

lib/client.js
jorgebay: I took the name from the java driver but Im not very fond of it neither, `isMetadataSyncEnabled` sounds better, I'll change it.

lib/client.js
tolbertam: Looks good, I think that solution solves 2 very well.  1 doesn't seem easily solveable (java driver has the same problem).  If the schema doesn't agree, the control node may not have the schema when it fetches it, however by the time the schema does agree, we'll get a schema_change even over the control connection, which will get us back up to date.  So I think the only weakness there is the possibility that the metadata in the driver is wrong immediately after the query completes, but I think that's ok given that the schema doesn't agree and when it does agree, it'll be up to date.

jorgebay:
Renamed `isMetadataEnabled` to `isMetadataSyncEnabled`.

tolbertam:
Is it required to call anything before fetching metadata is `isMetadataSyncEnabled` set to false?  Tried retrieving a materialized view with it set up that way and I get the following error:

```
     Uncaught Error: Materialized views are not supported on Cassandra versions below 3.0
      at NotSupportedError.DriverError (lib/errors.js:14:19)
      at new NotSupportedError (lib/errors.js:117:28)
      at SchemaParserV1.getMaterializedView (lib/metadata/schema-parser.js:524:12)
      at Metadata.getMaterializedView (lib/metadata/index.js:403:22)
      at test/integration/short/metadata-tests.js:862:29
      at Object.each (lib/utils.js:422:5)
      at Array.checkAllView (test/integration/short/metadata-tests.js:861:17)
      at next [as _onTimeout] (lib/utils.js:646:17)
```

For some reason it's using SchemaParserV1, but I do see in code that it does set the SchemaParser to V3, i'll keep digging.

tolbertam:
Ah i see,  I need to call `metadata.refreshKeyspaces` prior, that seems to get me further.   I think i remember you previously mentioning this (otherwise I probably would have not thought to try it).  I'll keep working at it.

tolbertam:
Nice, all the schema calls i've made so far seem to work with `isMetadataSyncEnabled` set to false.  The `refreshKeyspaces` step isn't needed either.  Super cool.  I'll have some test changes coming but so far so good.

tolbertam:
Question about number 4 from [your list of assertions](https://github.com/datastax/nodejs-driver/pull/149#discussion_r63838465) when automatic metadata retrieval is disabled:

> Should build tokens: tokens are retrieved and primary replicas are populated but keyspace-specific replicas per token are lazily calculated when the first call to getReplicas() for a given keyspace is made. This could still be valid with metadata retrieval is disabled.

I think as things currently are this won't work, the reason being because `Metadata.getReplicas` checks keyspaces, but they may not be there / may be out of sync.   Should it be required for the user to call `refreshKeyspaces` for this to work?   In that case it should work if you call it explicitly before hand.   I added a test in bc0711d that currently fails since it doesn't call refreshKeyspaces at the moment (but we can change it to do so).

test/integration/short/metadata-tests.js
```diff
@@ -147,32 +178,39 @@ describe('Metadata', function () {
...
+      var nonSyncClient = newInstance({ isMetadataSyncEnabled: false });
+      var clients = [client, nonSyncClient];
       utils.series([
         client.connect.bind(client),
-        helper.toTask(client.execute, client, helper.createKeyspaceCql('ks_udt_meta', 3)),
+        nonSyncClient.connect.bind(nonSyncClient),
+        helper.toTask(client.execute, client, helper.createKeyspaceCql('ks_udt_meta', 2)),
         helper.toTask(client.execute, client, 'USE ks_udt_meta'),
         helper.toTask(client.execute, client, 'CREATE TYPE type_changing (id uuid, name ascii)'),
-        helper.delay(1000),
```
tolbertam: A nice side effect of the change to not complete the request until the schema is updated is we shouldn't have to have these delays anymore, but will see if that is truly the case via CI.

jorgebay:
`Metadata#buildTokens()` populates the token ring in order and the primary replicas.
`Metadata#getReplicas()` uses that information to calculate (once) all the replicas per token. If the keyspace metadata is not retrieved / up to date it wont work.

I think that behaviour is correct, disabling metadata sync should not be very common and should be for users that understand what they are doing. Maybe we can document somewhere (in the `isMetadataSyncEnabled` flag?) that disabling it would require users to call `refreshKeyspaces()` manually to get the proper replicas information.

Otherwise, we could disable avoid calling `buildTokens()` altogether.

tolbertam:
> Metadata#buildTokens() populates the token ring in order and the primary replicas.
Metadata#getReplicas() uses that information to calculate (once) all the replicas per token. If the keyspace metadata is not retrieved / up to date it wont work.

I see, so the token data is built, but the keyspace information isn't.

> I think that behaviour is correct, disabling metadata sync should not be very common and should be for users that understand what they are doing. Maybe we can document somewhere (in the isMetadataSyncEnabled flag?) that disabling it would require users to call refreshKeyspaces() manually to get the proper replicas information.

That sounds fair to me.  I think if we make it clear in the documentation that you are responsible for refreshing keyspaces or alternatively not using Token Aware, that enables a user to still use the feature if they know what they are doing, which like you mention they should probably be an advanced user to use the option in the first place.

If you don't happen to call refreshKeyspaces though what happens?  I see that in `getReplicas` that if the ring is null it returns null, but if the keyspace doesn't exist, it finds the closest replica from the token, which may not be the best if you are using a multi-DC replication strategy.   However I think TokenAware will not prioritize the replica if it isn't local right?  So that may work in our favor.   If that is true, the only weakness of returning the closest replica is that it only returns the one host which could hotspot load to the one node, but I think that is a contained enough problem.  

If i'm right about the above I think we should make mention that the behavior in this case is that it will return the closest replica.

jorgebay:
About `Metadata#getReplicas()` I was thinking about it too, I think the safest way would be to return `null` when the keyspace is not found and differentiate the behaviour when the keyspace is not specified (return closest) and not found (return `null` as metadata was not obtained).

tolbertam:
>  I think the safest way would be to return null when the keyspace is not found and differentiate the behaviour when the keyspace is not specified (return closest) and not found (return null as metadata was not obtained).

Hrmm, that does make a lot of sense, 👍 .  This way if the user did not know to refresh the keyspaces manually, the penalty is not really as risky as returning only 1 replica (and possibly hotspotting), instead you just don't get Token Aware.  I like that idea a lot!

tolbertam:
I've updated the tests to reflect the behavior we discussed.  If the keyspace metadata is present (via `refreshKeyspaces` called prior), we expect the correct replicas.   If it is not, it returns null.   The test `should return null if keyspace metadata is not present` will fail until `getReplicas` is updated to return null if keyspace metadata is not present.

jorgebay:
I've updated the `getReplicas()` to return `null` and the docs.

Its on 90b97531b13c89487822bcb5a66cfed57580bd0b but github doesn't show it (seems to be having issues)...

tolbertam:
Looks good, 👍  on this PR :)

jorgebay:
Squashed into 2 commits, will `ff-only` merge once CI finishes.

PR #151, opened 2016-05-17
Handle empty map values gracefully

jorgebay:


tolbertam:
Can't really find a good explicit way to reproduce a null value in the place of a map value, although I know it is possible through older versions.  Fix looks pretty straightforward, so lgtm 👍

PR #152, opened 2016-05-18
InetAddress: validate for IPv4-mapped IPv6

jorgebay:


jorgebay:
Explicitly validate IPv4-mapped IPv6, according to feedback from #141.

tolbertam:
This works great 👍.  I set up dnsmasq with a hostname that had multiple A and AAAA records.   One thing I was worried about was whether or not the hosts would be retained if they weren't in `system.peers`, that does seem to be the case which is good (way its always been).  It looks like records not found in `system.peers` are discarded as desired.  This is important for the case of 'localhost' particularly since its typical for localhost to resolve to 127.0.0.1 and ::1 and less common for C* to be bound on the ipv6 interface.

In my case, dnsmasq always randomizes the A / AAAA records it returns, but the AAAA records are always tried last since they are fetched after the A records, but that's not really a big deal.  It would be nice to randomize the order the contact points are tried ([NODEJS-104](https://datastax-oss.atlassian.net/browse/NODEJS-104)) as to put less burden on one particular contact point in the case the DNS server always returns in the same order and a user has a ton of clients.

tolbertam:
Woops, ignore previous comment, meant for other PR.

test/unit/inet-address-tests.js
```diff
@@ -97,19 +97,27 @@ describe('InetAddress', function () {
...
-        ['fe::13.1.68.3',                   '00fe000000000000000000000d014403']
+        ['::ffff:254.255.52.32',            '00000000000000000000fffffeff3420']
       ].forEach(function (item) {
         var ip = InetAddress.fromString(item[0]);
         assert.strictEqual(ip.toString('hex'), item[1]);
       });
     });
+    it('should throw TypeError for invalid IPv6 address with embedded IPv4 address', function () {
+      [
+        ['0:0:0:0:0:0:129.144.52.38'],
```
tolbertam: Shouldn't this one (`0:0:0:0:0:0:129.144.52.38`) be technically valid based on 2.5.5.1, or are we not supporting it because it is considered deprecated?

tolbertam: Ah, i see from [this discussion](https://github.com/datastax/nodejs-driver/pull/141#discussion_r62544720) that we do intend not to support this.

tolbertam:
lgtm 👍

PR #153, opened 2016-05-18
Fix paging test: use consistency level quorum

jorgebay:


test/integration/short/client-each-row-tests.js
```diff
@@ -263,12 +263,12 @@ describe('Client', function () {
         client.shutdown.bind(client)
       ], done);
     });
-    vit('2.0', 'should autoPage on parallel different tables', function (done) {
+    for (var i = 0; i < 50; i++)
```
tolbertam: Is the for loop intentional?

jorgebay: Yes, it was. I was seeing fails on appveyor related to this test that occurred from time to time. It was indeed related to the consistency level, as the results are clean.

Sorry I didnt mark this pr as WIP.

tolbertam: Ah ok cool, makes sense 👍

PR #155, opened 2016-06-01
Execution profiles

jorgebay:
Usage:

```javascript
const metricsProfile = new ExecutionProfile({
  consistency: consistency.quorum,
  readTimeout: 30000
});
const client = new Client({ 
  contactPoints: ['host1'], 
  profiles: { 
    'metrics': metricsProfile 
  }
});

// Using execution profile name
client.execute(query, params, { executionProfile: 'metrics' }, callback);

// Using execution profile instance
client.execute(query, params, { executionProfile: metricsProfile }, callback);

// Using default execution profile (overriding a setting)
client.execute(query, params, { consistency: consistency.one }, callback);
```

PR #156, opened 2016-06-01
Execution profiles and query options fixed properties

jorgebay:
Contains execution profiles NODEJS-261 (#155) and creation of the query options using fixed property names (no dynamic shallow copying) NODEJS-201.

Usage:

```javascript
const metricsProfile = new ExecutionProfile({
  consistency: consistency.quorum,
  readTimeout: 30000
});
const client = new Client({ 
  contactPoints: ['host1'], 
  profiles: { 
    'metrics': metricsProfile 
  }
});

// Using execution profile name
client.execute(query, params, { executionProfile: 'metrics' }, callback);

// Using execution profile instance
client.execute(query, params, { executionProfile: metricsProfile }, callback);

// Using default execution profile (overriding a setting)
client.execute(query, params, { consistency: consistency.one }, callback);
```


jorgebay:
Thanks to NODEJS-201, I'm getting a 10% improvement in throughput rate on my local machine (from ~18,000 r/s to ~20.000 r/s) compared to master.

lib/client.js
```diff
@@ -913,28 +918,24 @@ Client.prototype._setHostListeners = function () {
...
-    if (err) return callback(err);
-    var hosts = utils.iteratorToArray(iterator);
-    utils.timesLimit(hosts.length, 32, function warmupEachCallback(i, next) {
-      var h = hosts[i];
-      var distance = h.getDistance(loadBalancingPolicy);
-      if (distance !== types.distance.local) {
-        //do not warmup pool for remote or ignored hosts
-        return next();
+  var hosts = this.hosts.values();
+  utils.timesLimit(hosts.length, 32, function warmupEachCallback(i, next) {
```
tolbertam: Not related, but I've always wondered, why 32?   Is it just an arbitrary number to get a decent amount of connections spun up in the poo.

jorgebay: Good question, I arbitrarily used `32` when implementing NODEJS-138 to limit parallelism, but any low number would do.
I'll set it in a constant to avoid having to think about `32` and improve readability.

lib/control-connection.js
```diff
@@ -25,10 +25,11 @@ var schemaChangeTypes = {
 /**
  * Represents a connection used by the driver to receive events and to check the status of the cluster
  * @param {Object} options
+ * @param {ProfileManager} profileManager
  * @extends EventEmitter
  * @constructor
  */
-function ControlConnection(options) {
+function ControlConnection(options, profileManager) {
```
tolbertam: `this.loadBalancingPolicy` doesn't pull from `profileManager`, although it looks like it isn't used anymore so could just be removed.

jorgebay: My intention was to use `this.profileManager.getDefaultLoadBalancingPolicy()` but it looks I unintentionally kept the reference to `options.policies.loadBalancing` on line 43.

The question is: should we use the lb policy defined in the options (`options.policies.loadBalancing`) or the one in the default profile? Most of the time should be the same... 
But if not, what does it mean to have different lb policies (options.policies vs default profile)... I think the one in the default profile should be used.

tolbertam: I agree that using the one from the defaultProfile makes the most sense (especially since it will usually be options.policies anyways, and in the case it isn't its probably because it was set explicitly in the default profile instead of options.policies)

jorgebay: In the following example:
```javascript
const client = new Client({ 
  contactPoints: ['host1'], 
  profiles: { 
    'default': new ExecutionProfile({
      loadBalancing: myLbPolicy,
      retry: myRetryPolicy
    }) 
  }
});
```
Makes sense to use `myLbPolicy` as default, regardless of the default lbp policy.

If you agree, I'll document it (and remove the unused field).

tolbertam: 👍

lib/client.js
```diff
@@ -358,68 +367,56 @@ Client.prototype.execute = function () {
  * <p>
  *   When dealing with paged results, [ResultSet#nextPage()]{@link module:types~ResultSet#nextPage} method can be used
  *   to retrieve the following page. In that case, <code>rowCallback()</code> will be again called for each row and
- *   this callback will be invoked when all rows in the following page has been retrieved.
+ *   the final callback will be invoked when all rows in the following page has been retrieved.
  * </p>
+ * @example <caption>Using per-row callback and arrow functions</caption>
+ * client.eachRow(query, params, { prepare: true }, (n, row) => console.log(n, row), err => console.error(err));
+ * @example <caption>Overloads</caption>
+ * client.eachRow(query, rowCallback);
```
tolbertam: Isn't the same somewhat else true for `execute` and `stream` as well?  (in terms of the function arguments being overloaded)

test/integration/short/client-each-row-tests.js
```diff
@@ -125,10 +125,6 @@ describe('Client', function () {
             assert.ifError(err);
             assert.strictEqual(rowCount, 100);
             assert.strictEqual(rowCount, result.rowLength);
-            assert.ok(result.rowLengthArray);
```
tolbertam: Why remove these assertions?

lib/client.js
```diff
@@ -440,70 +437,62 @@ Client.prototype.eachRow = function () {
...
  */
-Client.prototype.stream = function () {
-  var args = Array.prototype.slice.call(arguments);
-  if (typeof args[args.length-1] !== 'function') {
-    //the callback is not required
-    args.push(utils.noop);
-  }
-  args = utils.parseCommonArgs.apply(null, args);
+Client.prototype.stream = function (query, params, options, callback) {
+  callback = utils.bindDomain(callback || utils.noop);
```
jorgebay: In this case, `callback` is an optional parameter (normally shouldn't be used) as you have `stream.on('end', handler)` event.
Optional parameters are left-to-right, so `callback` will only be defined if `params` and `options` are defined.

The only method that breaks the left-to-right rule is `eachRow()`, because the last `callback` is optional but `rowCallback` isn't (I don't like it but its legacy and we have to support it, at least until the next major version). In that case, I've included the overloads in the doc.

test/unit/client-tests.js
```diff
@@ -55,27 +57,22 @@ describe('Client', function () {
...
+      ccMock.prototype.init = function (cb) {
+        initCounter++;
+        //Async
+        setTimeout(cb, 100);
+      };
+      Client.__set__("ControlConnection", ccMock);
       var client = new Client(options);
       client.on('connected', function () {emitCounter++;});
-      utils.times(1000, function (n, next) {
+      utils.timesSeries(1, function (n, next) {
```
tolbertam: is this accidental?  It still seems to work with `util.times(100...)` and appears to be the intent of the test (to try multiple times and ensure init'd once)

test/integration/short/client-each-row-tests.js
jorgebay: `rowLengthArray` was never part of the public API, it was meant for testing but it's not a good way to expose just to test it, so I took the chance to remove it.

test/unit/client-tests.js
jorgebay: nice catch, will fix.

lib/client.js
tolbertam: I see, it all makes sense to me now, thanks for explaining, 👍

lib/client.js
jorgebay: The other methods only support left-to-write optional arguments. 
For example: In `#execute(query, params, options, callback)`
`params` and `options` are optional. The second optional parameter (`options`) can only be defined if the first is defined (`params`).

tolbertam: this was explained [here](https://github.com/datastax/nodejs-driver/pull/156/files#r65418499)

lib/client-options.js
```diff
@@ -129,6 +129,61 @@ function extend(baseOptions, userOptions) {
...
+ * - {@link ExecutionProfile}.
+ * - {@link QueryOptions} from the default options.
+ * @param {Client} client
+ * @param {QueryOptions|function} userOptions
+ * @param {Function} [rowCallback]
+ * @param {Boolean} [logged]
+ */
+function createQueryOptions(client, userOptions, rowCallback, logged) {
+  var profile =
+    client.profileManager.getProfile(userOptions && userOptions.executionProfile) || utils.emptyObject;
```
tolbertam: I really, really like how this is the only place the profile is resolved, and then is attached to created options, it keeps the rest of the code very simple.  There isn't a lot of duplicated logic all over the place around 'is it on the query options?  is it on the profile? is it on the base options?' and so on, it's all contained here 👍

lib/request-handler.js
```diff
@@ -16,20 +16,22 @@ var retryOnNextHostDecision = {
...
  * @constructor
  */
 function RequestHandler(client, options) {
   this.client = client;
-  this.loadBalancingPolicy = options.policies.loadBalancing;
-  this.retryPolicy = options.policies.retry;
+  this.profileManager = client.profileManager;
+  var profile = client.profileManager.getDefault();
+  this.loadBalancingPolicy = profile.loadBalancing;
+  this.retryPolicy = profile.retry;
```
tolbertam: Shouldn't retryPolicy be derived from options.retry instead?  Otherwise query level retry will not be used, from client-options#createQueryOptions#160:

```javascript
retry: ifUndefined3(userOptions.retry, profile.retry, defaultQueryOptions.retry),
```

tolbertam: Another thing I was wondering is why the retry policy can be specified on the query options, but not the load balancing policy?   My guess:   The load balancing policies are used in control connection for determining host distance, so if the user were to provide an lbPolicy in the query options, it could violate the distance resolution there.

lib/request-handler.js
```diff
@@ -16,20 +16,22 @@ var retryOnNextHostDecision = {
...
  * @param {Object} options
  * @constructor
  */
 function RequestHandler(client, options) {
   this.client = client;
-  this.loadBalancingPolicy = options.policies.loadBalancing;
-  this.retryPolicy = options.policies.retry;
+  this.profileManager = client.profileManager;
+  var profile = client.profileManager.getDefault();
+  this.loadBalancingPolicy = profile.loadBalancing;
```
tolbertam: Another thing I was wondering is why the retry policy can be specified on the query options, but not the load balancing policy? My guess: The load balancing policies are used in control connection for determining host distance, so if the user were to provide an lbPolicy in the query options, it could violate the distance resolution there.

lib/request-handler.js
```diff
@@ -78,8 +77,8 @@ RequestHandler.prototype.iterateThroughHosts = function (iterator, callback) {
       return callback(new errors.NoHostAvailableError(this.triedHosts));
     }
     host = item.value;
-    // set the distance relative to the client first, using the load balancing policy
-    var distance = host.getDistance(this.loadBalancingPolicy);
+    // set the distance relative to the client first
+    var distance = this.profileManager.getDistance(host);
```
tolbertam: Shouldn't you use the loadBalancingPolicy on this RequestHandler to determine distance instead?

lib/client-options.js
```diff
@@ -129,6 +129,61 @@ function extend(baseOptions, userOptions) {
...
+  // userOptions can be undefined and could be of type function (is an optional parameter)
+  userOptions = (!userOptions || typeof userOptions === 'function') ? utils.emptyObject : userOptions;
+  var defaultQueryOptions = client.options.queryOptions;
+  // Using fixed property names is 2 order of magnitude faster than dynamically shallow clone objects
```
tolbertam: The behavior of things when it comes to performance with js always surprises me :D

lib/client.js
```diff
@@ -129,10 +132,12 @@ var clientOptions = require('./client-options');
  * </p>
  * Default: false.
```
tolbertam: There are some query options that could be candidates for execution profile options as well, I think these are reasonable considerations:

* fetchSize
* customPayload
* retryOnTimeout
* traceQuery

The other ones seem like they'd apply to a specific query or context (i.e. you wouldn't want to set the same pageState for a set of queries, or have a profile that always uses prepared statements).  That being said, I'm not exactly sure if these options are worth it / or needed.

lib/request-handler.js
tolbertam: I see, i had this wrong, the options passed in here are from the Client, not at the query level, those are passed in in send, at which point this.retryPolicy is set to what is coming from options.

lib/client-options.js
```diff
@@ -129,6 +129,61 @@ function extend(baseOptions, userOptions) {
...
+    consistency: ifUndefined3(userOptions.consistency, profile.consistency, defaultQueryOptions.consistency),
+    customPayload: ifUndefined(userOptions.customPayload, defaultQueryOptions.customPayload),
+    executionProfile: profile,
+    fetchSize: ifUndefined(userOptions.fetchSize, defaultQueryOptions.fetchSize),
+    hints: userOptions.hints,
+    logged: ifUndefined(userOptions.logged, logged),
+    pageState: userOptions.pageState,
+    prepare: ifUndefined(userOptions.prepare, defaultQueryOptions.prepare),
+    readTimeout: ifUndefined3(userOptions.readTimeout, profile.readTimeout, defaultQueryOptions.readTimeout),
+    retry: ifUndefined3(userOptions.retry, profile.retry, defaultQueryOptions.retry),
```
tolbertam: retry isn't defined in `defaultOptions()` but is rather on `client.options.policy` so if its not defined on the user provided query options or profile it will be undefined.

lib/execution-profile.js
```diff
@@ -0,0 +1,129 @@
...
+ * @ignore
+ */
+function ProfileManager(options) {
+  this.profiles = options.profiles || {};
+  this.setDefault(options);
+  // use an array of profiles to iterate more efficiently
+  this.profilesArray = Object.keys(this.profiles).map(function (name) {
+    // set required properties
+    var p = this.profiles[name];
+    p.loadBalancing = p.loadBalancing || this._defaultProfile.loadBalancing;
```
tolbertam: One thing I bumped into that I couldn't wrap my head around at first was that it is probably not a good idea to share ExecutionProfile instances between Client instances.  Just like you wouldn't share LoadBalancingPolicies because it has host-specific details.  We should document profiles should not be shared between client instances.  Alternatively we could decorate ExecutionProfiles as they are evaluated with some kind of property indicating that they are already tied to a Cluster or something to that effect.

lib/client-options.js
tolbertam: Fixed in 6349dcf  (hope that is the right way to go)

lib/request-handler.js
```diff
@@ -143,19 +142,19 @@ RequestHandler.prototype.log = utils.log;
...
-    if (options && options.captureStackTrace) {
+    options = options || utils.emptyObject;
+    if (options.captureStackTrace) {
       //noinspection JSUnresolvedFunction
       Error.captureStackTrace(this.stackContainer);
     }
-    if (options && options.retry) {
-      this.retryPolicy = options.retry;
-    }
+    var profile = options.executionProfile || this.profileManager.getDefault();
```
tolbertam: In createQueryOptions, if there is no matching profile found, options.executionProfile gets set to utils.emptyObject which evaluates so the default profile is not pulled in,  this causes this.loadBalancingPolicy to be undefined.

tolbertam: Also technically couldn't:

```javascript
this.retryPolicy = options.retry || profile.retry;
```

simply be:

```javascript
this.retryPolicy = options.retry;
```

Since options.retry should be set to profile.retry if there is no query level retry in createQueryOptions?

tolbertam: > In createQueryOptions, if there is no matching profile found, options.executionProfile gets set to utils.emptyObject 

Now that I think about it,  instead of setting options.executionProfile to utils.emptyObject, why not set it to profileManager.getDefault() there?  I think it would solve this problem as a side effect.

tolbertam:
Pushed up a few tests around using timeout set in profiles, will push some more tests tomorrow.

lib/request-handler.js
jorgebay: Before execution profiles, users could override the retry policy at execution / statement level. Load balancing policies were not allowed and are not allowed. Lbp requires initialization and also to be considered when determining the host distance.

lib/request-handler.js
jorgebay: According to the design doc, all load balancing policies should be used to determine the (closest) distance.
If a execution profile uses a lbp that identifies that a host is `remote` / `ignored` but other lbp, in a different exec profile, identifies the same host as `local`. The host is considered as `local` (having a pool according to the distance).

lib/client.js
jorgebay: `retryOnTimeout` was used before `onRequestError()` was added to the `RetryPolicy` prototype. Now it's maintained for historical reasons.

Could be useful is to have the `fetchSize`, though.

lib/client-options.js
jorgebay: 👍

test/integration/short/timeout-tests.js
```diff
@@ -7,6 +7,7 @@ var Connection = require('../../../lib/connection');
 var utils = require('../../../lib/utils');
 var types = require('../../../lib/types');
 var errors = require('../../../lib/errors');
+var ExecutionProfile = require('../../../lib/execution-profile').ExecutionProfile;
 
 describe('client read timeouts', function () {
   this.timeout(120000);
```
jorgebay: Can we take this opportunity to speed up this test suite: use `before()` / `after()` to create the cluster and use `beforeEach()` to resume the paused nodes.
Reusing the same cluster would break test isolation (single test errors can affect following tests in the suite to fail) but this test suite takes ~8 mins to complete...

lib/request-handler.js
jorgebay: `RequestHandler` has 3 public methods: `#send()`, `#prepareMultiple()` (preparing for batches) and `setKeyspace()` (not so easy to see, I've created [NODEJS-266][NODEJS-266] to use proper notation for private methods).

For all cases, the load-balancing policy (and retry policy) should be the one defined in the query options (inheriting from profiles) and be the same for as long `RequestHandler` lives (`final`/`readonly` in java or C#). Your doubts are valid, its not clear in the code: I'll change it to instance fields initiated on the `RequestHandler`.

[NODEJS-266]: https://datastax-oss.atlassian.net/browse/NODEJS-266

test/integration/short/timeout-tests.js
tolbertam: Great idea.  I'll do that.

lib/client.js
tolbertam: > retryOnTimeout was used before onRequestError() was added to the RetryPolicy prototype. Now it's maintained for historical reasons.

Good point, we should consider removing it in the next major release :)

lib/request-handler.js
tolbertam: But if a user is giving a specific profile as part of their query execution, shouldn't we obey the distance rules for the policy for that profile?

For example, if a user is executing a query that they want to target an analytics or search datacenter, they would use a DCAwareRoundRobinPolicy with the target DC in the profile for those requests, and use a different profile for other requests where they don't want to target those DCs.

tolbertam: Ah, but maybe my concerns are covered by the fact that we use the profile's load balancing policy ultimately when we call newQueryPlan?

lib/request-handler.js
tolbertam: Thanks, looking at your recent commit those changes make sense to me (moving the assignment of the lb and retry policy to always be covered by the constructor)

test/integration/short/timeout-tests.js
tolbertam: Pushed that change.  It shortens timeout-tests from ~7-8 mins to 2 mins, although somewhat curious to see how reliably it runs in CI as i'm worried there might be a timing window after resuming a node where it isn't responsive, but usually unpausing gets thing in order really quick.  We'll see how it goes.

lib/request-handler.js
jorgebay: `this.loadBalancingPolicy` will be used for the query plan, the `distance()` is to determine the pool size. We should not change the distance in short periods of time, that is why the "closest" distance wins.

test/integration/short/timeout-tests.js
tolbertam: It looks like there are more test regressions that are unrelated to this - digging into it.

lib/request-handler.js
tolbertam: That makes sense.  I was conflating distance with host selection (newQueryPlan).  I think `this.loadBalancingPolicy` being used for the query plan and `profileManager.getDistance` being used for distance makes sense.  Especially given your point about how that is used for determining pool size.

lib/client-options.js
```diff
@@ -129,6 +129,61 @@ function extend(baseOptions, userOptions) {
...
+    captureStackTrace: ifUndefined(userOptions.captureStackTrace, defaultQueryOptions.captureStackTrace),
+    consistency: ifUndefined3(userOptions.consistency, profile.consistency, defaultQueryOptions.consistency),
+    customPayload: ifUndefined(userOptions.customPayload, defaultQueryOptions.customPayload),
+    executionProfile: profile,
+    fetchSize: ifUndefined(userOptions.fetchSize, defaultQueryOptions.fetchSize),
+    hints: userOptions.hints,
+    logged: ifUndefined(userOptions.logged, logged),
+    pageState: userOptions.pageState,
+    prepare: ifUndefined(userOptions.prepare, defaultQueryOptions.prepare),
+    readTimeout: ifUndefined3(userOptions.readTimeout, profile.readTimeout, defaultQueryOptions.readTimeout),
```
jorgebay: Instead of 
```javascript
readTimeout: ifUndefined3(userOptions.readTimeout, profile.readTimeout, defaultQueryOptions.readTimeout)
```

It should be
```javascript
readTimeout: ifUndefined3(userOptions.readTimeout, profile.readTimeout, client.options.socketOptions.readTimeout)
```

test/integration/short/timeout-tests.js
tolbertam: Fixed

lib/client.js
jorgebay: Done

jorgebay:
Trying to extend executions profiles from another module, I found that it makes sense for the `ExecutionProfile` to have a name property (as there aren't hashcodes in Javascript).

So I propose, instead of having the profiles keyed by name:
```javascript
const client = new Client({ 
  contactPoints: ['host1'], 
  profiles: { 
    'default': new ExecutionProfile(profileOptions1),
    'metrics': new ExecutionProfile(profileOptions2) 
  }
});
```

Make profiles an `Array`, and the name of the `ExecutionProfile` a mandatory argument:
```javascript
const client = new Client({ 
  contactPoints: ['host1'], 
  profiles: [
    new ExecutionProfile('default', profileOptions1) 
    new ExecutionProfile('metrics', profileOptions2) 
  ]
});
```

I'll make the change

tolbertam:
Pushed some more tests for execution profiles, including tests that ensure an `ArgumentError` is present in the err parameter of a callback that will fail.  Wasn't sure if `ArgumentError` was the correct expectation or not, feel free to change based on how you think it should raise an error.

test/integration/long/execution-profile-tests.js
```diff
@@ -0,0 +1,93 @@
...
+var helper = require('../../test-helper.js');
+var Client = require('../../../lib/client.js');
+var types = require('../../../lib/types');
+var utils = require('../../../lib/utils.js');
+var loadBalancing = require('../../../lib/policies/load-balancing.js');
+var DCAwareRoundRobinPolicy = loadBalancing.DCAwareRoundRobinPolicy;
+var WhiteListPolicy = loadBalancing.WhiteListPolicy;
+var ExecutionProfile = require('../../../lib/execution-profile.js').ExecutionProfile;
+
+describe('ProfileManager', function() {
```
tolbertam: I created this as a long test because it creates 6 nodes, but it isn't too slow (takes about 30 seconds including time to create/teardown server).

test/integration/long/execution-profile-tests.js
```diff
@@ -0,0 +1,93 @@
...
+    loadBalancing: new WhiteListPolicy(new DCAwareRoundRobinPolicy('dc3'), [helper.ipPrefix + '6' + ":9042"])
+  });
+  var emptyProfile = new ExecutionProfile('empty');
+
+  var profiles = [dc1Profile, dc2Profile, wlProfile, emptyProfile];
+
+  function newInstance(options) {
+    options = options || {};
+    options = utils.extend({
+      profiles: profiles
```
jorgebay: We should use different profile instances for different clients. It works here because the cluster is the same, but `lbp.init()` is being called multiple times (per each client).

Something like: `profiles: createProfiles()`

test/integration/long/execution-profile-tests.js
```diff
@@ -0,0 +1,93 @@
...
+var WhiteListPolicy = loadBalancing.WhiteListPolicy;
+var ExecutionProfile = require('../../../lib/execution-profile.js').ExecutionProfile;
+
+describe('ProfileManager', function() {
+  this.timeout(120000);
+  before(helper.ccmHelper.start('2:2:2'));
+  after(helper.ccmHelper.remove);
+
+  var dc1Profile = new ExecutionProfile('default', {
+    loadBalancing: new DCAwareRoundRobinPolicy('dc1')
```
jorgebay: We could add a method to check that `init()` is being called only once:
```javascript
// ...
  loadBalancing: initOnlyOnce(new DCAwareRoundRobinPolicy('dc2'))
// ...
function initOnlyOnce(policy) {
  var baseInit = policy.init;
  policy._initCalled = 0;
  policy.init = function () {
    policy._initCalled++;
    baseInit.apply(policy, arguments);
  };
  return policy;
}

// ...
client.options.profiles.forEach(function (p) {
  assert.strictEqual(p.loadBalancing._initCalled, 1);
});
```

This assertion can be interesting as `emptyProfile` uses the same lbp as `dc1Profile`.

test/integration/long/execution-profile-tests.js
jorgebay: On Windows CI env, 6-node cluster might be too much.
Do you think we can use a 4 nodes (`2:2`) and move it to "short" category?

jorgebay:
I've pushed a commit to validate that the profile exists: Jenkins is looking good, Appveyor is still failing (on test setup but ASF servers seem fine).

test/integration/long/execution-profile-tests.js
tolbertam: Good point, thought I had fixed that (I raised sharing profiles as a concern the other day so not sure why i didn't think to do that here).  Will fix it.

test/integration/long/execution-profile-tests.js
tolbertam: Sure, that sounds good to me, will do that.

test/integration/long/execution-profile-tests.js
tolbertam: great idea - will implement this.

tolbertam:
> Appveyor is still failing (on test setup but ASF servers seem fine).

Will get to the bottom of that today.

tolbertam:
Pushed changes to tests based on feedback.  Also working on the appveyor issue.  I think if jenkins runs clean and if we can get appveyor into shape this should be good to go 👍

tolbertam:
Fixed the appveyor issues and while I was at it also added runtime for nodejs 6 and upgraded to C* 3.0.6.   Apparently a recent commit on ccm master breaks the pull down in this environment (but not in windows in general).  Using the version for pypi fixes the issue, i'll try to track down what the actual problem is, but for now this is fixed and everything is running good 👍

PR #158, opened 2016-06-08
Allow custom query options

jorgebay:
Fixes breaking change introduced in #156 for NODEJS-201 that didn't allowed custom query options.

I think a found a good compromise between shallow cloning user query options + profile options + default options and having a fixed set of options. It eats up some of the throughput rate improvements from #156.

In my local machine compared to baseline (3b7424618f0343c68062e3dbfc4f9f8ac19c0212):
  - Fixed set of options: ~12% more throughput.
  - Allowing custom options: ~7%

### DIFF
--- a/lib/client-options.js
+++ b/lib/client-options.js
@@ -154,7 +154,7 @@ function createQueryOptions(client, userOptions, rowCallback, logged) {
   userOptions = (!userOptions || typeof userOptions === 'function') ? utils.emptyObject : userOptions;
   var defaultQueryOptions = client.options.queryOptions;
   // Using fixed property names is 2 order of magnitude faster than dynamically shallow clone objects
-  return {
+  var result = {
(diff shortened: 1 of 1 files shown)


lib/client-options.js
```diff
@@ -179,6 +179,26 @@ function createQueryOptions(client, userOptions, rowCallback, logged) {
...
+    key = userOptionsKeys[i];
+    if (key === 'executionProfile') {
+      // Execution profile was the only value that could has been "replaced"
+      continue;
+    }
+    value = userOptions[key];
+    if (value === undefined) {
+      continue;
+    }
+    result[key] = value;
```
tolbertam: It seems safe to override `result[key]` if it exists because the default behavior is to use userOptions anyways (in all cases but executionProfile like you documented above) 👍

tolbertam:
Looks good and appears to have fixed the issue 👍

PR #160, opened 2016-06-10
Allow options.readTimeout to be used when 0.

tolbertam:
For [NODEJS-273](https://datastax-oss.atlassian.net/browse/NODEJS-273)

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -592,7 +592,7 @@ Connection.prototype.invokeCallback = function (handler, err, response) {
  * @returns {Object} The timeout object
  */
 Connection.prototype.setRequestTimeout = function (request, options) {
-  var millis = (options && options.readTimeout) || this.options.socketOptions.readTimeout;
+  var millis = (options && options.readTimeout !== undefined) ? options.readTimeout : this.options.socketOptions.readTimeout;
   if (!(millis > 0)) {
     //read timeout disabled
     return null;
--- a/test/integration/short/timeout-tests.js
+++ b/test/integration/short/timeout-tests.js
@@ -19,7 +19,7 @@ describe('client read timeouts', function () {
     helper.ccmHelper.resumeNode(2, done);
   });
   describe('when socketOptions.readTimeout is not set', function () {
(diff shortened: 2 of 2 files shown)


lib/connection.js
```diff
@@ -592,7 +592,7 @@ Connection.prototype.invokeCallback = function (handler, err, response) {
  * @returns {Object} The timeout object
  */
 Connection.prototype.setRequestTimeout = function (request, options) {
-  var millis = (options && options.readTimeout) || this.options.socketOptions.readTimeout;
+  var millis = (options && options.readTimeout !== undefined) ? options.readTimeout : this.options.socketOptions.readTimeout;
```
tolbertam: Adapted this from clientOptions.ifUndefined to be consistent with that behavior.

test/integration/short/timeout-tests.js
```diff
@@ -196,9 +196,9 @@ describe('client read timeouts', function () {
     it('should be used instead of socketOptions.readTimeout and profile.readTimeout for prepared queries executions',
       getMoveNextHostTest(true, true, 3123, 1 << 24, { executionProfile: 'aProfile', readTimeout: 3123 }, profiles()));
     it('should suppress socketOptions.readTimeout and profile.readTimeout when set to 0 for simple queries',
-      getTimeoutErrorExpectedTest(false, false, 3000, { executionProfile: 'aProfile', readTimeout: 0}, profiles()));
-    it('should suppress socketOptions.readTimeout and profile.readTimeout  when set to 0 for prepared queries executions',
-      getTimeoutErrorExpectedTest(true, true, 3000, { executionProfile: 'aProfile', readTimeout: 0}, profiles()));
+      getTimeoutErrorNotExpectedTest(false, false, 1000, { executionProfile: 'aProfile', readTimeout: 0}, profiles()));
```
tolbertam: These tests weren't effective previously as we only paused the node for 2 seconds, so we gave it enough time to respond within 3 seconds.

jorgebay:
Lgtm!

PR #161, opened 2016-06-15
Update build.yaml, remove nodejs 5 for 6.  Add C* 3.7

tolbertam:
Also add nodejs 6 to .travis.yml.  Use 3.0.7 in appveyor.yml.

Rationale:

* nodejs 5 is not an LTS release, so to reduce integration build axis its not included.
* C* 3.7 is the latest 'tock' release.  Would be good to include that on the matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: node_js
 node_js:
+  - "6"
   - "5"
   - "4"
   - "0.12"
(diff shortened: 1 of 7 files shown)


tolbertam:
Small change needed for a test w/ 3.7, fixing shortly

tolbertam:
should be good to go

PR #162, opened 2016-06-17
Doc: filter duplicates and table of contents

jorgebay:


doc/TOC.md
```diff
@@ -0,0 +1,34 @@
+The DataStax Node.js Driver for Apache Cassandra is a modern [feature-rich](#features) and highly tunable Node.js
```
tolbertam: It doesn't look like there is a #features part of this page, did you mean to link to submodules?

tolbertam: Ah i see, the tagline/link comes from the README.md page in our repo.

tolbertam:
Looks great, only other comment I have is it'd be good to update README.md in the repo root to point to the TOC page root instead of Client.html (but only after this is deployed of course) for the API Docs link.

doc/TOC.md
jorgebay: Fixed, thanks!

PR #164, opened 2016-06-27
Update code samples

jorgebay:
Use a package.json with dependencies for code sample modules.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -24,4 +24,14 @@ build:
   - npm: install
   - npm: run-script ci
     graceful: true
+  - script: |
+      cd examples
+      npm install
+      npm install ../
+      ccm create test_samples -n 1 -b -s -v $TEST_CASSANDRA_VERSION
+      node runner.js
+      run_rc=$?
+      ccm remove
+      exit $run_rc
+    strict: false
   - xunit: "*.xml"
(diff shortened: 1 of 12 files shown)


tolbertam:
Added CI enablement ([NODEJS-282](https://datastax-oss.atlassian.net/browse/NODEJS-282)) and also updated runner to time out runs (for [NODEJS-283](https://datastax-oss.atlassian.net/browse/NODEJS-283)) and count number of failures.

build.yaml
```diff
@@ -24,4 +24,14 @@ build:
   - npm: install
   - npm: run-script ci
     graceful: true
+  - script: |
+      cd examples
+      npm install
+      npm install ../
```
jorgebay: Is this needed?

tolbertam: The `npm install ../` step points to the local cassandra driver, where as if i were to just use npm install, it'd use the version from master.

jorgebay: You are right! I've changed the dependency to point to local file in the package.json:
```
"dependencies": {
  "cassandra-driver": "file:../",
  // ...
}
```

It should not be required now.

tolbertam: Nice 👍

tolbertam:
lgtm 👍

PR #165, opened 2016-06-27
Log hardware stats on startup and finish.

tolbertam:
Also moves appveyor powershell scripts into appveyor folder (so don't pollute repo root).

WIP until confident it properly works on finish.


tolbertam:
Removing wip tag, should be good to go.

appveyor.yml
```diff
@@ -11,12 +11,15 @@ environment:
 platform:
   - x64
 install:
-  - ps: .\appveyor.ps1
+  - ps: .\appveyor\hardwareinfo.ps1
```
jorgebay: + 1 on the scripts to a folder.

What do you think about creating a generic folder for current build scripts (for now just appveyor related ones), named `build-scripts`,  `ci` or something like that (not `build` as it's most commonly used for output folder for c modules), without the name of the vendor, that way we can also add it to the `.npmignore` file.

jorgebay:
Looking at the output, starting the tests with 1.46GB of RAM on Windows Server is too low. I don't know which policy they use to allocate more RAM during the tests or what's the performance of the RAM that's allocated.

Test runs on appveyor used to succeed more than 95% of the times. Now, a test run is likely to succeed less than 50% of the times: nothing changed on our side, it should be an issue on their side.
Creating a 3 node cluster in a reliable way using ccm is not too much to ask to a CI environment.

I'd say, either we discuss this issue with appveyor support and find a solution or we should remove it for integration tests (we could use the free open source license for unit tests).

appveyor.yml
tolbertam: +1000000 on this, we could even move appveyor.yml in there, since that can be in a custom place (not for .travis or build.yaml though)

tolbertam:
update for anyone watching:  the appveyor team fixed our problem (we were on a different infrastructure, we've since been quickly migrated)

tolbertam:
Moved files in `appveyor/` directory to `ci/`.  I'll also move `appveyor.yml` into `ci` if you approve.  As part of that we'll need to update our external appveyor configuration to look for `ci/appveyor.yml`.  After we make that change appveyor will only do builds for branches that have `ci/appveyor.yml`

jorgebay:
Go ahead and move appveyor.yml, its OK to just build master and new branches on Windows.

tolbertam:
Sounds good, will push that shortly.

tolbertam:
Rebased, made that change and squashed.  Also updated appveyor configuration to look for `ci/appveyor.yml`

ci/appveyor.yml
```diff
@@ -11,12 +11,15 @@ environment:
...
+  - ps: .\ci\appveyor.ps1
   - npm install
 build: off
 test_script:
   - "npm run %ci_type%"
+on_finish:
+  - ps: .\ci\hardwareinfo.ps1
 cache:
   - node_modules -> package.json
   - C:\Users\appveyor\.ccm\repository
```
jorgebay: I'm still trying to wrap my head around appveyor cache invalidation, shouldn't this be:
```
   - C:\Users\appveyor\.ccm\repository -> .\ci\appveyor.ps1
```

tolbertam: In theory we would never want to invalidate the .ccm\repository directory, since the cache is per matrix configuration, which includes the full cassandra version, which shouldn't change.  However if things get broken or we fundamentally change how C* is installed, it might be nice to invalidate this anyways.

PR #166, opened 2016-06-27
NODEJS-283 Close newly opened connection if pool was shutdown.

tolbertam:
For [NODEJS-283](https://datastax-oss.atlassian.net/browse/NODEJS-283).

### DIFF
--- a/lib/host-connection-pool.js
+++ b/lib/host-connection-pool.js
@@ -154,6 +154,9 @@ HostConnectionPool.prototype.clearNewConnectionAttempt = function () {
  * @param {Function} callback
  */
 HostConnectionPool.prototype._attemptNewConnection = function (callback) {
+  if(this.shuttingDown) {
+    callback(Error("Pool is shutting down"));
+  }
   var c = this._createConnection();
   var self = this;
   this.once('open', callback);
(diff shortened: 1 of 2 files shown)


lib/host-connection-pool.js
```diff
@@ -169,6 +172,11 @@ HostConnectionPool.prototype._attemptNewConnection = function (callback) {
       c.close();
       return self.emit('open', err);
     }
+    // close the connection if pool was shutdown.
```
tolbertam: This seemed to be the most straightforward fix, although there may be a more proactive way to close the connection that is in the process of connecting.

jorgebay: +1 I think waiting for it to finish opening is the right thing to do.

jorgebay:
There are few ways, that I can think of, to test this issue.
Migrating all tests to use mocha flag `--no-exit` is a pain. We can set a timeout and check the pool state using the internal properties: `host.pool.connections.length`.
I will try to create a test for this after 3.1.0 release. 

Luckily it is now easier to identify this issue with the code samples module.

jorgebay:
I pushed a test that shows that it can still leaks connections.
The underlying reason is that the method `setAsShutdown()` resets the value of the `shuttingDown` property.

`shuttingDown` is being used to cover the situation where the load balancing policy changes the selected hosts considered as local/remote and marks them as ignored. In those cases, the driver should drain the connection (wait for the in-flight requests to finish) and reset the state of the pool, leaving it ready to be selected in the future.

We should introduce a new state in the pool when its being shutdown for good...

tolbertam:
Closing in favor of #169

PR #167, opened 2016-06-29
Resolve host name using dns.lookup() in case DNS queries fail

jorgebay:
In case the methods `dns.resolve4()` and `dns.resolve6()` are not able to resolve an address for a given host name, use `dns.lookup()` that performs a `getaddrinfo()` system call and might resolve using host system definitions.
Ignore `resolve4()` and `resolve6()` call errors.

### DIFF
--- a/lib/control-connection.js
+++ b/lib/control-connection.js
@@ -85,7 +85,7 @@ ControlConnection.prototype.init = function (callback) {
   }
   utils.series([
     function resolveNames(next) {
-      utils.each(self.options.contactPoints, function (name, eachNext) {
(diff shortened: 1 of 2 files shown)


jorgebay:
I've added additional unit tests using a dns mocks, increasing the coverage for failure scenarios.
I don't see a way we could add integration tests without using system host files, but I think new coverage is good enough.

tolbertam:
Just tried this out and it seems to work good.   Added an entry 'cseeds' to /etc/hosts pointing to 127.0.0.1, 127.0.0.2, 127.0.0.3, ::1.   With master it fails, on NODEJS-284 branch it works (and resolves all 4 addresses). 👍

PR #168, opened 2016-06-29
fixes 3.1.0 breaking change - control connection no longer accepting local hostnames

sparkida:
When servers expose local or private hosts which cannot be resolved by dns lookup, the connection will fail. In 3.0.x, a `dns.lookup` was performed which would resolve these hostnames https://github.com/datastax/nodejs-driver/blob/3.0/lib/control-connection.js#L90. Currently, the `resolveAll` call assumes address as ipv4 or ipv6 https://github.com/datastax/nodejs-driver/blob/3.1/lib/control-connection.js#L692. This breaks our setup.

### DIFF
--- a/lib/control-connection.js
+++ b/lib/control-connection.js
@@ -692,6 +692,15 @@ ControlConnection.prototype.shutdown = function () {
 function resolveAll(name, callback) {
   var addresses = [];
   utils.parallel([
+    function lookupHost(next) {
+      dns.lookup(name, function lookupHostCallback(err, ip) {
+        if (ip) {
+          addresses.push(ip);
+        }
+        // Ignore error
+        next();
+      });
+    },
     function resolve4(next) {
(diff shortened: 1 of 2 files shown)


tolbertam:
@sparkida thanks for your contribution.  This should also be fixed by #167 ([NODEJS-284](https://datastax-oss.atlassian.net/browse/NODEJS-284)) which was merged in earlier today and we plan on doing a hotfix release to address this.  Could you take a look and see if that fixes the issue?

Satyricon:
having the same issue

jorgebay:
Thanks @sparkida and @Satyricon 

We will push a v3.1.1 now that includes #167 fix.

jorgebay:
v3.1.1 is now available on npm: https://groups.google.com/a/lists.datastax.com/forum/#!topic/nodejs-driver-user/qt0hq7TYEWk

PR #169, opened 2016-07-01
Introduce host pool states

jorgebay:
Track states of the HostConnectionPool as a FSM.
Differentiate between pool closing, as a result of change in distance, and shutting down, as a result of client shutdown.
Check the state of the pool when creating it and when increasing its size.
When the control connection host is ignored, move to a different host.

Included some tests to check the transitions between local/remote and ignore.

Fixes NODEJS-283: Included an integration test trying to force a race condition between a pool increasing size and being shutdown.

### DIFF
--- a/lib/host-connection-pool.js
+++ b/lib/host-connection-pool.js
@@ -11,6 +11,28 @@ var connectionIndex = 0;
 var connectionIndexOverflow = Math.pow(2, 15);
 
 /**
+ * Represents the possible states of the pool.
+ * Possible state transitions:
+ *  - From initial to closing: The pool must be closed because the host is ignored.
+ *  - From initial to shuttingDown: The pool is being shutdown as a result of a client shutdown.
+ *  - From closing to initial state: The pool finished closing connections (is now ignored) and it resets to
+ *    initial state in case the host is marked as local/remote in the future.
+ *  - From closing to shuttingDown (rare): It was marked as ignored, now the client is being shutdown.
+ *  - From shuttingDown to shutdown: Finished shutting down, the pool should not be reused.
+ * @private
+ */
+var state = {
+  // Initial state: open / opening / ready to be opened
+  initial: 0,
+  // When the pool is being closed as part of a distance change
+  closing: 1,
+  // When the pool is being shutdown for good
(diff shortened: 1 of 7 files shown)


lib/host-connection-pool.js
```diff
@@ -11,6 +11,28 @@ var connectionIndex = 0;
 var connectionIndexOverflow = Math.pow(2, 15);
 
 /**
+ * Represents the possible states of the pool.
+ * Possible state transitions:
```
tolbertam: Nice, this doc is very helpful for understanding.

lib/host-connection-pool.js
```diff
@@ -255,49 +317,58 @@ HostConnectionPool.prototype.drainAndShutdown = function () {
...
-  // check that after sometime (readTimeout + 2 secs) the connections have been drained
-  var delay = (this.options.socketOptions.readTimeout || defaultOptions.socketOptions.readTimeout) + 2000;
-  setTimeout(function checkShutdown() {
-    if (wasSetAsShutdown) {
-      return;
-    }
-    wasSetAsShutdown = true;
-    self.setAsShutdown();
+  // Check that after sometime (readTimeout + 100ms) the connections have been drained
+  var delay = (this.options.socketOptions.readTimeout || defaultOptions.socketOptions.readTimeout) + 100;
```
tolbertam: Is there a main reason for reducing the delay time, is it just because 2 seconds is probably too much time?

tolbertam:
This looks excellent, 👍 .   I also rebased this on the examples-update branch to ensure the issue that we originally ran into was indeed fixed.  I opened up [NODEJS-285](https://datastax-oss.atlassian.net/browse/NODEJS-285) and [NODEJS-286](https://datastax-oss.atlassian.net/browse/NODEJS-286) which i intend on addressing as part of the upcoming release to ensure we don't have any other possible leaks.

lib/host-connection-pool.js
jorgebay: Yes, we should add some additional time to ensure that it happens after `readTimeout` but 2 secs made no sense.

PR #170, opened 2016-08-11
protect getPeersSchemaVersions from null schema version values

orourkedd:
`Client.prototype._waitForSchemaAgreement` fails if its internal call to `self.metadata.getPeersSchemaVersions` returns an array of schema entries with `null` values.  It fails on line 595 when `toString()` is called on the null value.

This patch filters out null values before it returns schema versions.

### DIFF
--- a/lib/metadata/index.js
+++ b/lib/metadata/index.js
@@ -150,7 +150,7 @@ Metadata.prototype.refreshKeyspace = function (name, callback) {
       delete self.keyspaces[name];
       return callback();
     }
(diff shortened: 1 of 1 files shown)


jorgebay:
Makes sense, nice catch!

Our appveyor builds were not working properly (unrelated to this patch), I had to do a rebase to make it run properly on Windows CI: https://ci.appveyor.com/project/DataStax/nodejs-driver/build/387.orourkedd-master

PR #172, opened 2016-08-18
Merged docs.yaml, etc. from dita_customizations

stamhankar999:


docs.yaml
```diff
@@ -31,8 +31,11 @@ links:
...
     href:  https://www.npmjs.org/package/cassandra-driver
+api_docs:
+  v3.1: http://docs.datastax.com/en/drivers/nodejs/3.1/
+  v3.0: http://docs.datastax.com/en/drivers/nodejs/3.0/
 versions:
   - name: v3.1
     ref: HEAD
   - name: v3.0
-    ref: 3.0
+    ref: v3.0.2
```
jorgebay: As a general rule, wouldn't it be best to reference to the branch instead of the tag, that way we avoid maintaining references to tags in this file.

.nav
```diff
@@ -0,0 +1,4 @@
+getting-started
```
jorgebay: Is this file required? It describes the same navigation already available in the [docs.yaml](https://github.com/datastax/nodejs-driver/blob/v3.1.1/docs.yaml#L5-L20).

stamhankar999: It turns out section order in docs.yaml does not translate to section order
in output docs.

-Sandeep
Ruby Driver Dude

<http://cassandrasummit.org/Email_Signature>

PR #173, opened 2016-08-19
Remove connection from host pool when closed by server side

jorgebay:
`Connection` instances now emit `'socketClose'` events and `HostConnectionPool` emits `'remove'` events. This way, the `Host` can listen when a connection has been removed from the pool and decide if it's now considered as down or not.

Removed unclear `connecting` and `isClosed` internal flags in `Connection`.
Includes fixes and improvements for `ControlConnection` mechanism for reconnection.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -59,6 +59,15 @@ function Connection(endpoint, protocolVersion, options) {
   this.encoder = new Encoder(protocolVersion, options);
   this.keyspace = null;
   this.emitDrain = false;
+  /**
+   * Determines if the socket is open and startup succeeded, whether the connection can be used to send requests / 
+   * receive events
+   */
+  this.connected = false;
+  /**
+   * Determines if the socket can be considered as open
+   */
+  this.isSocketOpen = false;
 }
 
 util.inherits(Connection, events.EventEmitter);
@@ -78,8 +87,12 @@ Connection.prototype.bindSocketListeners = function() {
   this.netClient.on('close', function() {
     self.log('info', 'Connection to ' + self.endpoint + ' closed');
     self.isSocketOpen = false;
+    var wasConnected = self.connected;
     self.close();
-    self.emit('close');
+    if (wasConnected) {
+      // Emit only when it was closed unexpectedly
+      self.emit('socketClose');
+    }
   });
 
   this.protocol = new streams.Protocol({ objectMode: true });
@@ -628,6 +641,7 @@ Connection.prototype.onTimeout = function (streamId, millis) {
 
 /**
  * Closes the socket (if not already closed) and cancels all in-flight requests.
+ * Multiple calls to this method have no additional side-effects.
  * @param {Function} [callback]
  */
 Connection.prototype.close = function (callback) {
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -113,6 +113,7 @@ Connection.prototype.bindSocketListeners = function() {
 
 /**
  * Connects a socket and sends the startup protocol messages.
+ * Note that when open() callbacks in error, the caller should immediately call {@link Connection#close}.
  */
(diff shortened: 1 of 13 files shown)


lib/connection.js
```diff
@@ -186,7 +199,6 @@ Connection.prototype.startup = function (callback) {
 };
 
 Connection.prototype.errorConnecting = function (err, destroy, callback) {
```
tolbertam: Shouldn't this change isSocketOpen to false?

tolbertam: hrmm, maybe this would be covered by the close event being handled at line 89

jorgebay: According to the [docs](https://nodejs.org/api/net.html#net_event_error_1):
> Emitted when an error occurs. The 'close' event will be called directly following this event.

But I understand your concern, we should make sure that all situations were `errorConnecting()` is invoked, the connection is left in a consistent state. Let me check...

jorgebay: There are possible 3 calls to `errorConnecting()`:

1. When connecting the socket, as a handler of the `'error'` event. In this case, the error will be called before `isSocketOpen` is set to `true` (it will never be set). ✅ 
2. When connecting the socket, as a handler of `'timeout'` event. The same as before, but `destroy()` should be called, to prevent other events from firing ✅ 
3. When there was an error as part of `STARTUP` request / response. In this case, the socket will be open and the `isSocketOpen` flag set to `true`, and as [the `open()` caller received an error, it should call to `close()` it](https://github.com/datastax/nodejs-driver/blob/75be61652868d6ad726993300d58095dc9bf969f/lib/host-connection-pool.js#L199). But `socket.destroy()` should not be called, which currently is ❌

I'll fix the last use case.

test/integration/short/client-pool-tests.js
```diff
@@ -369,6 +369,45 @@ describe('Client', function () {
...
+          function checkInitialState(next) {
+            var hosts = client.hosts.values();
+            assert.ok(hosts[0].isUp());
+            assert.ok(hosts[1].isUp());
+            next();
+          },
+          function stopNode(next) {
+            helper.ccmHelper.stopNode(nodeNumber, next);
+          },
+          helper.delay(300),
```
tolbertam: Is the 300ms delay needed?  From what I can tell the driver should react quickly enough (usually before the process exists as result of stopNode).

tolbertam:
This looks great, 👍 .

It looks to be working as intended for me as well.  I originally encountered this issue while testing [NODEJS-106](https://datastax-oss.atlassian.net/browse/NODEJS-106) so i went back to the stubbed tests I wrote using scassandra and adjusted them so they no longer execute queries before checking if the host is down.  

This was previously required because the connections wouldn't be removed from the pool until you sent a request on them, so the hosts would remain up until then.   Since the connections are now removed from the pool as soon as they go down, querying is no longer necessary.  

I made this change (0f0654ded123d638875f8e289fba2b14f7a5b837 | [branch](https://github.com/tolbertam/nodejs-driver/tree/NODEJS-265-stubbed)), to validate that querying was no longer necessary and that is the case, where against master the test now fails since it doesn't have this logic.  Eventually i'll incorporate these stubs tests instead of using branches on a personal fork.

I've also updated our duration tests to track the state of isSocketOpen and connected on each connection in the pool, it's been running for an hour or two and things are looking good.

test/integration/short/client-pool-tests.js
jorgebay: As the server is the one terminating the tcp connection, theoretically it may happen that the `Connection` `'close'` event hasn't been emitted when ccm stop exits (acknowledged but was not processed by the event loop yet), but I think a few ms would be enough to make sure...

jorgebay:
Merged.
Really looking forward for stubs tests!

PR #174, opened 2016-08-22
Reset keyspace on the Connection state after changing it

jorgebay:
Fixes: NODEJS-288.


tolbertam:
Looks good to me.   The test change to execute some in parallel/serial seems to expose the existing issue, 👍.

PR #175, opened 2016-08-22
Reset the ControlConnection state after failed init attempt

jorgebay:
- Shutdown all connection pools on failed initialization attempt to avoid reconnection attempts.
- Created test for failed connection attempts, shutting down and logging events.


tolbertam:
Looks great, verified with a version of the script on the issue that attempts are no longer connected and the process exits.

PR #176, opened 2016-08-24
Avoid using retryPolicy.onRequestError() when retryHandler is set

jorgebay:
Fixes NODEJS-296.

### DIFF
--- a/lib/request-handler.js
+++ b/lib/request-handler.js
@@ -251,7 +251,7 @@ RequestHandler.prototype.prepareMultiple = function (queries, callbacksArray, op
   Error.captureStackTrace(this.stackContainer = {});
   var self = this;
   this.retryHandler = function () {
-    //Use a custom retryHandler
+    // Use a custom retryHandler
     self.prepareMultiple(queries, callbacksArray, options, callback);
   };
   this.getNextConnection(options, function (err, c) {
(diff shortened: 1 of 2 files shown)


lib/request-handler.js
```diff
@@ -251,7 +251,7 @@ RequestHandler.prototype.prepareMultiple = function (queries, callbacksArray, op
   Error.captureStackTrace(this.stackContainer = {});
   var self = this;
   this.retryHandler = function () {
-    //Use a custom retryHandler
+    // Use a custom retryHandler
     self.prepareMultiple(queries, callbacksArray, options, callback);
```
tolbertam: Not related to the change, but does this retry preparing indefinitely?  Any concerns about this running repeatedly?  Although looking at the cases it would be exercised (socket error, overloaded/is_bootstrapping/truncate, operation timeout) it doesn't seem like it would be too big of a concern except for maybe the timeout case for whatever reason (i.e. if the prepare behavior causes some weird pause/delay on the cassandra side every time).

jorgebay: It should retry to prepare on all hosts selected by the LBP, not indefinitely.

We could add a test to make sure: use 2 hosts, warmup, pause both of them and try to send a prepared batch. The expected error should be instance of `NoHostAvailableError`.

tolbertam: Ah of course, that makes sense, I didn't consider at the time that it would use the query plan to select hosts. :+1:

tolbertam: > We could add a test to make sure: use 2 hosts, warmup, pause both of them and try to send a prepared batch. The expected error should be instance of NoHostAvailableError.

That sounds good and easy to add, will do push that up shortly.

tolbertam:
Test error doesn't look related, I'll see if I can fix it/make more resilient on another branch.

jorgebay:
+1 for the "Prepared batch exhausting hosts" test!

(the [build failure](https://ci.appveyor.com/project/DataStax/nodejs-driver/build/425.NODEJS-296/job/240y22369igrt6ke) is definitely unrelated: the underlying cause is that the `queryOptions` passed to `newInstance()` in that test overrides the default `queryOptions` value (ie: `extend()` vs `deepExtend()`)

tolbertam:
> (the build failure is definitely unrelated: the underlying cause is that the queryOptions passed to newInstance() in that test overrides the default queryOptions value (ie: extend() vs deepExtend())

Ah!  I didn't notice that the queryOptions were being overriden in the `newInstance` call.  I was scratching my head at why this would happen if quorum was used for reads and writes.  That explains it then.  I'll fix on a separate branch and see if there's anywhere else this could be a problem.

PR #177, opened 2016-08-24
DCAwareRoundRobinPolicy: Use a local reference to hosts

jorgebay:


lib/policies/load-balancing.js
```diff
@@ -209,20 +209,22 @@ DCAwareRoundRobinPolicy.prototype.newQueryPlan = function (keyspace, queryOption
     this.index = 0;
   }
   this._sliceNodesByDc();
+  // Use a local reference of hosts
+  var localHostsArray = this.localHostsArray;
```
tolbertam: It took me a little bit to wrap my head around this, is this what was happening previously to cause the bug?

1. user executes query, which calls newQueryPlan to resolve hosts.
2. a Host is added or removed before/during the query plan is iterated over.  As part of this _cleanHostCache is called which sets localHostsArray/remoteHostsArray to null.
3. query plan is iterated over, because localHostsArray is null, we hit the failure.

Getting a local reference of these arrays makes it so if `this.localHostsArray` is set to null, the previous array is still maintained by this local reference, so we can iterate over the host map made at the time of newQueryPlan being called.

tolbertam: Just tried running the test with the previous version of the code and that seems to be exactly what is happening.

tolbertam:
This looks good to me, 👍 .  It is really hard to reproduce explicitly but seems pretty realistic to happen in practice (albeit rarely), since it would require hosts to be added/removed while iterating over a query plan, which would require a host to go down/get added maybe during a query timeout and then when iterating over the next host you'd hit this.  I think because of that, and the straightforward nature of the change, the unit test is sufficient.

lib/policies/load-balancing.js
jorgebay: You've got it, sorry for not being more descriptive in the ticket... I've should have included more info other than title and stack trace.

I've just copied your description into the ticket for future reference.

jorgebay:
There is a test failure that occurred in the last appveyor build https://ci.appveyor.com/project/DataStax/nodejs-driver/build/420.master/job/kj332kjkj62f9ys4

Looks unrelated but we should try to understand the cause of the failure and fix it.

jorgebay:
The test failure was unrelated, its a race condition in the integration test.
It looks like sometimes C* (on Windows?) could take more time to send the "new node" notification.

PR #178, opened 2016-08-25
Make control connection tests more resilient

tolbertam:
Remove 5 second delays from tests in favor of waiting for a condition to be met within 20 seconds with an attempt every second.  This should allow the tests more time to meet the condition and also possibly complete faster if the condition is met sooner.

### DIFF
--- a/test/integration/short/control-connection-tests.js
+++ b/test/integration/short/control-connection-tests.js
@@ -92,16 +92,11 @@ describe('ControlConnection', function () {
       var cc = newInstance({ policies: { loadBalancing: new TestLoadBalancing() } });
       utils.series([
         cc.init.bind(cc),
-        function (next) {
-          //wait for all initial events
-          setTimeout(next, 5000);
-        },
-        function (next) {
-          helper.waitOnHost(function () {
-            //noinspection JSCheckFunctionSignatures
-            helper.ccmHelper.stopNode(2);
-          }, cc, 2, 'down', next);
-        },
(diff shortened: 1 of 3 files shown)


test/integration/short/control-connection-tests.js
```diff
@@ -92,16 +92,11 @@ describe('ControlConnection', function () {
...
-          setTimeout(next, 5000);
-        },
-        function (next) {
-          helper.waitOnHost(function () {
-            //noinspection JSCheckFunctionSignatures
-            helper.ccmHelper.stopNode(2);
-          }, cc, 2, 'down', next);
-        },
+        // Don't stop the node until we know it's up.
+        helper.waitOnHostUp(cc, 2),
```
tolbertam: One possible downside of this change is that we previously checked for Host 'down' / 'up' events to be emitted, instead we now check the state of the Host (isUp, or not present in the host map at all), which is set as a side effect of the status/topology events (like the Host events I suppose). Is this adequate?  I think it is, but can also add some logic to wait for the host events as well.

jorgebay: I think the change makes sense:
- Up / down host events are the result of a change (ie: was down now it's up), where now you can instantly check it if its up/down and if its not, wait for it.
- Host events are not the result of a control connection status event (at least not always) after [NODEJS-106](https://datastax-oss.atlassian.net/browse/NODEJS-106)

PR #179, opened 2016-08-26
Use deepExtend for Client options in tests

tolbertam:
This is probably overkill, but based on the [discussion here](https://github.com/datastax/nodejs-driver/pull/176#issuecomment-242717496) it seems like using deepExtend is a better choice for extending client options in the tests when calling `newInstance`.  In most cases this is not needed, but it is useful in the event when there are options passed in that could replace base options (like `helper.baseOptions`' use of policies.retry.RetryMultipleTimes being replaced by other policy specifications like LBP).

### DIFF
--- a/test/integration/long/client-metadata-tests.js
+++ b/test/integration/long/client-metadata-tests.js
@@ -151,5 +151,5 @@ function validateMurmurReplicas(client) {
 
 /** @returns {Client}  */
 function newInstance(options) {
-  return new Client(utils.extend({}, helper.baseOptions, options));
+  return new Client(utils.deepExtend({}, helper.baseOptions, options));
 }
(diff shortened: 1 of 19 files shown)


tolbertam:
Test failure was:

```
Condition still false after 20 attempts: function () {
      var host = self.findHost(client, number);
      return host === undefined ? false : !host.isUp();
    }
```

From #178 I guess 20 seconds isn't enough time waiting for a host to go down?  Should we bump it up to a minute?

jorgebay:
Looks good to me.
I'm look into the recent build failures on appveyor, like [this one](use-deep-extend-for-client-options).

jorgebay:
Fix for appveyor recent failures: #181

PR #181, opened 2016-08-29
Test: use additional init delay for ControlConnection STATUS_CHANGE

jorgebay:
Recently [we removed an initial delay in a control connection integration test](https://github.com/datastax/nodejs-driver/pull/178/files#diff-8b7c72a01eed76ae230bbb5e6cd1c32bL97) for internode gossip to settle on Windows before marking the first node as down.

This caused several build failures on appveyor. Here's a fix with a longer delay for Windows. I've tested it [here](https://ci.appveyor.com/project/DataStax/nodejs-driver/build/442.test-status-change-fix/job/6k7y0wpg27reu4cw) and [here](https://ci.appveyor.com/project/DataStax/nodejs-driver/build/441.test-status-change-fix/job/8t2jdyuf14edyrh2).

### DIFF
--- a/test/integration/short/control-connection-tests.js
+++ b/test/integration/short/control-connection-tests.js
@@ -92,6 +92,7 @@ describe('ControlConnection', function () {
       var cc = newInstance({ policies: { loadBalancing: new TestLoadBalancing() } });
       utils.series([
         cc.init.bind(cc),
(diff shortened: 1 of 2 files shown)


tolbertam:
Looks good, I had figured that waiting for the node to be up was sufficient for this particular test, but it definitely isn't, 👍 .

PR #182, opened 2016-08-30
Connection: prevent socket 'error' listener to be executed before 'close' listener

jorgebay:
In some platforms (OSX, Windows), the 'error' handler is called before 'close' when the server sent a `FIN` packet as a result of a higher protocol version than supported.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -177,7 +177,7 @@ Connection.prototype.startup = function (callback) {
         //The host closed the connection, close the socket
         setImmediate(function decreasingVersionClosing() {
           self.close(function decreasingVersionOpening() {
-          //Retry
+            // Attempt to open with the correct protocol version
             self.open(callback);
           });
         });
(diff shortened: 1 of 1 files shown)


lib/connection.js
```diff
@@ -669,6 +669,11 @@ Connection.prototype.close = function (callback) {
     }
     setImmediate(callback);
   });
+  // Prevent 'error' listener to be executed before 'close' listener
+  this.netClient.removeAllListeners('error');
+  // Add a noop handler for 'error' event to prevent Socket to throw the error
+  this.netClient.on('error', utils.noop);
```
jorgebay: We should subscribe to using `on()` and not `once()` because `'error'` event can be fired after closed, multiple times. ie: if we attempt to call `socket.write()`.

jorgebay:
Removed the [WIP] flag, its ready to be reviewed.

tolbertam:
lgtm, 👍 .  Ran a smoke test on both OS X and Windows 2012 with versions 1.2.19, 2.0.17 and 2.1.15.   Kicked off a full short suite on appveyor [here](https://ci.appveyor.com/project/DataStax/nodejs-driver/build/465.NODEJS-303-smoke) for 2.0.17/2.1.15/3.7 which is in progress.  Note that the auth tests are failing, looked into it and reproduced it locally, increasing the start delay from 5 to 10 seconds fixed the issue (needed to give the node more time to set up auth).

PR #184, opened 2016-08-30
[do not merge] Introduce 'ci_smoke' for quick verification on appveyor

tolbertam:
Exercises core functionality.  Will be useful for running against
uncommon configurations where we don't want to always run the full blown
suite on commit.  Should extend coverage as we currently only ruin unit
against those suites on windows.

~~Do not merge until #182 is merged, then this should be rebased on master.~~ [done]

Do not merge, would like to get some more runtime to make sure this remains stable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "test": "./node_modules/.bin/mocha test/unit -R spec -t 5000",
     "ci": "./node_modules/.bin/mocha test/unit test/integration/short -R mocha-multi",
     "ci_unit": "./node_modules/.bin/mocha test/unit -R mocha-multi",
-    "ci_smoke": "./node_modules/.bin/mocha test/unit -R mocha-multi && \"./node_modules/.bin/mocha\" test/integration/short -g \"\\u007csmoke\\u007c\" -R mocha-multi"
+    "ci_smoke": "./node_modules/.bin/mocha test/unit test/integration/short/smoke-tests.js -R mocha-multi"
   },
   "engines": {
     "node" : ">=0.10.0"
--- a/package.json
+++ b/package.json
@@ -35,7 +35,8 @@
   "scripts":  {
     "test": "./node_modules/.bin/mocha test/unit -R spec -t 5000",
     "ci": "./node_modules/.bin/mocha test/unit test/integration/short -R mocha-multi",
-    "ci_unit": "./node_modules/.bin/mocha test/unit -R mocha-multi"
+    "ci_unit": "./node_modules/.bin/mocha test/unit -R mocha-multi",
+    "ci_smoke": "./node_modules/.bin/mocha test/unit -R mocha-multi && \"./node_modules/.bin/mocha\" test/integration/short -g \"\\u007csmoke\\u007c\" -R mocha-multi"
   },
   "engines": {
     "node" : ">=0.10.0"
--- a/test/integration/short/client-batch-tests.js
+++ b/test/integration/short/client-batch-tests.js
@@ -11,7 +11,7 @@ var vit = helper.vit;
 
 describe('Client', function () {
   this.timeout(120000);
-  describe('#batch(queries, {prepare: 0}, callback)', function () {
+  describe('#batch(queries, {prepare: 0}, callback) |smoke|', function () {
     var keyspace = helper.getRandomName('ks');
     var table1 = keyspace + '.' + helper.getRandomName('tblA');
     var table2 = keyspace + '.' + helper.getRandomName('tblB');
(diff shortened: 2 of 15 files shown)


test/integration/short/client-batch-tests.js
```diff
@@ -11,7 +11,7 @@ var vit = helper.vit;
 
 describe('Client', function () {
   this.timeout(120000);
-  describe('#batch(queries, {prepare: 0}, callback)', function () {
+  describe('#batch(queries, {prepare: 0}, callback) |smoke|', function () {
```
tolbertam: Unsure if this is the best solution (adding `|smoke|` to the description), but seems to work.

package.json
```diff
@@ -35,7 +35,9 @@
   "scripts":  {
     "test": "./node_modules/.bin/mocha test/unit -R spec -t 5000",
     "ci": "./node_modules/.bin/mocha test/unit test/integration/short -R mocha-multi",
-    "ci_unit": "./node_modules/.bin/mocha test/unit -R mocha-multi"
+    "ci_unit": "./node_modules/.bin/mocha test/unit -R mocha-multi",
+    "ci_smoke": "./node_modules/.bin/mocha test/unit -R mocha-multi; ./node_modules/.bin/mocha test/integration/short -g \"\\u007csmoke\\u007c\" -R mocha-multi",
+    "ci_smoke_win": "./node_modules/.bin/mocha test/unit -R mocha-multi & \"./node_modules/.bin/mocha\" test/integration/short -g \"\\u007csmoke\\u007c\" -R mocha-multi"
```
tolbertam: This is a bit bizarre for windows to have a separate command like this, explaining all the nuances.

* `\u007c` is a vertical pipe, used instead of `|` because that throws windows off.
* `\\u007c` the backslash is escaped, otherwise it is interpreted as '|' before calling script.
* Double quotes are used because single quotes don't work on windows
* A separate command for windows is needed because `;` doesn't work to separate commands on command line.  `&&` could work, but if any of the unit tests fail, the smoke tests won't run (although maybe that's ok?).
* We run mocha twice because we don't want the smoke grep filter applied to the unit tests.
* In `ci_smoke_win` the second mocha call is called in escaped quotes, this is needed for some reason after the `&`.  However, it is not valid for the first mocha invocation.

tolbertam:
Should take about ~5 minutes for a `ci_smoke_win` to run which seems reasonable to me.  We should probably let this run a bit to tease out any issues/unreliabilities exposed by running older versions of C* against ccm on windows.

tolbertam:
Experimenting with running smoke tests on travis as well.   Travis also supports os x builds, so I figured i'd give it a go.  One concern i have is it might not do too well running a 3 node ccm cluster, but we'll see.

tolbertam:
Looks like I have some work to do to get the travis os x build going - not sure how fully baked that support is either - will look into that more today.

tolbertam:
Looks to be working good on travis now for both osx and linux, still to determine:

* If adding `|smoke|` to description is the best way to identify smoke tests.
* If using 3 node clusters (mostly the prepare tests) is a good idea - especially given limited resources in travis environment (but seems to be working ok so far).
* If the tests identified are adequate for smoke testing (Is it took much? Not much?  Aiming for 5 minute build times)

jorgebay:
Its looking great!
To identify smoke test, we could use some specific BDD interface methods, like:
- `describeTag(name, tag, fn)`
- `vdescribe(cassVersion, name, optionalTag, fn)`
- `vit(cassVersion, name, optionalTag, testFn)`

Something like:
```javascript
vdescribe('2.0', '#batch()', 'smoke', function () {
  // ...
});
```

Otherwise, we could reorder the tests to include them in just 1 file (at least the ones that use a single node cluster): `test/integration/short/smoke-tests.js`

tolbertam:
I kind of like the idea of extending `vdescribe`/`vit`, etc. to include tags.    On the other hand I think I really like your idea of making a smoke-tests.js file with isolated smoke tests.  A majority of the tests can share 1 ccm cluster (other than ssl/ipv6 tests), which will make the tests run quicker and more reliably.  I'll look into that possibility.   We can probably keep the suite pretty small initially, exercising only basic functionality and extend it as we see fit.

tolbertam:
Rebased on master just for a sanity test for 3.1.5.

tolbertam:
Closing this for now, just so it's not listed as an open PR.  It works, but the Travis CI workers are under provisioned for OS X (long wait times and poorly performing).   I'll work to find an alternative way of verifying on these hosts.

tolbertam:
Moved to personal fork https://github.com/tolbertam/nodejs-driver/tree/NODEJS-304 for sake of keeping branch hygiene good.

PR #186, opened 2016-09-19
Fix read streaming frame flags

jorgebay:
Avoid parsing frame flags multiple times for streaming frames.
Refactor result parsing to attempt to make it more readable according to the spec.
Use a single catch block to handle range errors from all result kinds.

### DIFF
--- a/test/integration/short/client-each-row-tests.js
+++ b/test/integration/short/client-each-row-tests.js
@@ -533,10 +533,10 @@ describe('Client', function () {
         client.shutdown(done);
       });
     });
-    it('should retrieve large result sets in parallel', function (done) {
+    vit('2.0', 'should retrieve large result sets in parallel', function (done) {
       insertSelectTest(table, 50000, 20, 50000, { prepare: true }, done);
     });
-    it('should query multiple times in parallel with query tracing enabled', function (done) {
+    vit('2.0', 'should query multiple times in parallel with query tracing enabled', function (done) {
       insertSelectTest(table, 50000, 2000, 10, { prepare: true, traceQuery: true }, done);
     });
   });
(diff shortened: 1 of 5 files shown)


test/integration/short/client-each-row-tests.js
```diff
@@ -534,37 +534,10 @@ describe('Client', function () {
...
-          }, seriesNext);
-        }
-      ], function (err) {
-        client.shutdown();
-        done(err);
-      });
+      insertSelectTest(table, 50000, 50, 50000, { prepare: true}, done);
+    });
+    it('should query multiple times in parallel with query tracing enabled', function (done) {
+      insertSelectTest(table, 50000, 10000, 10, { prepare: true, traceQuery: true }, done);
```
tolbertam: is this done many times to attempt to reproduce the frame is not completely read in yet?

jorgebay: When sending multiple requests to read, it's likely that the response chunk received may contain an incomplete frame (the last one might be sliced in multiple chunks).

I'm trying to have incomplete frames with streaming enabled as a way to reproduce the original issue, but this may not occur on all runs...

test/integration/short/client-each-row-tests.js
```diff
@@ -567,10 +567,15 @@ function insertTestData(client, table, length, callback) {
...
-    consistency: types.consistencies.quorum,
-    fetchSize: rowLength
-  }});
+  var client = newInstance({
+    queryOptions: {
+      consistency: types.consistencies.quorum,
+      fetchSize: rowLength
+    },
+    socketOptions: {
+      readTimeout: 100000
```
tolbertam: This is probably a good idea given how unreliable the CI environments we use can be at times.

PR #190, opened 2016-10-06
Disable waiting for trace events in parallel query test

tolbertam:
`should query multiple times in parallel with query tracing enabled` was timing out in CI because of a change introduced in C* 2.2.8/3.0.9 ([CASSANDRA-11465](https://issues.apache.org/jira/browse/CASSANDRA-11465)) that causes C* to wait for trace events to be written before returning a response.   For this particular test we are not interested in the trace responses, which often timeout because of the amount of burden put on the events table from 2000 queries that each page several times.

### DIFF
--- a/test/integration/short/client-each-row-tests.js
+++ b/test/integration/short/client-each-row-tests.js
@@ -169,7 +169,9 @@ describe('Client', function () {
     before(function (done) {
       var client = newInstance();
       utils.series([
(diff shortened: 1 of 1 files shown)


tolbertam:
This may only need to be temporary, see: https://issues.apache.org/jira/browse/CASSANDRA-12754

jorgebay:
Makes sense

PR #191, opened 2016-11-10
TokenAwarePolicy: Use a pseudo random start index for yielding replicas

jorgebay:


lib/policies/load-balancing.js
```diff
@@ -319,39 +336,32 @@ TokenAwareIterator.prototype.iterate = function (callback) {
...
-    return {value: host, done: !host};
+  var host;
+  if (this.localReplicaIndex < this.localReplicas.length) {
+    host = this.localReplicas[(this.startIndex + (this.localReplicaIndex++)) % this.localReplicas.length];
+    return { value: host, done: false };
   }
-  //return child hosts
-  var item = this.childIterator.next();
-  if (!item.done) {
+  this.startIndex = -1;
```
tolbertam: Since the local and remote replicas are separate from each other, is there any reason why the same randomly generated start index couldn't be used for each (choosing the starting local and remote points)?

jorgebay: Good question, I went for that approach initially but it that causes uneven distribution of the replicas in the case the amount remote and local replicas differ, ie: 

- local replicas: 3; remote: 6; possible start indexes: `[0, 1, 2]`.

If we use a `rnd * total replicas`, we still can have an undesired distribution, ie: 

- local replicas 4; remote 6; possible start indexes [0...9], that for local replicas when applying `mod` translates to `[0, 1, 2, 3, 0, 1, 2, 3, 0, 1]`.

lib/policies/load-balancing.js
```diff
@@ -319,39 +336,32 @@ TokenAwareIterator.prototype.iterate = function (callback) {
...
+  // Return hosts from child policy
+  var item;
+  while ((item = this.childIterator.next()) && !item.done) {
     if (this.replicaMap[item.value.address]) {
-      return this.computeNext();
+      // Avoid yielding replicas from the child load balancing policy query plan
```
tolbertam: This isn't entirely related to this PR, but was wondering what you thought about this.  It looks like this ordering logic is the same as the previous implementation (though easier to follow).  

However, couldn't the child policy return non-replica remote hosts causing this to prioritize non-replica remote hosts before replica remote hosts?

The ordering seems like this:

1. Yield local replica hosts.
2. Yield non-replica hosts.
3. Yield remote replica hosts.

The java and python drivers behaves slightly differently:

1. Yield local replica hosts.
2. Delegate to child policy and filter out hosts that are not local replicas.

I think the optimal solution would be something like:

1. Yield local replica hosts.
2. Yield non-replica local hosts.
3. Yield remote replica hosts.
4. Yield non-replica remote hosts.

Although that seems kind of tricky because you would need to keep a separate structure to track non-replica remote hosts while iterating in step 2.  Also I think the benefits of prioritizing remote replica hosts are probably marginal once you only have remote replicas to choose, so maybe the java/python solution is probably the most simple/effective.

lib/policies/load-balancing.js
tolbertam: Aha, good point.  I hadn't considered that the local/remote replica count could be different!

lib/policies/load-balancing.js
```diff
@@ -294,15 +297,29 @@ TokenAwarePolicy.prototype.newQueryPlan = function (keyspace, queryOptions, call
  */
 function TokenAwareIterator(keyspace, queryOptions, replicas, childPolicy) {
```
tolbertam: Another thing I wanted to point out was that the java driver makes whether or not to randomize replicas a configurable parameter to the `TokenAwarePolicy` constructor.  I think the reason why is initially we introduced the parameter and kept the default behavior to not shuffle the replicas, and in the next major release we made shuffling the default.  I think in hindsight that probably didn't need to be done, and if it hadn't been done it probably would have never been a parameter.  So my thoughts are:

1. Is it a concern that we are changing the behavior in a patch release?   I think it is not a concern, I don't see why the user would care that the primary replicas was always considered first and everything was always done in the same order.  It only does harm to not shuffle, and it is an implementation detail.
2. Should this be configurable?  I think it shouldn't be, similar to my thoughts about \#1.   I don't see the use case for it *not* being random.

jorgebay: Good arguments.
Additionally, the driver exposes the `client.getReplicas()` that users does to retrieve the replicas and the ordering haven't changed (closest token and the following replicas) but in the case of `TokenAwarePolicy` I think that providing the closest replica first is a bug and we should always distribute the load.

lib/policies/load-balancing.js
jorgebay: I've just pushed a fix that does the following:

1. Yield local replica hosts.
1. Delegate to child policy and filter out hosts that are not local replicas.
1. Yields the remote replicas that haven't been selected by the child policy (if any)

tolbertam: With 2 now including remote replicas if emitted by the child policy, is 3 needed?  The only case where it would emit something is where the child policy doesn't contain the remote replicas, and if it doesn't there is probably a reason it doesn't (although you could use that same line of reasoning for the local replicas).

jorgebay: I was thinking that the child policy can still control if the driver should create connections to it by setting the distance, checking the host datacenter. In that way, the child policy implementer doesn't have to include all the nodes... but in the case that the child policy doesn't correctly handles distance for those replicas, it could lead to unexpected bebaviour (unwanted inter-dc traffic).
I will remove that logic.

jorgebay: Fixed to match the java/python behavior (only 1 & 2).

lib/policies/load-balancing.js
```diff
@@ -294,15 +297,28 @@ TokenAwarePolicy.prototype.newQueryPlan = function (keyspace, queryOptions, call
...
   this.keyspace = keyspace;
-  this.replicas = replicas;
   this.childPolicy = childPolicy;
   this.queryOptions = queryOptions;
-  this.childHosts = [];
-  this.remoteReplicas = [];
+  this.localReplicas = [];
+  this.localReplicaIndex = 0;
+  this.remoteReplicaIndex = 0;
```
tolbertam: small nit, remoteReplicaIndex isn't needed anymore.

jorgebay: fixed!

tolbertam: Excellent, thanks!  This looks great.  I have a few test updates coming very shortly, will be +1ing the PR soon.

PR #194, opened 2016-12-15
Support promises

jorgebay:
Some info:
- The _promisification_ of the methods is done explicitly one by one, it could have been done per prototype (like a script at the end of each module) but I think this way is easier for user to trace.
- Provided a way to "bring your own Promise module" by exposing a `promiseFactory` method. I've ran a quick microbenchmark using default factory and bluebird and I didn't find any significant difference in performance.
- For callback based invocations (current), it adds 2 extra method invocations but the expected performance is the same.
- Changed the `examples` folder to run only on nodejs versions greater than > 0.12. I've adapted all the examples to use promises as it is much easier to read than using an external control flow library like `async` and the callback based API.
- Adapted most of the docs for the same reason.
- Added a new doc entry for "Promise and callback-based API"

### DIFF
--- a/doc/features/promise-callback/README.md
+++ b/doc/features/promise-callback/README.md
@@ -1,0 +1,40 @@
+# Promise and callback-based API
+
+The driver supports both [promises][promise] and callbacks for the asynchronous methods exposed in the `Client` and
+`Metadata` prototypes, you can choose the approach that suits your needs.
+
+## Promise-based API
+
+```javascript
+client.execute('SELECT name, email FROM users')
+  .then(result => console.log('User with email %s', result.rows[0].email));
+```
+
+When a `callback` is not provided as the last argument, the driver will return a `Promise`, without the need to 
+_promisify_ the driver module. Returned promises are instances of [`Promise` global object][promise] and are created
+using the default constructor: `new Promise(executor)`.
+
+In case you want the driver use a third party `Promise` module (ie: [bluebird][bluebird]) to create the promise
+instances, you can optionally provide your own factory method when creating the `Client` instance, for example:
+
+```javascript
+const BbPromise = require('bluebird');
+const client = new Client({
+  contactPoints: contactPoints,
+  promiseFactory: BbPromise.fromCallback
+});
+```
+
+## Callback-based API
+
+All asynchronous methods of the driver supports an optional `callback` as the last argument.
+
+```javascript
+client.execute('SELECT name, email FROM users', function(err, result) {
+  assert.ifError(err);
+  console.log('User with email %s', result.rows[0].email);
+});
+```
+
+[promise]: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Promise
+[bluebird]: http://bluebirdjs.com/
--- a/lib/metadata/index.js
+++ b/lib/metadata/index.js
@@ -131,6 +131,10 @@ Metadata.prototype.buildTokens = function (hosts) {
 
 /**
  * Gets the keyspace metadata information and updates the internal state of the driver.
+ * <p>
+ *   If a <code>callback</code> is provided, the callback is invoked when the keyspaces metadata refresh completes.
+ *   Otherwise, it returns a <code>Promise</code>.
+ * </p>
  * @param {String} name Name of the keyspace.
  * @param {Function} [callback] Optional callback.
  */
@@ -166,6 +170,10 @@ Metadata.prototype._refreshKeyspaceCb = function (name, callback) {
 
 /**
  * Gets the metadata information of all the keyspaces and updates the internal state of the driver.
+ * <p>
+ *   If a <code>callback</code> is provided, the callback is invoked when the keyspace metadata refresh completes.
+ *   Otherwise, it returns a <code>Promise</code>.
+ * </p>
  * @param {Boolean} [waitReconnect] Determines if it should wait for reconnection in case the control connection is not
  * connected at the moment. Default: true.
  * @param {Function} [callback] Optional callback.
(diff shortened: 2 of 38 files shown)


jorgebay:
README and examples updated, docs still to-do.

doc/features/promise-callback/README.md
```diff
@@ -0,0 +1,40 @@
...
+using the default constructor: `new Promise(executor)`.
+
+In case you want the driver use a third party `Promise` module (ie: [bluebird][bluebird]) to create the promise
+instances, you can optionally provide your own factory method when creating the `Client` instance, for example:
+
+```javascript
+const BbPromise = require('bluebird');
+const client = new Client({
+  contactPoints: contactPoints,
+  promiseFactory: BbPromise.fromCallback
```
tolbertam: Tried this out and it didn't seem to work since what ultimately seems to happen is promiseFactory is passed a `function executor(resolve, reject)` and [`Promise.fromCallback`](http://bluebirdjs.com/docs/api/promise.fromcallback.html) expects a function callback that takes an error first:

> The resolver function is passed a callback that expects to be called back according to error-first node conventions.

Using `Promise.prototype.apply` seemed to work for me (although there's probably a better way).

jorgebay: Indeed, it wasn't working...
I fixed the `promiseFactory` signature to work as expected.

tolbertam: Awesome, fix is working great 👍

tolbertam:
Some preliminary numbers comparing using callbacks vs. promises (using throughput-test.js and throughput-test-promises.js on riptano/nodejs-driver-sut#2), including different promise factory implementations (es6 promises (default), bluebird and q):

### Write throughput test
compare_with_promises_write

### Read throughput test
compare_with_promises_read

As you can see there is some overhead when using promises, which I think we anticipated.  However, what I need to look into where the distribution of that overhead is coming from as it could be my approach to the promise version of the throughput tests.   I used bluebird to some what mimic async's `timesLimit` functionality, and maybe there are some ineffencies there.

Will do a more thorough analysis soon.

tolbertam:
> I used bluebird to some what mimic async's timesLimit functionality, and maybe there are some ineffencies in my approach there.

After doing some local profiling, it is very likely that this is the culprit, jumping out to GCE to rerun the test with a simplified version that is very close implementation wise to `throughput-test.js` but uses promises, instead of having an entirely promise based approach using bluebird.

test/integration/short/metadata-tests.js
```diff
@@ -272,6 +305,40 @@ describe('Metadata', function () {
...
+              next();
+            },
+            client.shutdown.bind(client)
+          ], done);
+        });
+        return;
+      }
+      it('should return the trace in a promise', function () {
+        var client = newInstance();
+        client.connect()
```
tolbertam: Noticed in a few places that the promise isn't being returned from the function, causing mocha to finish the test prematurely, will fix.

lib/metadata/index.js
```diff
@@ -158,11 +170,26 @@ Metadata.prototype.refreshKeyspace = function (name, callback) {
...
+    this._refreshKeyspacesCb(waitReconnect, cb);
+  });
+};
+
+/**
+ * @param {Boolean|Function} waitReconnect
+ * @param {Function} callback
+ * @private
+ */
+Metadata.prototype._refreshKeyspacesCb = function (waitReconnect, callback) {
```
tolbertam: One problem here is the logic at 195,  if someone simply calls `refreshKeyspaces()`, `waitReconnect` will be undefined, and this would set callback = undefined, and then set it to noop, so the promise never completes.

In the callback case, things work fine.  I think maybe we need to adjust the logic in refreshKeyspaces to check in the same way.

lib/metadata/index.js
```diff
@@ -174,29 +174,29 @@ Metadata.prototype._refreshKeyspaceCb = function (name, callback) {
  *   If a <code>callback</code> is provided, the callback is invoked when the keyspace metadata refresh completes.
  *   Otherwise, it returns a <code>Promise</code>.
  * </p>
- * @param {Boolean} [waitReconnect] Determines if it should wait for reconnection in case the control connection is not
+ * @param {Boolean|Function} [waitReconnect] Determines if it should wait for reconnection in case the control connection is not
  * connected at the moment. Default: true.
  * @param {Function} [callback] Optional callback.
  */
 Metadata.prototype.refreshKeyspaces = function (waitReconnect, callback) {
+  if (typeof waitReconnect === 'function' || typeof waitReconnect === 'undefined') {
```
jorgebay: Nice catch!

jorgebay:
Nice!
I think the benchmark results make sense. The impact of using Promises are greater according to [Bluebird's benchmark](https://github.com/petkaantonov/bluebird/tree/master/benchmark) but the difference gets smaller when you include the overhead of IO operations, parsing the messages, ...

It would be interesting to see the difference between using bluebird promisification of the `Client` methods in contrast with the built-in promise support. I think it could be a common question from the users: "What's the impact of using built-in promise support vs bluebird promisification".
I would expect slightly better results of promisifyAll-generated method vs built-in promise support (using `BluebirdPromise.fromCallback()`) but still far from callback baseline.

ie:

```javascript
client.execute(query);
// VS
BluebirdPromise.promisifyAll(Client.prototype);
client.executeAsync(query)
```

tolbertam:
Good idea, it didn't occur to me that I should try contrasting with promisfyAll but it makes sense.  Pretty easy for me to rerun the test at this point, so will do 👍

tolbertam:
Comparing using promisify with 3.1.6 and using bluebird with NODEJS-189, the throughput is pretty much exactly the same.  With that, I think we can feel confident in the performance of this feature 👍 

compare_with_promisify_write

compare_with_promisify_read

PR #196, opened 2016-12-22
Expose ResultSet @@iterator

jorgebay:


test/unit/address-resolution-tests.js
```diff
@@ -7,7 +7,7 @@ var addressResolution = require('../../lib/policies/address-resolution');
 var EC2MultiRegionTranslator = addressResolution.EC2MultiRegionTranslator;
 
 describe('EC2MultiRegionTranslator', function () {
-  this.timeout(5000);
+  this.timeout(10000);
```
tolbertam: I think this is the second time we've bumped this up, is weird how it can take > 5 secs at times, lets see if this fixes it :)

tolbertam:
looks nice and straightforward.  Will write a few integration tests, but don't imagine any problems.

tolbertam:
Should this work for paging?  When I use `autoPage: true` in my options it doesn't fetch on page boundaries.  Although, not sure how that can work in an async manner, so maybe that is for the best, however I can see that not being intuitive for users.

It looks like there is a [proposal](https://tc39.github.io/proposal-async-iteration/) for async iteration in tc39 that could be nice in the future.

I think we should document that this iterator will only work for fetched rows and that it will not page on boundaries.

jorgebay:
Auto page query option is only valid for `eachRow()` method, [the docs states that you can use `autoPage` for `eachRow()`](https://github.com/datastax/nodejs-driver/tree/master/doc/features/paging) but it doesn't specify that it is only supported there, I will add it in the jsdocs along with specifying that the `@@iterator` only works within the buffered results (first page).

About async iteration, we should start working on it soon as Babel supports `for await()` for some time. It looks challenging to implement it on ES5 but possible. On a side note, I don't know how much time we should support ES5 as [Node.js 0.10 and Node.js 0.12 reached OEL](https://github.com/nodejs/LTS).

PR #197, opened 2016-12-23
Expose isIdempotent query option

jorgebay:


doc/features/tuning-policies/README.md
```diff
@@ -143,20 +143,24 @@ A retry policy centralizes the handling of query retries, minimizing the need fo
...
+- `#onRequestError(info, consistency, err)`: defines whether to retry and at which consistency level on an 
 unexpected error, invoked in the following situations:
     - On a client timeout, while waiting for the server response , being the error an instance of 
     `OperationTimedOutError`.
     - On a connection error (socket closed, etc.).
     - When the contacted host replies with an error, such as `overloaded`, `isBootstrapping`, `serverError`, etc. In 
     this case, the error is instance of `ResponseError`
 
-A default and base retry policy is included.
+A default and base retry policy is included, along with `IdempotenceAwareRetryPolicy` that considers.
```
tolbertam: I think since we are doing a minor release, we can jump to what the java driver ultimately did, which is, if the query is non-idempotent, don't even pass it to the retry policy on a WriteTimeout / RequestError and log that this is happening the first time we encounter a WriteTimeout/Request Error for non idempotent operation:

https://github.com/datastax/java-driver/blob/3.1.2/driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java#L375-L380

https://github.com/datastax/java-driver/blob/3.1.2/driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java#L530-L540

jorgebay: I think there is a risk of breaking user code, as we mark all queries non-idempotent by default the driver won't sent use the retry policy for all cases.
I think going through the same transition (expose `IdempotenceAwareRetryPolicy` initially) as the java driver did is the safer route. We can change to the desired behaviour (ignore retrying logic for non-idempotent queries) on a major version. 

Being conservative with behavioural changes now will allow us to only maintain the latest 3.x branch, providing patches only to the latest minor and advising users to upgrade to the latest version (within the same major). Once we accumulate enough breaking changes (and the timing is right), we can provide a new major and maintain both 3.x and 4.x branches.

I've created a ticket for the breaking change: https://datastax-oss.atlassian.net/browse/NODEJS-331.

lib/client.js
```diff
@@ -144,6 +144,13 @@ var warmupLimit = 32;
  * @property {Number} [fetchSize] Amount of rows to retrieve per page.
  * @property {Array|Array<Array>} [hints] Type hints for parameters given in the query, ordered as for the parameters.
  * <p>For batch queries, an array of such arrays, ordered as with the queries in the batch.</p>
+ * @property {Boolean} [isIdempotent] Defines whether the query can be applied multiple times without changing the result
+ * beyond the initial application.
+ * <p>
+ *   The query execution idempotence can be used at [RetryPolicy]{@link module:policies/retry~RetryPolicy} level to
+ *   determine if an statement can be retried in case of replica unavailability or write timeout.
```
tolbertam: Shouldn't this be

> ...determine if a statement can be retried in case of request error or write timeout.

Unavailability always defers to child policy, which request error checks idempotence like write timeout.

doc/features/tuning-policies/README.md
```diff
@@ -143,20 +143,24 @@ A retry policy centralizes the handling of query retries, minimizing the need fo
...
+- `#onRequestError(info, consistency, err)`: defines whether to retry and at which consistency level on an 
 unexpected error, invoked in the following situations:
     - On a client timeout, while waiting for the server response , being the error an instance of 
     `OperationTimedOutError`.
     - On a connection error (socket closed, etc.).
     - When the contacted host replies with an error, such as `overloaded`, `isBootstrapping`, `serverError`, etc. In 
     this case, the error is instance of `ResponseError`
 
-A default and base retry policy is included.
+A default and base retry policy is included, along with `IdempotenceAwareRetryPolicy` that considers query idempotence.
```
tolbertam: Would be good to make a note here (or somewhere else, upgrade notes?) that in the future Non-idempotent queries will never be passed to retry policy in future.

lib/policies/retry.js
```diff
@@ -129,14 +130,66 @@ RetryPolicy.prototype.rethrowResult = function () {
...
+}
+
+util.inherits(IdempotenceAwareRetryPolicy, RetryPolicy);
+
+IdempotenceAwareRetryPolicy.prototype.onReadTimeout = function (info, consistency, received, blockFor, isDataPresent) {
+  return this._childPolicy.onReadTimeout(info, consistency, received, blockFor, isDataPresent);
+};
+
+/**
+ * If the query is not idempotent, it return a rethrow decision. Otherwise, it relies on the child policy to decide.
```
tolbertam: it return -> it returns

tolbertam:
I marked this as `notable` because even though this PR doesn't change behavior, users should still be aware of the change and future direction.  Retrying non-idempotent queries is never a good idea, but because we cannot reliably determine if a query is idempotent on our own (depend on user to set idempotence), the user should be aware of this.

doc/features/tuning-policies/README.md
jorgebay: good idea, I will do it!

PR #198, opened 2016-12-23
Linting

jorgebay:
- Add eslint config file
- lint lib source files
- Run eslint as part of the Travis build

Exclude `integer.js` from linting as we are not the authors.

- [x]  Used #197 as base branch to avoid conflicts on `lib/request-handler.js`, I will rebase once merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,17 @@
 language: node_js
 node_js:
+  - "7"
   - "6"
-  - "5"
   - "4"
   - "0.12"
   - "0.10"
-  - "iojs"
 sudo: false
 cache:
   directories:
     - node_modules
+script:
+  - if [ "$TRAVIS_NODE_VERSION" != "0.10" ] && [ "$TRAVIS_NODE_VERSION" != "0.12" ]; then
+      npm install -g eslint;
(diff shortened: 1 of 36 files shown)


tolbertam:
Great idea, I think we should either merge this first and then rebase/update all the other branches, or merge all the other branches first and then rebase and apply this.  wdyt?

jorgebay:
We can rebase it after #197 is merged.

.travis.yml
```diff
@@ -1,12 +1,17 @@
...
   - "0.10"
-  - "iojs"
 sudo: false
 cache:
   directories:
     - node_modules
+script:
+  - if [ "$TRAVIS_NODE_VERSION" != "0.10" ] && [ "$TRAVIS_NODE_VERSION" != "0.12" ]; then
+      npm install -g eslint;
+      eslint lib --ignore-pattern "/lib/types/integer.js";
```
tolbertam: should we lint tests too?  Or maybe should we save that for another PR?  (I can do the work for that if you think we should)

jorgebay: We should enable linting for tests too, this pr is only a first step, we can do it in another pr.

PR #199, opened 2017-01-10
NODEJS-333: Reduce number of configurations to build in jenkins on commit

tolbertam:
On commit for all branches create a 'per_commit' job that runs the
following configurations:

  * nodejs 0.10 with latest C*
  * nodejs 6 with C* 1.2 and latest

On primary branches, create a 'nightly' job that runs all
configurations nightly if there were any changes.

Note that the 'nightly' jobs wont be created until the build.yaml lands on those branches.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -1,8 +1,33 @@
+schedules:
+  commit:
+    # per commit job for all branches to run a subset of configs.
+    schedule: per_commit
+    matrix:
+      exclude:
+        # Exclude all 0.12 and 4 builds.
+        - nodejs: ['0.12', '4']
+        # Only build with latest for 0.10
+        - nodejs: '0.10'
+          cassandra: ['1.2', '2.0', '2.1', '2.2', '3.0']
+        # Only build with 1.2 and latest for 0.10
+        - nodejs: '6'
+          cassandra: ['2.0', '2.1', '2.2', '3.0']
+  nightly:
+    # nightly job for primary branches to run all configs.
+    schedule: nightly
+    branches:
+      # regex matches primary branch format (2.1, 3.x, 3.0.x, 3.1.x, master, etc).
+      include: ["/((\\d+(\\.[\\dx]+)+)|master)/"]
+  adhoc:
+    # adhoc job for non-primary braches that doesn't have a schedule but may be used to run all configs.
+    schedule: adhoc
+    branches: 
+      exclude: ["/((\\d+(\\.[\\dx]+)+)|master)/"]
 nodejs:
-  - "0.10"
-  - "0.12"
-  - "4"
-  - "6"
+  - '0.10'
+  - '0.12'
+  - '4'
+  - '6'
 os:
   - ubuntu/trusty64
 cassandra:
@@ -11,7 +36,7 @@ cassandra:
   - 2.1
   - 2.2
   - 3.0
-  - 3.7
+  - 3.9
 build:
   - type: envinject
     properties: |


build.yaml
```diff
@@ -1,8 +1,27 @@
...
+        - nodejs: '0.10'
+          cassandra: ['1.2', '2.0', '2.1', '2.2', '3.0']
+        # Only build with 1.2 and latest for 0.10
+        - nodejs: '6'
+          cassandra: ['2.0', '2.1', '2.2', '3.0']
+  nightly:
+    schedule: nightly
+    branches:
+      # regex matches primary branch format (2.1, 3.x, 3.0.x, 3.1.x, master, etc).
+      include: ["/((\\d+(\\.[\\dx]+)+)|(master))/"]
```
tolbertam: this regex is a bit crazy, but it works :)

tolbertam:
Example build matrix showing only 3 configurations ran:

screen shot 2017-01-10 at 2 44 57 pm

Does that look about right, or would you like different configurations built on commit?

build.yaml
jorgebay: The regex is not that crazy :)
There is an extra parenthesis around master, ie: (w/o escaping): `/((\d+(\.[\dx]+)+)|master)/`.
Is the escaping for `\` needed?

jorgebay:
Besides the regex comment, it's looking good!

One question: would be still possible to build the full matrix on demand?
For example, if we need to check a behaviour against an specific version of nodejs and we don't want to wait for nightly?

build.yaml
tolbertam: Yeah, I think there is definitely some redundancy there, primarily in surrounding a lot in parentheses where it's not needed (don't really need capture groups).   The escaping might not be needed either (it is for java but not for ruby iirc).  I'll tweak this.

tolbertam:
> For example, if we need to check a behaviour against an specific version of nodejs and we don't want to wait for nightly?

I think that capability would be nice for sure.  To enable that we could enable nightly build on branches as well and you could build those on demand.  They would only build nightly if there were any changes.    Alternatively, maybe there is a schedule type that never builds on a schedule, so an 'adhoc' job would be created for each branch that would never build unless a user triggered it?  wdyt?  I'll look and see if that is a possibility.

jorgebay:
> an 'adhoc' job would be created for each branch that would never build unless a user triggered it

You mean something like `.adhoc`, in the same way as `.nightly` / `.commit`, but it's not started automatically? That sounds like a good idea! If you see that it's too much effort, just including all runtimes (against latest C* version) for commit schedule would be good enough. It would be going from 24 jobs per commit to 4 or 5.

tolbertam:
> You mean something like .adhoc, in the same way as .nightly / .commit, but it's not started automatically? That sounds like a good idea! 

Exactly!  I think actually I will add an 'adhoc' schedule definition into our job creator as I think it's a good pattern in general.  Should be easy to do, will propose that to rest of team and come back to this ticket.

tolbertam:
Updated to add an 'adhoc' config as suggested.  This is only added for non-primary branches because the nightly build suffices for those (it can be triggered manually).  I also updated the regex as suggested, but the escaping was required for yaml.

image

tolbertam:
The change needed to support adhoc builds has been deployed to our jenkins server, so if everything looks good this is ready to merge.

jorgebay:
I see it on Jenkins, nice!

jorgebay:
Yay!
screen shot 2017-01-17 at 12 56 51

PR #200, opened 2017-01-11
Linting for tests

tolbertam:
- Add test specific globals to eslintrc
- Define 'eslint' script target for npm and have travis use it
- Add test directory to linting

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,6 @@ cache:
 script:
   - if [ "$TRAVIS_NODE_VERSION" != "0.10" ] && [ "$TRAVIS_NODE_VERSION" != "0.12" ]; then
       npm install -g eslint;
-      eslint lib --ignore-pattern "/lib/types/integer.js";
+      npm run eslint;
     fi
   - npm test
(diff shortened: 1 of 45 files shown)


test/unit/client-tests.js
```diff
@@ -1,3 +1,5 @@
+// remove strict declaration since this test does rewiring.
+// eslint-disable-next-line strict
```
jorgebay: We can still enable strict mode.
6 tests sets the value of the `Client` read-only property `profileManager`, which is not needed (can be removed).

It was my bad when implementing execution profiles, the manager was initially set on `connect()`. The final implementation uses a read-only property instead.

jorgebay:
WOW! This is great!
I only found the issue that I mention above on `unit/client-test.js` around strict mode, that I've made in the past so I'm going to send a fix for it and merge it.

.travis.yml
```diff
@@ -12,6 +12,6 @@ cache:
 script:
   - if [ "$TRAVIS_NODE_VERSION" != "0.10" ] && [ "$TRAVIS_NODE_VERSION" != "0.12" ]; then
       npm install -g eslint;
-      eslint lib --ignore-pattern "/lib/types/integer.js";
+      npm run eslint;
```
jorgebay: Much cleaner and easy to remember

PR #201, opened 2017-01-17
Duration type support

jorgebay:

PR #203, opened 2017-01-20
Test: Add helper.setup() method

jorgebay:
Reduce test boilerplate code by introducing a setup method. I've used already started using it in a couple of fixtures, but the idea is to start using it for new tests.

I'll rebase #201 after this is merged.


landim:
Looks good to me. Setup code on test side more clean. Nice.

PR #204, opened 2017-01-25
Expose protocol versions and make negotiation more resilient

jorgebay:
Introduced protocol versions and internal functions that check if a feature is supported by a provided protocol version.

I'm not very happy for including the protocol version "enum" in the types module, but as we are using instance members (`protocolVersion` is an object instance), I didn't want to pollute the root module. It's also consistent with the rest of object instances (as enums) exposed in the driver.

`protocolVersion` object is exposed and can be used when selecting a max protocol version:

```javascript
const client = new cassandra.Client({ 
  contactPoints: [ 'host1', 'host2' ],
  protocolOptions: { maxVersion: cassandra.types.protocolVersion.v3 }
});
```

It should work with any server version (w/ or wo/ CASSANDRA-11464)

### DIFF
--- a/lib/client-options.js
+++ b/lib/client-options.js
@@ -75,66 +75,104 @@ function extend(baseOptions, userOptions) {
         'the following values are valid contact points: ipAddress, hostName or ipAddress:port', i, hostName));
     }
   }
-  if (!options.policies) {
-    throw new TypeError('policies not defined in options');
-  }
-  if (!(options.policies.loadBalancing instanceof policies.loadBalancing.LoadBalancingPolicy)) {
-    throw new TypeError('Load balancing policy must be an instance of LoadBalancingPolicy');
-  }
-  if (!(options.policies.reconnection instanceof policies.reconnection.ReconnectionPolicy)) {
-    throw new TypeError('Reconnection policy must be an instance of ReconnectionPolicy');
-  }
-  if (!(options.policies.retry instanceof policies.retry.RetryPolicy)) {
-    throw new TypeError('Retry policy must be an instance of RetryPolicy');
-  }
-  if (!(options.policies.addressResolution instanceof policies.addressResolution.AddressTranslator)) {
-    throw new TypeError('Address resolution policy must be an instance of AddressTranslator');
-  }
-  if (options.policies.timestampGeneration !== null &&
-      !(options.policies.timestampGeneration instanceof policies.timestampGeneration.TimestampGenerator)) {
-    throw new TypeError('Timestamp generation policy must be an instance of TimestampGenerator');
+  if (!options.logEmitter) {
+    options.logEmitter = function () {};
   }
   if (!options.queryOptions) {
     throw new TypeError('queryOptions not defined in options');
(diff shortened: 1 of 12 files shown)


lib/client-options.js
```diff
@@ -220,15 +258,14 @@ function ifUndefined3(v1, v2, v3) {
...
-  var value = defaultValue;
   if (typeof userOptions.timestamp !== 'undefined') {
-    value = userOptions.timestamp;
+    return userOptions.timestamp;
   }
-  else if (client.controlConnection.protocolVersion > 2 && client.options.policies.timestampGeneration) {
-    // Use the timestamp generator
-    value = client.options.policies.timestampGeneration.next(client);
+  var timestampGenerator = client.options.policies.timestampGeneration;
+  if (types.protocolVersion.supportsTimestamp(client.controlConnection.protocolVersion) && timestampGenerator) {
```
tolbertam: Big fan of having functions for checking support instead of basing off the version number 👍

lib/types/protocol-version.js
```diff
@@ -0,0 +1,130 @@
+'use strict';
+
+/**
+ * Contains information for the different protocol versions supported by the driver.
+ * @type {Object}
+ * @property {Number} v1 Cassandra protocol v1, supported in Apache Cassandra 1.2-->2.2.
+ * @property {Number} v2 Cassandra protocol v2, supported in Apache Cassandra 2.0-->2.2.
+ * @property {Number} v3 Cassandra protocol v3, supported in Apache Cassandra 2.1-->3.x.
+ * @property {Number} v4 Cassandra protocol v4, supported in Apache Cassandra 2.2-->3.x.
+ * @property {Number} v5 Cassandra protocol v5, in beta from Apache Cassandra 3.x+. Currently not supported by the
```
tolbertam: We will have to remind ourselves to remove the `Currently not supported by the driver` part after we actually support v5 ;)

lib/types/protocol-version.js
```diff
@@ -0,0 +1,130 @@
...
+ * driver.
+ * @property {Number} maxSupported Returns the higher protocol version that is supported by this driver.
+ * @property {Number} maxSupported Returns the lower protocol version that is supported by this driver.
```
tolbertam: should be `minSupported` here

lib/types/protocol-version.js
```diff
@@ -0,0 +1,130 @@
...
+   * Determines whether the collection length is encoded using 32 bits.
+   * @param {Number} version
+   * @return {boolean}
+   * @ignore
+   */
+  uses4BytesCollectionLength: function (version) {
+    return (version >= this.v3);
+  },
+  /**
+   * When using certain versions
```
tolbertam: This doc line looks incompleted.

jorgebay:
Great! Pushed a fix for the jsdoc comments. 
I'll leave it open for another day to wait for feedback on the java-driver side.

tolbertam:
Doc changes look great, 👍

PR #205, opened 2017-01-27
3.0

utkarshut:
I am trying to get all the rows from cassandra database as there are huge number of rows i am doing streaming .

 client.stream(sql)
  .on('readable', function () {
    // 'readable' is emitted as soon a row is received and parsed
    var row;
    this.collection=[];
        console.log('readable executed');
    while (row = this.read()) {
        this.collection.push(row); 
        console.log("A");
    }
  })
  .on('end', function () {
    // Stream ended, there aren't any more rows
    console.log('end');
        console.log(this.collection);
        callback(this.collection);
  })
  .on('error', function (err) {
        console.log(err);
    // Something went wrong: err is a response error from Cassandra
  });   
};

But in output screen i am getting this.collection as empty array because after pushing all rows in this .collection it again print readable executed (in console screen) that set my this.collection as [ ]
. Why "on('readable', function ()"  is executing twice ?
How can i solve it or if their any alter way tell me asap

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,5 +237,5 @@ Unless required by applicable law or agreed to in writing, software distributed 
 [mailinglist]: https://groups.google.com/a/lists.datastax.com/forum/#!forum/nodejs-driver-user
 [jira]: https://datastax-oss.atlassian.net/projects/NODEJS/issues
 [streams2]: http://nodejs.org/api/stream.html#stream_class_stream_readable
-[cql-udt]: http://cassandra.apache.org/doc/cql3/CQL.html#createTypeStmt
+[cql-udt]: http://cassandra.apache.org/doc/latest/cql/types.html#user-defined-types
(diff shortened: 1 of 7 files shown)


jorgebay:
You can use the mailing list or JIRA to send questions or doubts.

About the readable event, it's normal that the event is fired more than once (when rows are ready), you can read more in the Node.js docs: https://nodejs.org/api/stream.html#stream_event_readable

Also, if you are retrieving large result sets, you should ready the paging doc: http://docs.datastax.com/en/developer/nodejs-driver/latest/features/paging/

utkarshut:
@jorgebay  
Thanks jorge for your support.
Kindly provide me mailing list for doubts

jorgebay:
https://github.com/datastax/nodejs-driver#getting-help

PR #207, opened 2017-02-08
Ping cassandra to check status for tests

landim:


test/test-helper.js
```diff
@@ -817,19 +818,33 @@ Ccm.prototype.waitForUp = function (callback) {
...
+    var ping = true;
+    utils.whilst(function () {
+      return nodesTested < self.nodeLength;
+    }, function pingNodes (nextIp) {
+      var pingIp = '127.0.0.' + (nodesTested + 1);
+      nodesTested++;
+      var conn = net.createConnection(9042, pingIp)
+            .on('error', function() {
+              ping = false;
+              conn.end();
```
jorgebay: No need to call `end()`, `allowHalfOpen` is false and the `error` event will be immediately fired: https://nodejs.org/api/net.html#net_event_error_1

test/test-helper.js
```diff
@@ -817,19 +818,33 @@ Ccm.prototype.waitForUp = function (callback) {
...
-      }
-      var regex = /Starting listening for CQL clients/mi;
-      started = regex.test(info.stdout.join(''));
+    var nodesTested = 0;
+    var ping = true;
+    utils.whilst(function () {
+      return nodesTested < self.nodeLength;
+    }, function pingNodes (nextIp) {
+      var pingIp = '127.0.0.' + (nodesTested + 1);
+      nodesTested++;
```
jorgebay: We could include in the trace that an attempt to connect to `pingIp:9042` is being made, ie: `helper.trace(...)`

test/test-helper.js
```diff
@@ -817,19 +818,33 @@ Ccm.prototype.waitForUp = function (callback) {
...
+      var conn = net.createConnection(9042, pingIp)
+            .on('error', function() {
```
jorgebay: Use 2 spaces for indentation.

test/test-helper.js
```diff
@@ -817,19 +818,33 @@ Ccm.prototype.waitForUp = function (callback) {
...
+      helper.trace('Attempt to ping ' + pingIp);
+      nodesTested++;
+      var conn = net.createConnection(9042, pingIp)
+        .on('error', function() {
+          ping = false;
+          nextIp();
+        })
+        .on('connect', function() {
+          ping = ping && true;
+          conn.end();
```
jorgebay: On a established connection, if the client ends (FIN) the connection before sending the STARTUP message, C*will error out before closing it.
This results in `'connect'` event, followed by `'error'`.

The fix should be something like this:
```
con
  .on('error', function () {
    if (ping) {
      // The connection was established, don't mind about errors after `socket.end()`
      return;
    }
    nextIp();
  })
  .on('connect', function () {
    ping = true;
    nextIp();
  });
```

landim:
There is no need for this validation as ccm is stable.

PR #193, opened 2016-11-21
Fix spelling in readme

rsp:
Some spelling fixes in README.md

PR #208, opened 2017-02-17
Fix spelling in readme - rebased

rsp:
This is the same as #193 but rebased on fresh master because since #193 was posted some conflicts appeared.


jorgebay:
lgtm! I will merge it as soon as travis finishes.

Thanks!

PR #209, opened 2017-03-15
Replace dead link with live library link

BenBrostoff:
It appears the [existing link to the Google Closure library is dead](http://docs.closure-library.googlecode.com/git/class_goog_math_Long.html). I linked [instead to the `long.js` library](https://github.com/dcodeIO/long.js ) I think the `nodejs-driver` uses.


jorgebay:
Thanks @BenBrostoff !

PR #210, opened 2017-03-23
Refactor and move error-tests to 'short'.

tolbertam:
Reuse the same cluster for all three tests, which makes them fast enough
to reclassify as 'short' tests.


jorgebay:
lgtm, I will merge it as soon as ci finishes.

PR #212, opened 2017-03-30
Optimize Murmur3Tokenizer using mutable longs

jorgebay:


tolbertam:
Did some preliminary profiling and this does seem to make quite a difference, nice!

Ran workload-standard-test in nodejs-driver-sut, which creates 4 byte keys that are somewhat random (uses `allocUnsafe`).  The amount of CPU time used in `Metadata.getReplicas` compared to the rest of the code decreases from 28.31% to 4.93% in this particular case:

3.2.1:

3 2 1_30secs

NODEJS-343-mutable-long:

nodejs-343_30secs

Note: 1st column is the 'self' time for the function itself, the 2nd column is the total time spent in the function.

tolbertam:
Some comparison numbers running workload-standard-test with a single node on a c4.xlarge:

boxplot-write-throughput

boxplot-read-throughput

As you'll see there is quite a noticeable throughput improvement when using TokenAwarePolicy with this change.  There is still some overhead in using TAP instead of DCAware by itself (which is expected), but the gap has closed quite a bit.

PR #213, opened 2017-04-07
Bugfix/nodejs 355 missing dot in name does not connect

guyellis:
Fixes confirmed bug from Jira [NODEJS-355](https://datastax-oss.atlassian.net/browse/NODEJS-355)

### DIFF
--- a/lib/control-connection.js
+++ b/lib/control-connection.js
@@ -98,7 +98,9 @@ ControlConnection.prototype.init = function (callback) {
         if (name.indexOf(':') > 0) {
           // IPv4 with port notation
           var parts = name.split(':');
-          return addHost(parts[0], parts[1], eachNext);
+          if (parts.length === 2) {
+            return addHost(parts[0], parts[1], eachNext);
+          }
         }
         if (net.isIP(name)) {
           return addHost(name, null, eachNext);
--- a/lib/control-connection.js
+++ b/lib/control-connection.js
@@ -95,7 +95,7 @@ ControlConnection.prototype.init = function (callback) {
           var portSeparatorIndex = name.lastIndexOf(']:');
           return addHost(name.substr(1, portSeparatorIndex - 1), name.substr(portSeparatorIndex + 2), eachNext);
         }
-        if (name.indexOf('.') > 0 && name.indexOf(':') > 0) {
+        if (name.indexOf(':') > 0) {
           // IPv4 with port notation
           var parts = name.split(':');
           return addHost(parts[0], parts[1], eachNext);
--- a/test/unit/control-connection-tests.js
+++ b/test/unit/control-connection-tests.js
@@ -69,7 +69,7 @@ describe('ControlConnection', function () {
         assert.ifError(err);
         var hosts = cc.hosts.values();
         assert.strictEqual(hosts.length, 2);
-        assert(hosts.indexOf('127.0.0.1:9999') >= 0);
+        assert.equal(hosts.filter(function (h) { return h.address === '127.0.0.1:9999'; }).length, 1);
         done();
       });
     });
--- a/test/unit/control-connection-tests.js
+++ b/test/unit/control-connection-tests.js
@@ -58,6 +58,21 @@ describe('ControlConnection', function () {
         done();
       });
     });
+    it('should resolve IPv4 and IPv6 addresses with non default port', function (done) {
+      if (!useLocalhost) {
+        return done();
+      }
+      var cc = new ControlConnection(clientOptions.extend({ contactPoints: ['localhost:9999'] }));
+      cc.getConnection = helper.callbackNoop;
+      cc.refreshOnConnection = helper.callbackNoop;
+      cc.init(function (err) {
+        assert.ifError(err);
+        var hosts = cc.hosts.values();
+        assert.strictEqual(hosts.length, 2);
+        assert.deepEqual(hosts.map(function (h) { return h.address; }).sort(), [ '127.0.0.1:9999', '::1:9999' ]);
+        done();
+      });
+    });
     it('should resolve all IPv4 and IPv6 addresses provided by dns.resolve()', function (done) {
       var ControlConnectionMock = rewire('../../lib/control-connection');
       ControlConnectionMock.__set__('dns', {
--- a/test/unit/control-connection-tests.js
+++ b/test/unit/control-connection-tests.js
@@ -69,7 +69,7 @@ describe('ControlConnection', function () {
         assert.ifError(err);
         var hosts = cc.hosts.values();
         assert.strictEqual(hosts.length, 2);
-        assert.deepEqual(hosts.map(function (h) { return h.address; }).sort(), [ '127.0.0.1:9999', '::1:9999' ]);
+        assert(hosts.indexOf('127.0.0.1:9999') >= 0);
         done();
       });
     });
--- a/test/unit/control-connection-tests.js
+++ b/test/unit/control-connection-tests.js
@@ -68,7 +68,7 @@ describe('ControlConnection', function () {
       cc.init(function (err) {
         assert.ifError(err);
         var hosts = cc.hosts.values();
-        assert.strictEqual(hosts.length, 2);
+        assert.strictEqual(hosts.length, 1);
         assert.equal(hosts.filter(function (h) { return h.address === '127.0.0.1:9999'; }).length, 1);
         done();
       });


jorgebay:
Sorry for the delay reviewing your pull request, we haven't forgot about it :)
I'll look at it this week!

lib/control-connection.js
```diff
@@ -95,7 +95,7 @@ ControlConnection.prototype.init = function (callback) {
           var portSeparatorIndex = name.lastIndexOf(']:');
           return addHost(name.substr(1, portSeparatorIndex - 1), name.substr(portSeparatorIndex + 2), eachNext);
         }
-        if (name.indexOf('.') > 0 && name.indexOf(':') > 0) {
+        if (name.indexOf(':') > 0) {
           // IPv4 with port notation
           var parts = name.split(':');
           return addHost(parts[0], parts[1], eachNext);
```
jorgebay: To avoid breaking ipv6 support, we should add a condition that parts is equals to 2.
ie: 
- `::1` -> ipv6
- `[::1]:9042` -> ipv6 with port already covered by previous condition.
- `whatever:9042` -> ipv4 or host with port

test/unit/control-connection-tests.js
```diff
@@ -58,6 +58,21 @@ describe('ControlConnection', function () {
...
+      if (!useLocalhost) {
+        return done();
+      }
+      var cc = new ControlConnection(clientOptions.extend({ contactPoints: ['localhost:9999'] }));
+      cc.getConnection = helper.callbackNoop;
+      cc.refreshOnConnection = helper.callbackNoop;
+      cc.init(function (err) {
+        assert.ifError(err);
+        var hosts = cc.hosts.values();
+        assert.strictEqual(hosts.length, 2);
```
jorgebay: I think asserting that it resolves at least to ipv4 loopback address (127.0.0.1) is good enough. Otherwise it may fail (ie: travis).

lib/control-connection.js
guyellis: Changes made and pushed.

test/unit/control-connection-tests.js
guyellis: Nothing needs to be changed here right?

guyellis:
@jorgebay Those failing tests look like a flaky setup on Travis. What do you think?

test/unit/control-connection-tests.js
jorgebay: My comment was not clear, sorry about that... I meant that it should check only for `contains '127.0.0.1:9999'`, ie: `hosts.map(h => h.address).indexOf('127.0.0.1:9999')`

guyellis: I've updated it to a truthy check on `hosts.indexOf('127.0.0.1:9999') >= 0`

Let me know if that's what you were getting at? Thanks.

jorgebay: The test is still failing because the `hosts.length` can be 1 or 2 depending on the system and because `hosts.indexOf('127.0.0.1:9999') >= 0` can not be `true` as `hosts` is an `Array<Host>`. Use [`Host#address`](http://docs.datastax.com/en/developer/nodejs-driver/3.2/api/class.Host/#address) instead.

guyellis: Interestingly the tests pass okay locally for me. Other tests fail though.

```
$ npm t

> cassandra-driver@3.2.1 test /home/guy/source/nodejs-driver
> mocha test/unit -R spec -t 5000

  ControlConnection
    #init()
      ✓ should resolve IPv4 and IPv6 addresses with non default port

  1 passing (49ms)
```

Would you expect a developer to be able to clone the repo, run `npm i` and then run `npm t` and for all the tests to pass?

guyellis: I've updated the test and pushed it. Let me know if that's what you're getting at?

guyellis:
@jorgebay I'd love to help further but there's obviously some type of setup that needs to be done on a developers machine before they can get all the tests to run successfully and I haven't been able to work out what that secret is yet...

```
  510 passing (17s)
  2 failing

  1) EC2MultiRegionTranslator #translate() should do a reverse and a forward dns lookup:

      Uncaught AssertionError: '92.242.140.2:9001' === '192.124.249.102:9001'
      + expected - actual

      -92.242.140.2:9001
      +192.124.249.102:9001
      
      at test/unit/address-resolution-tests.js:24:18
      at lib/policies/address-resolution.js:109:5
      at next (lib/utils.js:716:14)
      at lib/policies/address-resolution.js:102:9
      at GetAddrInfoReqWrap.asyncCallback [as callback] (dns.js:62:16)
      at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:78:10)

  2) Client #connect() should fail if no host name can be resolved:

      Uncaught AssertionError: Message was: All host(s) tried for query failed. First host tried, 92.242.140.2:9042: Error: Connection timeout. See innerErrors.
      + expected - actual

      -false
      +true
      
      at Client.<anonymous> (test/unit/client-tests.js:54:16)
      at connectFinished (lib/client.js:391:10)
      at next (lib/utils.js:713:14)
      at seriesFinished (lib/control-connection.js:139:5)
      at next (lib/utils.js:713:14)
      at done (lib/control-connection.js:163:5)
      at lib/control-connection.js:174:7
      at done (lib/control-connection.js:209:5)
      at next (lib/utils.js:855:14)
      at lib/utils.js:864:9
      at lib/control-connection.js:203:7
      at HostConnectionPool.afterCreating (lib/host-connection-pool.js:68:14)
      at whilstEnded (lib/host-connection-pool.js:151:21)
      at next (lib/utils.js:852:14)
      at HostConnectionPool.<anonymous> (lib/utils.js:864:9)
      at attemptOpenCallback (lib/host-connection-pool.js:201:19)
      at Connection.errorConnecting (lib/connection.js:217:3)
      at Socket.connectTimedOut (lib/connection.js:142:10)
      at Socket._onTimeout (net.js:352:8)



npm ERR! Test failed.  See above for more details.

```

jorgebay:
hm... that's weird. The implementation is done and the test looks good, I don't understand why its happening. I will look at it tomorrow and take it from there.

About the tests failing locally, it shouldn't, thanks for letting us know, we will address them to avoid bothering other contributors with env related failures.

Thanks for your contribution!!

tolbertam:
```
      Uncaught AssertionError: '92.242.140.2:9001' === '192.124.249.102:9001'
      + expected - actual

      -:9001
      +192.124.249.102:9001
```

I think since this test depends on DNS configuration, it could cause environmental failures.  Since these are unit tests we should try to remove that external dependency.  

```
      Uncaught AssertionError: Message was: All host(s) tried for query failed. First host tried, 92.242.140.2:9042: Error: Connection timeout. See innerErrors.
      + expected - actual

      -false
      +true
```

The odd thing about this test is that it must be resolving IP `92.242.140.2` for any host that can't be found (`not1.existent-host` shouldn't resolve to anything, although it's weird that datastax.com also resolved to that IP address).  Is it possible you are using a DNS server provided by your ISP?  I've seen some ISPs do that (so they can give you 'Not Found' web pages that have suggestions/ads).

lib/control-connection.js
```diff
@@ -95,10 +95,12 @@ ControlConnection.prototype.init = function (callback) {
           var portSeparatorIndex = name.lastIndexOf(']:');
           return addHost(name.substr(1, portSeparatorIndex - 1), name.substr(portSeparatorIndex + 2), eachNext);
         }
-        if (name.indexOf('.') > 0 && name.indexOf(':') > 0) {
+        if (name.indexOf(':') > 0) {
```
tolbertam: I might be wrong, but I think the intent of the code is such that if the notation is an ip address with a port, that we add it directly instead of resolving it as DNS (which is why it was failing for you before).  That was my thought given the comment at line 99 (`// IPv4 with port notation`).

Instead wouldn't it be better to split the contact point if it contains a colon into it's host and port name and then resolve the host (the `resolveAll` further below) with DNS and then add it with the port.

guyellis: @tolbertam You're spot on with my ISP (Cox) resolving the DNS so that it can show a page of adverts if you browse to anything that doesn't resolve. They even return a 200 status:

```
$ curl -I not1.existent-host
HTTP/1.1 200 OK
Server: nginx
Date: Mon, 15 May 2017 22:14:57 GMT
Content-Type: text/html; charset=utf-8
Connection: close
Expires: Mon, 15 May 2017 22:24:57 GMT
Cache-Control: max-age=600
```

guyellis: @tolbertam I see what you mean by the comment on line 99.

However if you trace the code execution you will see that the same condition was the way that `my.domain:9999` was getting "added" by the call to `addHost` inside that `if` condition and it is the reason why `my-domain:9999` is not getting added (if this fix is not applied).

This should not impact the IPv4 functionality of this section of code either. This change will allow all of these patterns to be added via the `addHost` call:
```
my-host:9999
my.host:9999
1.2.3.4:9999
```

Let me know what you think.

tolbertam: With the previous code `resolveAll` would be called with `my-host:9999` as `name` which wouldn't resolve anything from a DNS lookup.

What I was thinking was that if we only passed in the host part to resolveAll and then provided the port with each resolved IP address this would work and this way we would consistently always use DNS to resolve host names to IPs (whether we provide the port or not).   Here's [a patch](https://gist.github.com/tolbertam/6f27ca431ade44107b30915eccd47a94#file-nodejs-355-resolvewithport-patch) of what I was thinking.

guyellis: @tolbertam - I like that solution - good thinking.

jorgebay: nice catch @tolbertam ! I'll move this to a repo branch and we can apply your patch there.

jorgebay:
I've created a repo branch to incorporate changes from @guyellis and @tolbertam #217.

PR #214, opened 2017-04-24
Test reconnection attempts on decommissioned node

jorgebay:


jorgebay:
I think it makes sense moving it to long category... will do!

jorgebay:
hmmm... I think its a long test for something we can try to simulate otherwise (unit or with simulacron) or run it manually in the future if needed.

PR #217, opened 2017-05-17
Resolve domains without dots

jorgebay:
Includes changes in #213 by @guyellis and patch from @tolbertam

### DIFF
--- a/lib/control-connection.js
+++ b/lib/control-connection.js
@@ -95,10 +95,12 @@ ControlConnection.prototype.init = function (callback) {
           var portSeparatorIndex = name.lastIndexOf(']:');
           return addHost(name.substr(1, portSeparatorIndex - 1), name.substr(portSeparatorIndex + 2), eachNext);
         }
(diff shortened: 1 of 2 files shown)


jorgebay:
I've made a minor change in the test, to use `Array.prototype.filter` instead of `find` as it's not available in node <= 0.12.

The new approach looks good to me.

tolbertam:
> I've made a minor change in the test, to use Array.prototype.filter instead of find as it's not available in node <= 0.12.

D'oh!  Good catch 👍

jorgebay:
Merged, thanks @guyellis !

guyellis:
@tolbertam @jorgebay when will this get published to Npm?

jorgebay:
We don't have a target date yet but you can follow the progress of v3.2.2 release on JIRA: https://datastax-oss.atlassian.net/browse/NODEJS-358?filter=-4&jql=project%20%3D%20NODEJS%20AND%20fixVersion%20in%20(3.2.2)%20ORDER%20BY%20updated%20

PR #218, opened 2017-05-23
ControlConnection: reconnection refactoring

jorgebay:
Fixes: NODEJS-360

- Unified all the failover logic in a single method
- Reuse the query plan and move to the next host when there is a connection fail while reconnecting.
- Allow injection at constructor level
- Use a reconnection schedule instead of listening to `'up'` events from hosts as listening to events is more prone to race conditions.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -649,6 +649,7 @@ Client.prototype._shutdownCb = function (callback) {
     self.connected = false;
     self.isShuttingDown = true;
     var hosts = self.hosts.values();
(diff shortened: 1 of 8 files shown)


tolbertam:
Accidentally started doing all of my review on #219, so will continue doing so there (assume if I approve that then this is also approved).

jorgebay:
@tolbertam no problem, makes sense to review it as a single thing, we can close this one if the other is reviewed and merged.

jorgebay:
Merged as part of #219 !

PR #219, opened 2017-05-24
Reset ControlConnection when keyspace switch fails on init

jorgebay:
Fixes NODEJS-346.

It's implemented on top of #218 as both patches affect the `ControlConnection`, I'll rebase this once #218 is merged.

I've added a few more integration tests that check the state of the hosts but also had to run the following adhoc test as mocha will forcefully exit the process (we are not using mocha's `--no-exit` option):

```javascript
const client = new cassandra.Client({
  contactPoints: ['127.0.0.1'],
  keyspace: 'KS_DOES_NOT_EXISTS'
});

client.connect().catch(() => client.shutdown());
```

### DIFF
--- a/lib/control-connection.js
+++ b/lib/control-connection.js
@@ -77,7 +77,6 @@ function ControlConnection(options, profileManager, context) {
   this.nodeStatusChangeTimeout = null;
   this.reconnectionTimeout = null;
   this.hostIterator = null;
-  this.hostListenerRemover = null;
   this.triedHosts = null;
   if (context && context.borrowHostConnection) {
     this.borrowHostConnection = context.borrowHostConnection;
@@ -161,11 +160,18 @@ ControlConnection.prototype.init = function (callback) {
   });
 };
 
-ControlConnection.prototype.setHostListeners = function () {
+ControlConnection.prototype.setHealthListeners = function () {
   var host = this.host;
   var connection = this.connection;
   var self = this;
   var wasRefreshCalled = 0;
+
+  function removeListeners() {
+    host.removeListener('down', downOrIgnoredHandler);
+    host.removeListener('ignore', downOrIgnoredHandler);
+    connection.removeListener('socketClose', socketClosedHandler);
+  }
+
   function startReconnecting(hostDown) {
     if (wasRefreshCalled++ !== 0) {
       // Prevent multiple calls to reconnect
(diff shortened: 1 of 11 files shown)


lib/control-connection.js
```diff
@@ -378,38 +294,133 @@ ControlConnection.prototype.refreshHosts = function (newNodesUp, callback) {
...
-ControlConnection.prototype.hostDownHandler = function () {
-  this.log('warning', 'Host ' + this.host.address + ' used by the ControlConnection DOWN');
-  this.host = null;
-  this.connection = null;
-  this.refresh();
-};
-
 /**
- * Acquires a connection and refreshes topology metadata.
+ * Acquires a connection and refreshes topology and keyspace metadata.
```
tolbertam: Very nice doc, helps me understand the intent of this immensely.

lib/control-connection.js
```diff
@@ -378,38 +294,133 @@ ControlConnection.prototype.refreshHosts = function (newNodesUp, callback) {
...
+        if (err) {
+          self.log('error', 'ControlConnection could not retrieve a query plan to determine which hosts to use', err);
+          return next(err);
+        }
+        self.hostIterator = iterator;
+        next();
+      });
+    },
+    function getConnectionTask(next) {
+      self.borrowAConnection(next);
```
tolbertam: If initializing, won't none of the pools be established at this point?  Or does `borrowAConnection` create a connection somehow if not initialized?

tolbertam: I see, I think the act of calling `borrowConnection` on a Host with an empty pool will create a connection.

test/unit/control-connection-tests.js
```diff
@@ -133,85 +138,60 @@ describe('ControlConnection', function () {
...
-          return {
-            setDown: function () { downSet++; },
-            setDistance: function () { return types.distance.ignored; }
-          };
-        }
-      };
-      var event = { inet: { address: { toString: function () { return 'host1';}}}};
-      cc.nodeStatusChangeHandler(event);
-      assert.strictEqual(downSet, 1);
+    it('should continue iterating through the hosts when metadata retrieval fails', function (done) {
```
tolbertam: Nice, will be difficult to test at integration level so nice to have a unit test that exercises the behavior.

tolbertam:
> I've added a few more integration tests that check the state of the hosts but also had to run the following adhoc test as mocha will forcefully exit the process

Thanks for pointing that out,  I was looking at the new tests and did see (other than the host count check) that they were also passing on the old branch, but the key was whether or not everything was cleaned up.  Knowing that detail helped me understand.

As far as `--no-exit` goes, haven't forgotten about that and will revisit that at some point ([NODEJS-286](https://datastax-oss.atlassian.net/browse/NODEJS-286)) 👍 .

jorgebay:
> As far as --no-exit goes, haven't forgotten about that and will revisit that at some point (NODEJS-286)

Thats a tough one, especially because most tests were created relying on mocha exiting the process... :)

tolbertam:
Put this through some paces a little bit today and also ran a short endurance scenario where the control connection cycled 21 times over 3 hours (due to C* nodes going down).  Control connection was maintained throughout the entire time, the number of handles stayed stable, and the process exited when finished.  Comfortable 👍 'ing this for merge when convenient.

PR #220, opened 2017-05-26
Use query keyspace from prepared metadata

jorgebay:
- Fill in query keyspace from prepared metadata and use it from the token-aware policy.
- Refactor load-balancing integration tests to reuse a single 3-node cluster and assert that the replicas are being targeted by the lbp.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -179,6 +179,12 @@ var warmupLimit = 32;
  *   determine if an statement can be retried in case of request error or write timeout.
  * </p>
  * <p>Default: <code>false</code>.</p>
+ * @property {String} [keyspace] Specifies the keyspace for the query. Used for routing within the driver, this
+ * property is suitable when the query operates on a different keyspace than the current {@link Client#keyspace}.
+ * <p>
+ *   This property should only be set manually by the user when the query operates on a different keyspace than
+ *   the current {@link Client#keyspace} and using either batch or non-prepared query executions.
+ * </p>
  * @property {Boolean} [logged] Determines if the batch should be written to the batchlog. Only valid for
  * [Client#batch()]{@link Client#batch}, it will be ignored by other methods. Default: true.
  * @property {Buffer|String} [pageState] Buffer or string token representing the paging state.
@@ -1090,6 +1096,9 @@ Client.prototype._setQueryOptions = function (options, params, meta, callback) {
         //it is filled by the user
         //or it is not prepared
         return next();
+      }
+      if (!options.keyspace && meta.keyspace) {
+        options.keyspace = meta.keyspace;
       }
       if (util.isArray(meta.partitionKeys)) {
         //the partition keys are provided as part of the metadata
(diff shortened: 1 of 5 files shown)


jorgebay:
I'm looking at the build failures and I don't understand what is causing the ccm and test failures...

jorgebay:
I was able to reproduce the same issue locally, after running multiple times:

```
mocha test/integration/short/timeout-tests.js test/integration/short/load-balancing-tests.js
```

lib/client.js
```diff
@@ -179,6 +179,12 @@ var warmupLimit = 32;
  *   determine if an statement can be retried in case of request error or write timeout.
  * </p>
  * <p>Default: <code>false</code>.</p>
+ * @property {String} [keyspace] Specifies the keyspace for the query. Used for routing within the driver, this
```
tolbertam: This does seem useful for cases where you are executing a non-prepared query where the user provides the routing key explicitly and the keyspace cannot be discerned other than what is set on the connection (which may not be the same).   I think all drivers should maybe have an option for this (i.e. the java driver doesn't have this option as far as I know).

One thing that this could be conflated with eventually is [CASSANDRA-10145](https://issues.apache.org/jira/browse/CASSANDRA-10145) where at the protocol level we can set the keyspace for a statement.  That will be really nice to have, but one thing I worry about is if we introduce this query option now and then eventually overload it to also set the keyspace on the executed query at the protocol level, although on the other hand it wouldn't cause any issues because the user's intent after all is to declare that this is the keyspace the query uses.

tolbertam: Ah I stand corrected, in the java driver there is a way to set the keyspace on a simple statement for this exact case so I was incorrect about that.

jorgebay:
CI issues were related by incomplete query traces:
Query traces retrieval queries on `system_traces` keyspace are made using consistency level one, making it possible to return a trace incomplete `events` information.

This came as surprise to me, it's a bug and it affects Node.js, Java and C# client drivers at least: trace should be queried at at least LOCAL_QUORUM, I've filed NODEJS-370 for it.

The incomplete trace events caused the `times()` and `timesLimit()` loops to (which launch multiple parallel ops) throw an unhandled error. I've moved the assertion logic out of the loops and the test errors will no longer affect other areas of the test harness.

jorgebay:
As a workaround for the tests included in this pr, we could set the replication factor for `system_traces` keyspace to 1.

test/integration/short/load-balancing-tests.js
```diff
@@ -13,18 +13,20 @@ var WhiteListPolicy = loadBalancing.WhiteListPolicy;
...
-    queries: [
+  var setupQueries = [
       'CREATE KEYSPACE ks_simple_rp1 WITH replication = {\'class\': \'SimpleStrategy\', \'replication_factor\' : 1}',
       'CREATE KEYSPACE ks_network_rp1 WITH replication = {\'class\': \'NetworkTopologyStrategy\', \'datacenter1\' : 1}',
       'CREATE KEYSPACE ks_network_rp2 WITH replication = {\'class\': \'NetworkTopologyStrategy\', \'datacenter1\' : 2}',
       'CREATE TABLE ks_simple_rp1.table_a (id int primary key, name int)',
       'CREATE TABLE ks_network_rp1.table_b (id int primary key, name int)',
       'CREATE TABLE ks_network_rp2.table_c (id int primary key, name int)',
+    ];
+  if (helper.isCassandraGreaterThan('2.0')) {
```
jorgebay: I think that given that C* 1.2 is no longer a supported version, we can filter this test fixture out (ie: `vdescribe()`).

landim:
The failing test its not related to this content. So, I am 👍

PR #222, opened 2017-05-29
Fix Duration type encoding for nanos greater than 1 << 32

jorgebay:


landim:
Included test with the min and max range for 8 byte representation

landim:
There is already an PR to fix the inconsistency on the tests. The ones failing are not related to this PR. 👍

PR #223, opened 2017-05-30
Various test fixes

jorgebay:
Address recurrent CI failures.

### DIFF
--- a/test/integration/short/error-tests.js
+++ b/test/integration/short/error-tests.js
@@ -73,7 +73,8 @@ describe('Client', function () {
           });
         }
         else {
-          assert.strictEqual(err.failures, 1);
+          assert.strictEqual(typeof err.failures, 'number');
+          assert.ok(err.failures > 0);
         }
         assert.strictEqual(err.writeType, 'SIMPLE');
         done();
--- a/test/integration/short/error-tests.js
+++ b/test/integration/short/error-tests.js
@@ -12,20 +12,21 @@ var vdescribe = helper.vdescribe;
 describe('Client', function () {
   this.timeout(120000);
   vdescribe('2.2', 'with protocol v4 errors', function () {
-    var commonKs = helper.getRandomName('ks');
     var failWritesKs = helper.getRandomName('ks');
+    var failWritesTable = failWritesKs + '.tbl1';
     var setupInfo = helper.setup(2, {
-      keyspace: commonKs,
       clientOptions: {
         policies: { retry: new helper.FallthroughRetryPolicy() }
       },
       queries: [
         'CREATE TABLE read_fail_tbl(pk int, cc int, v int, primary key (pk, cc))',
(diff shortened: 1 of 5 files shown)


test/integration/short/error-tests.js
```diff
@@ -12,20 +12,21 @@ var vdescribe = helper.vdescribe;
...
       queries: [
         'CREATE TABLE read_fail_tbl(pk int, cc int, v int, primary key (pk, cc))',
+        helper.createKeyspaceCql(failWritesKs, 2, true),
+        helper.createTableCql(failWritesTable),
         helper.createKeyspaceCql('ks_func'),
         'CREATE TABLE ks_func.tbl1 (id int PRIMARY KEY, v1 int, v2 int)',
         'INSERT INTO ks_func.tbl1 (id, v1, v2) VALUES (1, 1, 0)',
-        "CREATE FUNCTION ks_func.div(a int, b int) RETURNS NULL ON NULL INPUT" +
-        " RETURNS int LANGUAGE java AS 'return a / b;'"
+        'CREATE FUNCTION ks_func.div(a int, b int) RETURNS NULL ON NULL INPUT RETURNS int LANGUAGE java AS' +
```
landim: single quote through all the code. 👍

test/integration/short/metadata-tests.js
```diff
@@ -250,10 +250,10 @@ describe('Metadata', function () {
         client.connect.bind(client),
         function executeQuery(next) {
           client.execute(helper.queries.basic, [], { traceQuery: true}, function (err, result) {
-            traceId = result.info.traceId;
             if (err) {
               return next(err);
             }
+            traceId = result.info.traceId;
```
landim: nice catch!

test/unit/stream-id-stack-tests.js
```diff
@@ -62,8 +63,8 @@ describe('StreamIdStack', function () {
     stack.clear();
   });
   it('should release unused groups', function (done) {
-    var releaseDelay = 50;
-    var stack = newInstance(3, releaseDelay - 10);
+    var releaseDelay = 100;
```
landim: I question myself if is it better to use a fixed value or a relative (percentage)

jorgebay: The `osPrecision` can be made into a helper property and reused, as any value greater than 15.6 (1000 / 64) will work on modern Windows, Linux and macOS.
Any interval greater than twice the precision for the delay would work, ie: `delay ± precision`

jorgebay:
hmm... it looks like there are still test failures related to auth. I don't understand why its happening as `-Dcassandra.superuser_setup_delay_ms=0` is set, maybe its not being correctly passed??

I see on the C# driver that a similar setup, is not failing: https://github.com/datastax/csharp-driver/blob/master/src/Cassandra.IntegrationTests/Core/SessionAuthenticationTests.cs#L39-L60

The only difference is the 10s vs 5s additional wait time.

PR #224, opened 2017-06-02
TokenAware lb policy test fixes

tolbertam:
Please hold off on merging for now, will have another fix that addresses a rare issue with traces.

### DIFF
--- a/test/integration/short/load-balancing-tests.js
+++ b/test/integration/short/load-balancing-tests.js
@@ -14,11 +14,13 @@ var vdescribe = helper.vdescribe;
 
 context('with a reusable 3 node cluster', function () {
   this.timeout(180000);
-  helper.setup(3, {
+  // pass in 3:0 to exercise CCM dc set up logic which will use a consistent data center name
+  // for both Apache Cassandra and DSE.
+  helper.setup('3:0', {
     queries: [
       'CREATE KEYSPACE ks_simple_rp1 WITH replication = {\'class\': \'SimpleStrategy\', \'replication_factor\' : 1}',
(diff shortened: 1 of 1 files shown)


tolbertam:
Removing wip tag, can't get the tracing issues to occur in openstack environment.  This fix is needed for DSE compatibility so i think it can be merged as is while i continue to wrap my head around the tracing issues.

tolbertam:
Ok so i was able to reproduce the other issue i was digging into in an openstack environment.  I think it may be related to the replica selection being random in token aware policy.   With 10 attempts, the possibility of the same replica (of 2 replicas) being chosen each time is 1/1024 ((1/2)^10) which is pretty low.  

But there are 10 series of 10 attempts in a test (5/512), and then 2 varieties of this test (5/256), and we have a nightly matrix of 24 configurations , which reduces the probability of this occurring in our nightly matrix to 15/32 or roughly 1/2 which seems about right.

I think to make the test more reliable we need to increase the number of queries executed per iteration.  Setting it to 100 will make the test take longer (not really noticeable), but also more reliable as the starting basis probability will be 1/2^100 which is a very low probability.

tolbertam:
Github must be having problems because my commit isn't showing up on the PR, but once that shows up this should be good to go

tolbertam:
closed/reopened and now the commit shows up.

jorgebay:
> But there are 10 series of 10 attempts in a test (5/512), and then 2 varieties of this test (5/256), and we have a nightly matrix of 24 configurations , which reduces the probability of this occurring in our nightly matrix to 15/32 or roughly 1/2 which seems about right.

Great development of something that is not evident at first. Its funny how choosing an "intermediate" number of attempts made it worst: the failure didn't appear on most "commit" builds but was still very likely to appear on nightly builds. For example, if I used `20` instead of `10` initially would have make it very unlikely on nightly as well.

PR #225, opened 2017-06-04
Fix features doc readme

tmitim:
- Fixes the link for dataypes that was incorrectly directed to /connection-pooling.
- Adds missing feature links.
- Reorders links by path name for continuity.


jorgebay:
Thanks @tmitim !

PR #227, opened 2017-06-27
NODEJS-373: Decode 0 length values in maps

tolbertam:
For [NODEJS-373](https://datastax-oss.atlassian.net/browse/NODEJS-373)

### DIFF
--- a/lib/encoder.js
+++ b/lib/encoder.js
@@ -1408,10 +1408,10 @@ function getLengthBufferV3(value) {
   }
   var lengthBuffer = new Buffer(4);
   if (typeof value === 'number') {
-    lengthBuffer.writeUInt32BE(value, 0);
+    lengthBuffer.writeInt32BE(value, 0);
   }
   else {
-    lengthBuffer.writeUInt32BE(value.length, 0);
+    lengthBuffer.writeInt32BE(value.length, 0);
   }
   return lengthBuffer;
 }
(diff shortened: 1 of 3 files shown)


lib/encoder.js
```diff
@@ -1442,7 +1442,7 @@ function handleBufferRef(buffer) {
  * @private
  */
 function decodeCollectionLengthV3(bytes, offset) {
-  return bytes.readUInt32BE(offset);
+  return bytes.readInt32BE(offset);
```
tolbertam: Noticed that this is actually a signed integer ([java driver code](https://github.com/datastax/java-driver/blob/3.x/driver-core/src/main/java/com/datastax/driver/core/CodecUtils.java#L63-L75), [cassandra code](https://github.com/apache/cassandra/blob/cassandra-3.0/src/java/org/apache/cassandra/transport/CBUtil.java#L417)) although not sure if this actually matters since collections shouldn't contain null values.

jorgebay: I think it doesn't affect the end result in this case, but int (not uint) is the correct type according to the protocol, so it makes sense to change it: https://git-wip-us.apache.org/repos/asf?p=cassandra.git;a=blob_plain;f=doc/native_protocol_v4.spec;hb=refs/heads/trunk

I think we should also change it on `getLengthBufferV3()` when writing it.

jorgebay: Would you like to adapt `getLengthBufferV3()` just for the sake of symmetry, as part of this patch?

tolbertam: Ah, forgot about that, sure i'll take care of that now.

PR #228, opened 2017-06-29
Pre-calculate string versions of tokens to reduce execution time

peterjroberts:
With high token counts the beginning of the inner loop (up the the first continue) can require executions thousands/millions of times. The stringify operation is by far the most expensive part of this section of the loop and is performed repeatedly on the same values.

The tokens can instead be stringified once outside the main loop.


jorgebay:
makes sense!

Going one step further, we could pre-calculate it for all keyspaces on `Metadata#buildTokens()`, that way the tokenToReplicaMappers could take 

`tokenizer, ringTokensAsStrings, primaryReplicas, datacenters`

 as parameters instead of

`tokenizer, ring, primaryReplicas, datacenters`.

What do you think?

peterjroberts:
At first look that seems like a good idea. Would you have any concerns about breaking compatibility with possible third-party mappers?

jorgebay:
The token-to-replicas mapper per keyspace is not part of the public API and third-party partitioners are not supported in the driver.

peterjroberts:
I've moved the stringification of the tokens up to `Metadata#buildTokens()` as suggested. What do you think?

jorgebay:
Its looking good but I'm not able to identify at first sight what's making [`load-balancing-tests.js` to fail](https://ci.appveyor.com/project/DataStax/nodejs-driver/build/848.master), I'll do a full review by the end of this week.

lib/metadata/index.js
```diff
@@ -125,6 +125,11 @@ Metadata.prototype.buildTokens = function (hosts) {
   this.primaryReplicas = primaryReplicas;
   //All the tokens in ring order
   this.ring = allSorted;
+  //Compute string versions as it's potentially expensive and frequently reused later
+  this.ringTokensAsStrings = [];
```
jorgebay: Use `Array` constructor with the fixed length.

lib/metadata/index.js
```diff
@@ -125,6 +125,11 @@ Metadata.prototype.buildTokens = function (hosts) {
...
+  for (var i = 0; i < allSorted; i++) {
```
jorgebay: I think you meant `i < allSorted.length`, as this always evaluates as `false`.

peterjroberts: Oh that's embarrassing. 

I've had a look through and can't see any unit tests any unit tests around this function which I could have modified to pick this error up. Am I missing it or is it just covered in the ci/integration tests (which I haven't managed to get running locally)?

jorgebay:
I've created a JIRA ticket to keep track of this: [NODEJS-376](https://datastax-oss.atlassian.net/browse/NODEJS-376).

lib/metadata/index.js
jorgebay: It's only covered in the integration tests (`test/short/load-balancing-tests.js`). If you want to run them locally you must install [ccm](https://github.com/pcmanus/ccm) or you can let CI do that for you :)

We could include a specific unit test for it on `test/unit/metadata-tests.js` for `buildTokens()` making sure that the necessary values are populated, ie: `primaryReplicas`, `ring`, `ringTokensAsStrings`, ...
Do you think you can include it as part of this patch? otherwise, I could do it.

jorgebay:
I'll wait for CI to finish and merge it.

lib/metadata/index.js
peterjroberts: As you've seen, I added a couple of tests around this. Only covering the ring/ringTokensAsStrings right now as I wouldn't be confident saying what the other values *should* be.

jorgebay: Thanks! I think its a good step to have at least those 2 covered in the unit tests.

jorgebay:
Merged! Landed in cdcd261cf862a998680f80b65b6e5cc27128cb19, thanks @peterjroberts!

PR #230, opened 2017-08-21
Fix buildTokens() tests

jorgebay:


jorgebay:
Fix test broken on master.
Override all tokenizer methods.

PR #231, opened 2017-08-28
Support prepare on all hosts and re-preparing on UP

jorgebay:
Encapsulated the query preparation logic in a PrepareHandler.


jorgebay:
While this pr is useful to start testing the new prepare behaviour (ie: introduce simulacron), I don't think it makes sense to review it for now as most of the `RequestHandler` changes will be affected by NODEJS-82.
We can hold off the review until NODEJS-82 is implemented and review both as a whole.

landim:
Integartion tests with simulacron at https://github.com/datastax/nodejs-driver/pull/237

PR #232, opened 2017-09-01
Speculative executions

jorgebay:
Introduced `RequestExecute` class to encapsulate a single flow of execution against a coordinator, handling individual retries and failover.
`RequestHandler` now contains one or more `RequestExecution` instances depending on the SpeculativeExecutionPolicy and whether or not the statement is idempotent.

I've implemented it on top of #231 to tackle this as a single review: there have been several changes on `RequestHandler`, due to individual execution handling AND moved most of the query preparation logic to `PrepareHandler`.


lib/operation-state.js
```diff
@@ -0,0 +1,140 @@
...
+ * @return {boolean}
+ */
+OperationState.prototype.isByRow = function () {
+  return this._rowCallback && (this.request instanceof ExecuteRequest || this.request instanceof QueryRequest);
+};
+
+/**
+ * Checks if the operation hasn't been already cancelled.
+ */
+OperationState.prototype.canBeSent = function () {
```
jorgebay: Not being used

lib/request-execution.js
```diff
@@ -0,0 +1,227 @@
...
+RequestExecution.prototype._retry = function (consistency, useCurrentHost) {
+  if (this._cancelled) {
+    // No point in retrying
+    return;
+  }
+  this._parent.log('info', 'Retrying request');
+  this._retryCount++;
+  if (typeof consistency === 'number') {
+    // TODO: Clone
+    this._parent.request.consistency = consistency;
```
jorgebay: Needs to be cloned

lib/policies/speculative-execution.js
```diff
@@ -0,0 +1,110 @@
...
+ * @classdesc
+ * A {@link SpeculativeExecutionPolicy} that schedules a given number of speculative executions,
+ * separated by a fixed delay.
+ * @constructor
+ * @param {Number} delay The delay between each speculative execution. Must be strictly positive.
+ * @param {Number} maxSpeculativeExecutions The number of speculative executions. Must be strictly positive.
+ * @extends {SpeculativeExecutionPolicy}
+ */
+function ConstantSpeculativeExecutionPolicy(delay, maxSpeculativeExecutions) {
+  if (!(delay > 0)) {
```
tolbertam: One upcoming change in the java driver that would be nice to carry along here is to allow 0 ms delay between speculative executions ([JAVA-1490](https://github.com/datastax/java-driver/pull/856)).  For users who need very quick response SLAs (i.e. 5ms), the ability to send multiple requests out immediately helps quite a bit.  The other nice part about this is you could special case 0ms delays by sending the requests immediately rather than scheduling them on a timer (which may have a perceived delay as well).

lib/request-execution.js
```diff
@@ -0,0 +1,230 @@
...
+var utils = require('./utils');
+
+var retryOnNextHostDecision = Object.freeze({
+  decision: retry.RetryPolicy.retryDecision.retry,
+  useCurrentHost: false,
+  consistency: undefined
+});
+
+/**
+ * Represents an execution of a Request against a single
```
tolbertam: is rest of comment missing?  Assuming means `against a single host`.

lib/request-execution.js
```diff
@@ -0,0 +1,230 @@
+'use strict';
```
tolbertam: Great idea breaking this out of request-handler.  Makes code more contained.

lib/request-handler.js
```diff
@@ -4,469 +4,233 @@ var util = require('util');
...
+RequestHandler.prototype._startNewExecution = function () {
+  var execution = new RequestExecution(this);
+  this._executions.push(execution);
   var self = this;
-  this.connection.sendStream(request, options, function readCallback(err, response) {
-    if (err) {
-      //Something bad happened, maybe retry or do something about it
-      return self._handleError(err, callback);
+  execution.start(function hostAcquired(host) {
+    if (!self.requestOptions.isIdempotent) {
```
tolbertam: Shouldn't this only be checked after the 1st execution?

jorgebay: The first execution has started, the check is made to schedule the next timeout.

tolbertam: Ah I see,  I was following the code wrong.  Had interpreted it that this included the first execution, but if that is not the case then this makes sense.

test/integration/short/client-pool-tests.js
```diff
@@ -641,6 +658,68 @@ describe('Client', function () {
...
+                  helper.ccmHelper.resumeNode(1);
+                }, 400);
+              }
+              utils.map([ false, true ], function execute(prepare, mapNext) {
+                client.execute('SELECT * FROM system.local', null, { prepare: prepare, isIdempotent: true }, mapNext);
+              }, function (err, results) {
+                assert.ifError(err);
+                assert.strictEqual(results.length, 2);
+                var expectedHost;
+                if (policy instanceof ConstantSpeculativeExecutionPolicy) {
```
tolbertam: It would be nice if there were a way to track whether or not there were multiple speculative executions executed on a request like a `speculativeExecutions` field which just simply had the number of executions.  This would offer a way to track whether or not speculative executions were happening.  For example, I see that you can't easily tell if a request was submitted to another host if we get a response for the first request first since it won't show up in `triedHosts` (from what I can tell so far).

tolbertam: Hrmm maybe i'm wrong about this.  I had a configuration issue where not all queries being sent were idempotent (so some weren't speculatively retried).  It looks like triedHosts would represent ones where speculative executions have begun, so perhaps a separate identifier isn't needed, unless we wanted to be able to distinguish between retries and speculative executions.

tolbertam: I made a patch for helping me count speculative execution ratio, not sure if worth incorporating the idea: https://gist.github.com/tolbertam/69434846c4fcf5bb04f7f12783c5a8f1

jorgebay: The patch looks good to me, push it!

tolbertam: Will do, thanks!

lib/request-execution.js
```diff
@@ -0,0 +1,231 @@
...
+  }
+  var decisionInfo = this._getDecision(err);
+  if (err.isSocketError) {
+    this._host.removeFromPool(this._connection);
+  }
+  if (!decisionInfo || decisionInfo.decision === retry.RetryPolicy.retryDecision.rethrow) {
+    if (this._request instanceof requests.QueryRequest || this._request instanceof requests.ExecuteRequest) {
+      err['query'] = this._request.query;
+    }
+    return this._parent.setCompleted(err);
```
jorgebay: `this._operation.setResult(err)` should be used instead, as we need to check that the operation hasn't been already cancelled.

There should be only one call to `this._parent.setCompleted()` and that is on `this._sendOnConnection().

jorgebay: Don't mind, the correct flow is:
- `Connection` should set the `OperationState` as completed or timed out.
- `RequestExecution` should signal the parent `RequestHandler` as completed.

tolbertam: This seems to be working great so far.  Have been running test for about 8 hours without issues with a 1.65% speculative execution rate.

test/integration/simulacron.js
```diff
@@ -0,0 +1,403 @@
...
+SimulacronCluster.prototype._getPath = function (endpoint, id) {
+  var path = '/' + endpoint + '/' + this.id;
+  if (id) {
+    path += '/' + id;
+  }
+  return encodeURI(path);
+};
+
+SimulacronCluster.prototype.log = function(id, callback) {
+  if (typeof id === 'function') {
```
tolbertam: I will find a better pattern for this as is done in a lot of methods (if only callback is passed, assume scope global).   Also there's some other repetition i'd like to clean up.

test/integration/short/speculative-execution-simulator-tests.js
```diff
@@ -0,0 +1,448 @@
...
+      }
+    };
+    var setupInfo = simulacron.setup('3', { clientOptions: clientOptions });
+    var client = setupInfo.client;
+    var cluster = setupInfo.cluster;
+    var query = "select * from data";
+    
+    var assertQueryCount = queryCounter(query, cluster);
+
+    it('should not start speculative executions if query is non-idempotent', function (done) {
```
tolbertam: This is just a start.  I looked at the java-driver 4.x tests for inspiration for the most part.  There's a lot of code duplication among the tests, will clean that up.

jorgebay:
I've rebased it onto master.

test/integration/short/speculative-execution-simulator-tests.js
```diff
@@ -49,9 +58,9 @@ describe('Client', function() {
             assert.strictEqual(queriedHost.nodeId, 0);
            
             // Should have only sent request to node 0.
-            assertQueryCount(0, 1);
-            assertQueryCount(1, 0);
-            assertQueryCount(2, 0);
+            assertQueryCount(node0, 1);
```
tolbertam: Need to fix this code, the assertions are done in callback which is out of band.

test/integration/simulacron.js
```diff
@@ -100,6 +116,42 @@ var simulacronHelper = {
...
+  /**
+   * Convenience function for setting up simulacron, starting a simulated cluster and optionally creating a client instance
+   * that connects to it in before hooks.  Cleans up after itself in after hooks.
+   * 
+   * @param {Array} dcs array of nodes-per-dc configuration (i.e. 2,2,2 creates 3 2 node dcs)
+   * @param {Object} [options]
+   * @param {Object} [options.initClient] Determines whether to create a Client instance.
+   * @param {Object} [options.clientOptions] The options to use to initialize the client.
+   */
+  setup: function (dcs, options) {
```
jorgebay: Nice!

test/integration/simulacron.js
```diff
@@ -111,200 +163,352 @@ var simulacronHelper = {
...
+ * Finds a node in the cluster by its id or address.
+ * 
+ * @param {Number|String} id Identifier of node.  If Number, assumes the id of the node in the data center.
+ * If String, assumes an 'ip:port' designation and looks up node by address.
+ * @param {Number} id2 Second identifier.  If present, 'id' is assumed to be the data center id.  If not
+ * present, 'id' is assumed to be the node id or address.  If number is passed in for 'id', 0 is the assumed
+ * data center id.
+ * @returns {SimulacronNode} The node, if found.
+ */
+SimulacronCluster.prototype.node = function() {
```
jorgebay: Having the parameters defined in the function declaration will make IDEs happier :)

jorgebay: We could simplify it a bit if we use `node = function (key, dc){}`, where `key` can be either the address or an index, and `dc` is optional.

test/integration/simulacron.js
```diff
@@ -111,200 +163,352 @@ var simulacronHelper = {
...
     }
+    dc = this.dcs[dcId];
+    return dc.node(nodeId);
   }
 };
 
+/**
+ * Finds a data center in the cluster by its id.
+ * 
+ * @param {Number} id Idenfitifer of the dc.
```
jorgebay: Typo

test/integration/simulacron.js
```diff
@@ -100,6 +116,42 @@ var simulacronHelper = {
...
+    if (initClient) {
+      var baseOptions = { contactPoints: [self.startingIp] };
+      client = new Client(utils.extend({}, options.clientOptions, baseOptions));
+      before(client.connect.bind(client));
+      after(client.shutdown.bind(client));
+    }
+    afterEach(simulacronCluster.clear.bind(simulacronCluster));
+    after(simulacronHelper.stop.bind(simulacronHelper));
+
+    var setupInfo = { cluster: simulacronCluster, client: client };
```
jorgebay: Redundant local variable.

test/integration/simulacron.js
```diff
@@ -111,200 +163,352 @@ var simulacronHelper = {
...
+ * 
+ * @param {Number} id Idenfitifer of the dc.
+ * @returns {SimulacronDataCenter} The data center, if found.
+ */
+SimulacronCluster.prototype.dc = function(id) {
+  return this.dcs[id];
+};
+
+/**
+ * @param {Number} dataCenterId The data center to return node addresses from, if not provided assumes 0.
```
jorgebay: Should be marked as optional.

test/integration/simulacron.js
```diff
@@ -111,200 +163,352 @@ var simulacronHelper = {
...
-    callback(new Error('invalid node address: ' + nodeAddress));
-  }
+/**
+ * Finds a node in the given data center by its id or address.
+ * 
+ * @param {Number|String} id Identifier of node.  If Number, assumes the id of the node in the data center.
+ * If String, assumes an 'ip:port' designation and looks up node by address.
+ * @returns {SimulacronNode} The node, if found.
+ */
+SimulacronDataCenter.prototype.node = function() {
```
jorgebay: Use the actual parameter field instead of `arguments` array whenever possible (for IDE support, param doc definition, code validation like eslint, ...)

test/integration/simulacron.js
```diff
@@ -111,200 +163,352 @@ var simulacronHelper = {
...
-    path: encodeURI(util.format(stopNodePath, self.id, node.dataCenterId, node.nodeId)),
-    port: self.port,
-    method: 'DELETE'
-  };
-  makeRequest(options, function(data) {
-    callback(data);
-  }).end();
+    // node not found, raise error.
+    throw new Error("No node found for " + args[0] + " in dc " + this.localId);
+  } else {
```
jorgebay: `else` not necessary.

jorgebay:
I'll rebase this, squash some commits, try to maintain some commit authoring info and merge it.

PR #237, opened 2017-09-05
Test re-prepare statement on nodes

landim:
- [x] Create simulatron helper;
- [x] Add test to verify if driver is re-preparing statement when node comes up again;
- [x] Add simulacron startup script on build.yaml (depends on https://github.com/datastax/simulacron/pull/16)

### DIFF
--- a/test/integration/short/client-pool-tests.js
+++ b/test/integration/short/client-pool-tests.js
@@ -559,7 +559,7 @@ describe('Client', function () {
             function wait() {
               if (waiting > 0) {
                 waiting++;
-                if (waiting > 100) {
+                if (waiting > 300) {
                   return timesNext(new Error('Timed out'));
                 }
                 return setTimeout(wait, 50);
--- a/test/integration/short/client-pool-tests.js
+++ b/test/integration/short/client-pool-tests.js
@@ -559,7 +559,7 @@ describe('Client', function () {
             function wait() {
               if (waiting > 0) {
                 waiting++;
-                if (waiting > 300) {
+                if (waiting > 100) {
                   return timesNext(new Error('Timed out'));
                 }
                 return setTimeout(wait, 50);
--- a/test/integration/short/prepare-simulator-tests.js
+++ b/test/integration/short/prepare-simulator-tests.js
@@ -9,12 +9,9 @@ var helper = require('../../test-helper');
 var reconnection = require('../../../lib/policies/reconnection');
 var simulacron = require('../simulacron');
 
-describe('Client #prepare #simulacron', function () {
+describe('Client', function () {
   this.timeout(20000);
   describe('Preparing statements on nodes behavior', function () {
-    var nodes = "5";
-    var clusterName = "simulacronTest";
-    var cassandraVersion = "3.10";
     var sCluster = null;
     var client = null;
     var query = util.format('SELECT * FROM ks.table1 WHERE id1 = ?');
@@ -26,7 +23,7 @@ describe('Client #prepare #simulacron', function () {
         [
           function startCluster(next) {
             sCluster = new simulacron.SimulacronCluster();
-            sCluster.start(nodes, cassandraVersion, "", clusterName, true, 1, next);
+            sCluster.start('5', {}, next);
           },
           function connectCluster(next) {
             var poolingOptions = {};
@@ -103,7 +100,6 @@ describe('Client #prepare #simulacron', function () {
           },
           function verifyIfNodeIsMarkedDown(next) {
             var nodeDown = client.hosts.get(nodeDownAddress);
-            nodeDown.checkIsUp();
             assert(!nodeDown.isUp());
             next();
           },
@@ -142,7 +138,6 @@ describe('Client #prepare #simulacron', function () {
               setTimeout(next, 1000); //give time for driver to re prepare statement
             });
             sCluster.resumeNode(nodeDownAddress, function() {
-              nodeDown.checkIsUp();
             });
           },
           function verifyPrepareQueryOnLastNode(next) {
--- a/test/integration/short/prepare-simulator-tests.js
+++ b/test/integration/short/prepare-simulator-tests.js
@@ -1,0 +1,137 @@
+'use strict';
+var assert = require('assert');
+var util = require('util');
+
+var Client = require('../../../lib/client');
+var types = require('../../../lib/types/index');
+var utils = require('../../../lib/utils');
+var helper = require('../../test-helper');
+var reconnection = require('../../../lib/policies/reconnection');
+var simulacron = require('../simulacron');
+
+describe('Client #prepare', function () {
+  this.timeout(20000);
+  describe('Preparing statements on nodes behavior', function () {
+    var nodes = "5";
+    var clusterName = "simulacronTest";
+    var cassandraVersion = "3.10";
+    var sCluster = null;
+    var client = null;
+    var query = util.format('SELECT * FROM ks.table1 WHERE id1 = ?');
+    before(function (done) {
+      simulacron.start(done);
+    });
+    beforeEach(function (done) {
+      utils.series(
+        [
+          function startCluster(next) {
+            sCluster = new simulacron.SimulacronCluster();
+            sCluster.start(nodes, cassandraVersion, "", clusterName, true, 1, next);
+          },
+          function connectCluster(next) {
+            var poolingOptions = {};
+            poolingOptions[types.distance.local] = 1;
+            client = new Client({
+              contactPoints: [sCluster.getContactPoints()[0]],
+              policies: {
+                reconnection: new reconnection.ConstantReconnectionPolicy(100)
+              },
+              pooling: poolingOptions
+            });
+            client.connect.bind(client);
+            next();
+          },
+          function primeQuery(next) {
+            sCluster.primeQueryWithEmptyResult(query, next);
+          },
+          function clearLog(next) {
+            sCluster.clearLog(next);
+          }
+        ], done);
+    });
+    afterEach(function (done) {
+      client.shutdown.bind(client);
+      sCluster.destroy(done);
+    });
+    after(function (done) {
+      simulacron.stop(done);
+    });
+    it('should prepare query on all hosts', function (done) {
+      var idRandom = types.Uuid.random();
+      client.execute(query, [idRandom], {prepare: 1}, function (err, result) {
+        assert.ifError(err);
+        assert.strictEqual(client.hosts.length, 5);
+        assert.notEqual(result, null);
+        assert.notEqual(result.rows, null);
+        utils.eachSeries(client.hosts.values(), function(host, next) {
+          sCluster.queryNodeLog(host.address, function(logs) {
+            assert.ifError(err);
+            var prepareQuery = logs.find(function(queryLog) {
+              return queryLog.type === "PREPARE"
+                && queryLog.query == query;
+            });
+            assert.notEqual(prepareQuery, undefined);
+            next();
+          });
+        }, done);
+      });
+    });
+    it('should re-prepare query when host go UP again', function (done) {
+      var idRandom = types.Uuid.random();
+      var nodeDownAddress = sCluster.getContactPoints()[4];
+      utils.series(
+        [
+          function stopLastNode(next) {
+            sCluster.stopNode(nodeDownAddress, next);
+          },
+          function prepareQuery(next) {
+            client.execute(query, [idRandom], {prepare: 1}, function (err, result) {
+              assert.ifError(err);
+              assert.strictEqual(client.hosts.length, 5);
+              assert.notEqual(result, null);
+              assert.notEqual(result.rows, null);
+              next();
+            });
+          },
+          function verifyLogs(next) {
+            utils.eachSeries(client.hosts.values(), function(host, nextHost) {
+              sCluster.queryNodeLog(host.address, function(logs) {
+                var prepareQuery = logs.find(function(queryLog) {
+                  return queryLog.type === "PREPARE"
+                    && queryLog.query == query;
+                });
+                if (prepareQuery == undefined) {
+                  assert.strictEqual(nodeDownAddress, host.address);
+                } else {
+                  assert.notEqual(prepareQuery, undefined);
+                }
+                nextHost();
+              });
+            }, next);
+          },
+          function resumeLastNode(next) {
+            var timeout = null;
+            var nodeDown = client.hosts.get(nodeDownAddress);
+            nodeDown.on('up', function() {
+              clearTimeout(timeout);
+              helper.trace("Node marked as UP");
+              setTimeout(next, 1000); //give time for driver to re prepare statement
+            });
+            sCluster.resumeNode(nodeDownAddress, function() {
+              timeout = setTimeout(next, 10000);
+            });
+          },
+          function verifyPrepareQueryOnLastNode(next) {
+            sCluster.queryNodeLog(nodeDownAddress, function(logs) {
+              var prepareQuery = logs.find(function(queryLog) {
+                return queryLog.type === "PREPARE"
+                  && queryLog.query == query;
+              });
+              assert.notEqual(prepareQuery, undefined);
+              next();
+            });
+          }
+        ], done);
+    });
+  });
+});
--- a/test/integration/short/prepare-simulator-tests.js
+++ b/test/integration/short/prepare-simulator-tests.js
@@ -67,11 +67,16 @@ describe('Client #prepare #simulacron', function () {
         utils.eachSeries(client.hosts.values(), function(host, next) {
           sCluster.queryNodeLog(host.address, function(logs) {
             assert.ifError(err);
-            var prepareQuery = logs.find(function(queryLog) {
-              return queryLog.type === "PREPARE"
-                && queryLog.query === query;
-            });
-            assert.notEqual(prepareQuery, undefined);
+            var prepareQuery;
+            for(var i = 0; i < logs.length; i++) {
+              var queryLog = logs[i];
+              if (queryLog.type === "PREPARE" && queryLog.query === query) {
+                prepareQuery = queryLog;
+              }
+            }
+            if (!prepareQuery) {
+              assert.fail('Query no prepared on all hosts');
+            }
             next();
           });
         }, done);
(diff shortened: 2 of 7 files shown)


test/integration/short/client-execute-prepared-tests.js
```diff
@@ -12,8 +12,117 @@ var vit = helper.vit;
...
+              timeout = setTimeout(next, 2000);
+            });
+          },
+          function verifyPrepareQueryOnLastNode(next) {
+            sCluster.queryNodeLog(nodeDownAddress, function(err, log) {
+              assert.ifError(err);
+              var prepareQuery = log.find(function(queryLog) {
+                return queryLog.query.startsWith("PREPARE");
+              });
+              assert.notEqual(prepareQuery, undefined);
```
landim: This tests is failing here. This tests the ability of the driver to prepare a statement in a node recently re-started. Maybe the test is not reflecting this scenario properly.

test/integration/short/client-execute-prepared-tests.js
```diff
@@ -12,8 +12,117 @@ var vit = helper.vit;
...
+    var client = null;
+    var query = util.format('SELECT * FROM ks.table1 WHERE id1 = ?');
+    beforeEach(function (done) {
+      utils.series(
+        [
+          function stopAnyCcmCluster(next) {
+            helper.ccmHelper.removeIfAny(next);
+          },
+          function startCluster(next) {
+            sCluster = new SimulacronCluster("5", "3.10", "", "test", true, 1, function (err) {
```
jorgebay: We should avoid callbacks in constructors, as you can't get the context of the current instance.

```js
var myVar = new MyConstructor(function() {
  // I can't access myVar here (it depends on implementation of MyConstructor)
});
```
vs
```js
var myVar = new MyConstructor();
myVar.method(function () {
  // I can access myVar
});
```

test/integration/short/index.js
```diff
@@ -0,0 +1,102 @@
+'use strict';
+var spawn = require('child_process').spawn;
+
+var simulacronHelper = {
```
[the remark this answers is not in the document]
jorgebay: I see now... it's required.
We should find another way (maybe starting and stopping simulacron on the test suite itself), as this way it's not possible to execute a single suite, without having to run index.js.

test/integration/short/index.js
```diff
@@ -0,0 +1,102 @@
...
+        }
+      }
+    });
+
+    return p;
+  },
+  start: function(cb) {
+    console.log('starts simulacron server!');
+    var self = this;
+    var simulacronJarPath = process.env['SIMULACRON_PATH'];
```
jorgebay: Throw error if not defined.

test/integration/short/client-execute-prepared-tests.js
```diff
@@ -12,8 +12,117 @@ var vit = helper.vit;
...
+              var clientOptions = utils.extend({}, helper.baseOptions);
+              clientOptions.contactPoints = [];
+              clientOptions.contactPoints.push(sCluster.dcs[0].nodes[0].address);
```
jorgebay: You can expose a property or method on `SimulacronCluster` named `contactPoints` and use it to construct the client options:

```js
var client = Client({ contactPoints: sCluster.contactPoints() });
```

test/integration/short/client-execute-prepared-tests.js
```diff
@@ -12,8 +12,117 @@ var vit = helper.vit;
...
+              client = new Client(clientOptions);
```
jorgebay: Use a different `client` per class if you are stopping nodes.
Use a `ConstantReconnectionPolicy` to be able to predict when the driver should already have that node marked as UP.

test/integration/short/client-execute-prepared-tests.js
jorgebay: Check whether the node is being marked as back up.

test/integration/short/client-execute-prepared-tests.js
```diff
@@ -12,8 +12,117 @@ var vit = helper.vit;
 var vdescribe = helper.vdescribe;
 var Uuid = types.Uuid;
 
+var SimulacronCluster = require('../simulacron-helper');
+
 describe('Client', function () {
   this.timeout(120000);
+  describe('Preparing statements on nodes behavior', function () {
```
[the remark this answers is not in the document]
landim: What about we create another test category (short, long and mock)?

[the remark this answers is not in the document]
jorgebay: hmmm, currently the tests are organized by the amount of time it takes to run them, I don't see why it shouldn't be in `short`.
If we want to exclude it in the future, for a certain platform, we can use a custom `describe`.

test/integration/simulacron-helper.js
```diff
@@ -0,0 +1,199 @@
+var Client = require('node-rest-client').Client;
```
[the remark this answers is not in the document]
landim: I have chosen an high level library, because the built-in module is fairly low-level, which would need to append the data (received in blocks on 'data' event) response and listing to 'end' event to parse the response in all requests.
`
const https = require('https');
 
https.get('url', (resp) => {
  let data = '';
 
  // appending data
  resp.on('data', (chunk) => {
    data += chunk;
  });
 
  // Wait to finish to parse JSON
  resp.on('end', () => {
    JSON.parse(data);
  });
 
}).on("error", (err) => {
  console.log("Error: " + err.message);
});
`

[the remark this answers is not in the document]
jorgebay: I would prefer to maintain a utility function against a core module than a dependency to a external module.

test/integration/short/client-execute-prepared-tests.js
```diff
@@ -12,6 +12,8 @@ var vit = helper.vit;
 var vdescribe = helper.vdescribe;
 var Uuid = types.Uuid;
 
+var SimulacronCluster = require('../simulacron-helper');
```
tolbertam: Looks like this is unused here

test/integration/simulacron-helper.js
```diff
@@ -0,0 +1,298 @@
+var helper = require('../test-helper');
```
tolbertam: Small nit, i think it would be nice to simply call it `simulacron` instead of `simulacron-helper` for brevity.

test/integration/short/prepare-simulator-tests.js
```diff
@@ -0,0 +1,142 @@
...
+    beforeEach(function (done) {
+      utils.series(
+        [
+          function startCluster(next) {
+            sCluster = new simulacronHelper.SimulacronCluster();
+            sCluster.start(nodes, cassandraVersion, "", clusterName, true, 1, next);
+          },
+          function connectCluster(next) {
+            client = new Client({
+              contactPoints: [sCluster.dcs[0].nodes[0].address],
```
tolbertam: A convenience function for getting all the addresses would be nice.   i.e. `getContactPoints` which simply returns the addresses of all nodes in dc0.

test/integration/short/prepare-simulator-tests.js
```diff
@@ -0,0 +1,142 @@
...
+      client.execute(query, [idRandom], {prepare: 1}, function (err, result) {
+        assert.ifError(err);
+        assert.strictEqual(client.hosts.length, 5);
+        assert.notEqual(result, null);
+        assert.notEqual(result.rows, null);
+        utils.eachSeries(client.hosts.values(), function(host, next) {
+          sCluster.queryNodeLog(host.address, function(err, log) {
+            assert.ifError(err);
+            var prepareQuery = log.find(function(queryLog) {
+              return queryLog.type.startsWith("PREPARE");
```
tolbertam: can be equals instead of startsWith right?   Would also be good to validate the query text as well.

test/integration/short/prepare-simulator-tests.js
```diff
@@ -0,0 +1,142 @@
...
+            });
+            sCluster.resumeNode(nodeDownAddress, function() {
+              timeout = setTimeout(next, 2000);
+            });
+          },
+          function verifyPrepareQueryOnLastNode(next) {
+            sCluster.queryNodeLog(nodeDownAddress, function(err, logs) {
+              assert.ifError(err);
+              var prepareQuery = logs.find(function(queryLog) {
+                return queryLog.query.startsWith("PREPARE");
```
tolbertam: should be `type` and not `query` (due to recent changes in simulacron)

test/integration/short/prepare-simulator-tests.js
```diff
@@ -0,0 +1,142 @@
...
+        }, done);
+      });
+    });
+    it('should re-prepare query when host go UP again', function (done) {
+      var idRandom = types.Uuid.random();
+      var nodeDownAddress = sCluster.dcs[0].nodes[4].address;
+      utils.series(
+        [
+          function executeNotPreparedQuery(next) {
+            client.execute(query, [idRandom], {}, function (err, result) {
```
tolbertam: Since already connected is this query still needed?

test/integration/short/prepare-simulator-tests.js
```diff
@@ -0,0 +1,142 @@
+'use strict';
```
tolbertam: This is pretty awesome, runs nice and fast!  Good work.

test/integration/simulacron-helper.js
```diff
@@ -0,0 +1,298 @@
...
+          cb(Error("Process exited with non-zero exit code: " + code));
+        }
+      }
+    });
+
+    return p;
+  },
+  start: function(cb) {
+    var self = this;
+    var simulacronJarPath = process.env['SIMULACRON_PATH'];
```
tolbertam: How about this?

- Check `SIMULACRON_PATH`, if not set, set to `$HOME/simulacron.jar`
- Check for existence of jar, if it doesn't exist, error out.

This way, if the user has their environment set up with the jar in the home directory already, there is no need to set the environment variable.  If the jar doesn't exist, an error can be emitted saying `Could not find <jarfile>, place jar in this path or set SIMULACRON_PATH appropriately`

test/integration/simulacron-helper.js
```diff
@@ -0,0 +1,298 @@
...
+    } else {
+      cb(Error("Process is not defined."));
+    }
+  },
+  baseAddress: "http://localhost:8187",
+  SimulacronCluster: SimulacronCluster
+};
+
+function SimulacronCluster() {
+  this.baseAddress = simulacronHelper.baseAddress;
```
tolbertam: We may get to point where we want to configure ip address ranges and http interface address, but I think it's fine to make it not configurable as you have for now.

test/integration/short/client-execute-prepared-tests.js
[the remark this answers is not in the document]
tolbertam: One benefit I can think of for putting it in it's only directory (like `test/simulated`) is that we could run these tests in isolation, but I think if we decide we want that we can always do that later.

tolbertam:
Just noticed that there are a few errors emitted by travis:

```
/home/travis/build/datastax/nodejs-driver/test/integration/short/prepare-simulator-tests.js
   71:35  error  Expected '===' and instead saw '=='  eqeqeq
  101:39  error  Expected '===' and instead saw '=='  eqeqeq
  103:34  error  Expected '===' and instead saw '=='  eqeqeq
  128:37  error  Expected '===' and instead saw '=='  eqeqeq
/home/travis/build/datastax/nodejs-driver/test/integration/simulacron.js
  2:1  error  Parsing error: The keyword 'const' is reserved
✖ 5 problems (5 errors, 0 warnings)
```

test/integration/short/prepare-simulator-tests.js
```diff
@@ -0,0 +1,137 @@
...
+          function resumeLastNode(next) {
+            var timeout = null;
+            var nodeDown = client.hosts.get(nodeDownAddress);
+            nodeDown.on('up', function() {
+              clearTimeout(timeout);
+              helper.trace("Node marked as UP");
+              setTimeout(next, 1000); //give time for driver to re prepare statement
+            });
+            sCluster.resumeNode(nodeDownAddress, function() {
+              timeout = setTimeout(next, 10000);
```
tolbertam: does it need to be 10 seconds or could it be shorter?

test/integration/simulacron.js
```diff
@@ -0,0 +1,290 @@
...
+  var options = {
+    host: self.baseAddress,
+    path: encodeURI(util.format(stopNodePath, self.id, node.dataCenterId, node.nodeId)),
+    port: self.port,
+    method: 'DELETE'
+  };
+  makeRequest(options, function(data) {
+    callback(data);
+  }).end();
+}
```
jorgebay: There are other linting issues: 
- missing semicolon
- no-else-return
- Use of console

You can run eslint locally: `npm run eslint`

test/integration/short/prepare-simulator-tests.js
```diff
@@ -0,0 +1,137 @@
...
+            });
+          },
+          function verifyLogs(next) {
+            utils.eachSeries(client.hosts.values(), function(host, nextHost) {
+              sCluster.queryNodeLog(host.address, function(logs) {
+                var prepareQuery = logs.find(function(queryLog) {
+                  return queryLog.type === "PREPARE"
+                    && queryLog.query == query;
+                });
+                if (prepareQuery == undefined) {
```
jorgebay: You can use `(!prepareQuery)` (`undefined`, `null` or `''`)

test/integration/short/prepare-simulator-tests.js
```diff
@@ -0,0 +1,137 @@
...
+                nextHost();
+              });
+            }, next);
+          },
+          function resumeLastNode(next) {
+            var timeout = null;
+            var nodeDown = client.hosts.get(nodeDownAddress);
+            nodeDown.on('up', function() {
+              clearTimeout(timeout);
+              helper.trace("Node marked as UP");
```
jorgebay: On my local machine, the node is not marked as down and back up but the test succeeds.

test/integration/short/prepare-simulator-tests.js
```diff
@@ -0,0 +1,137 @@
...
+                } else {
+                  assert.notEqual(prepareQuery, undefined);
+                }
+                nextHost();
+              });
+            }, next);
+          },
+          function resumeLastNode(next) {
+            var timeout = null;
+            var nodeDown = client.hosts.get(nodeDownAddress);
```
jorgebay: Verify that there is a node considered as down.

test/integration/short/client-pool-tests.js
```diff
@@ -559,7 +559,7 @@ describe('Client', function () {
             function wait() {
               if (waiting > 0) {
                 waiting++;
-                if (waiting > 100) {
+                if (waiting > 300) {
```
jorgebay: This test was fixed on https://github.com/datastax/nodejs-driver/pull/232/commits/34532c717aecd9f3d1098cd5771329e61b419c68
You can rollback this change.

test/integration/short/prepare-simulator-tests.js
```diff
@@ -0,0 +1,166 @@
...
+var util = require('util');
+
+var Client = require('../../../lib/client');
+var types = require('../../../lib/types/index');
+var utils = require('../../../lib/utils');
+var helper = require('../../test-helper');
+var reconnection = require('../../../lib/policies/reconnection');
+var simulacron = require('../simulacron');
+
+describe('Client #prepare #simulacron', function () {
```
jorgebay: The `#name()` notation is used to represent an instance method. I think using like `'Client'` would be fine.

test/integration/short/prepare-simulator-tests.js
```diff
@@ -0,0 +1,166 @@
...
+    var query = util.format('SELECT * FROM ks.table1 WHERE id1 = ?');
+    before(function (done) {
+      simulacron.start(done);
+    });
+    beforeEach(function (done) {
+      utils.series(
+        [
+          function startCluster(next) {
+            sCluster = new simulacron.SimulacronCluster();
+            sCluster.start(nodes, cassandraVersion, "", clusterName, true, 1, next);
```
jorgebay: Except the amount of nodes and the callback, the rest of the parameters can be optional with a default.
You could use an options parameter or something like that:

```js
sCluster.start(nodes, { optionalParams: value... }, next);
```

About the cassandra version, you can use `helper.getCassandraVersion()` internally.

test/integration/short/prepare-simulator-tests.js
```diff
@@ -0,0 +1,166 @@
...
+            }, next);
+          },
+          function resumeLastNode(next) {
+            var nodeDown = client.hosts.get(nodeDownAddress);
+            nodeDown.on('up', function() {
+              helper.trace("Node marked as UP");
+              setTimeout(next, 1000); //give time for driver to re prepare statement
+            });
+            sCluster.resumeNode(nodeDownAddress, function() {
+              nodeDown.checkIsUp();
```
jorgebay: We should avoid using internal methods in the integration tests (whenever possible).

test/integration/short/prepare-simulator-tests.js
```diff
@@ -0,0 +1,166 @@
...
+                assert.strictEqual(client.hosts.length, 5);
+                assert.notEqual(result, null);
+                assert.notEqual(result.rows, null);
+                nextIteration();
+              });
+            }, next);
+          },
+          function verifyIfNodeIsMarkedDown(next) {
+            var nodeDown = client.hosts.get(nodeDownAddress);
+            nodeDown.checkIsUp();
```
jorgebay: Internal method call.

test/integration/simulacron.js
```diff
@@ -0,0 +1,294 @@
...
+        helper.trace('Simulacron initialized!');
+        initialized = true;
+        cb();
+      }
+    });
+  },
+  stop: function(cb) {
+    if(this.sProcess !== undefined) {
+      if(this.sProcess.exitCode) {
+        helper.trace("Server already stopped with exit code %d.", this.process.exitCode);
```
jorgebay: `this.process` is not defined, probably meant `this.sProcess`.

test/integration/simulacron.js
```diff
@@ -0,0 +1,294 @@
...
+    }
+    if (!fs.existsSync(simulacronJarPath)) {
+      throw new Error("Simulacron jar not found at: " + simulacronJarPath);
+    }
+
+    var processName = 'java';
+    var params = ['-jar', simulacronJarPath, '--ip', '127.0.0.101'];
+    var initialized = false;
+
+    var timeout = setTimeout(function() {
```
jorgebay: I think this `timeout` is not necessary, as you already handled it in `_execute()` method.

landim: They are different timeouts. The one inside _execute just check if the process timeouts to start. The one inside start verify if the process start correctly (through logs).

landim:
Tests failing "Client #execute().should maintain the domain in the callbacks" fixed at https://github.com/datastax/nodejs-driver/commit/34532c717aecd9f3d1098cd5771329e61b419c68

jorgebay:
@tolbertam Do you want to do a final look into it before we merge it?

test/integration/simulacron.js
```diff
@@ -0,0 +1,310 @@
...
+      }
+    });
+
+    return p;
+  },
+  start: function(cb) {
+    var self = this;
+    var simulacronJarPath = process.env['SIMULACRON_PATH'];
+    if (!simulacronJarPath) {
+      simulacronJarPath = '$HOME/simulacron.jar';
```
tolbertam: This doesn't work it seems.  Need to resolve the `$HOME` environment variable from process.env otherwise it interprets `$HOME` literally

tolbertam: went ahead and pushed a commit for that.

jorgebay:
I'll merge it once Jenkins finishes.

PR #238, opened 2017-09-05
NODEJS-375 Allow async creation of Uuid

guzmo:
I was supposed to create a small benchmark of async vs sync but I've had no time :( I ran some small tests with https://www.npmjs.com/package/loadtest, but only on my computer, which are running a lot of other processes at the same time, so hard to tell. 
Hope the code is not too messy ;)

### DIFF
--- a/lib/types/uuid.js
+++ b/lib/types/uuid.js
@@ -34,17 +34,18 @@ Uuid.fromString = function (value) {
  * Creates a new random (version 4) Uuid.
  * @returns {Uuid}
  */
-Uuid.random = function () {
-  var buffer = getRandomBytes();
-  //clear the version
-  buffer[6] &= 0x0f;
-  //set the version 4
-  buffer[6] |= 0x40;
-  //clear the variant
-  buffer[8] &= 0x3f;
-  //set the IETF variant
-  buffer[8] |= 0x80;
-  return new Uuid(buffer);
+Uuid.random = function (cb) {
+  if (cb) {
+    getRandomBytes(function(err, buffer) {
+      if (err) {
+        return cb(err);
+      }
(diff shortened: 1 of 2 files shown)


guzmo:
The Synchronous function might be faster for creating a small amount of bytes ( which Uuid is doing I think ), but it might be safer to use the async because you might block the node process.
https://github.com/nodejs/help/issues/457

Well, your choice if you want to have it.

jorgebay:
As [discussed in the ticket](https://datastax-oss.atlassian.net/browse/NODEJS-375), I think is a good idea to have an asynchronous overload of `Uuid.random()`.

About `TimeUuid`, do you want to add support for callbacks for `TimeUuid.now()` and `TimeUuid.fromDate()` in this patch? Otherwise, we can create a separate ticket for it.

guzmo:
@jorgebay Can we create separate tickets for TimeUuid? :) One step at a time is usually a good thing and would be nice to at least get this fix in an upcoming release.
Does it look good otherwise? Some tests timed out, maybe I should look at them?

Cheers

jorgebay:
ok, I've created a separate ticket for `TimeUuid`: [NODEJS-387](https://datastax-oss.atlassian.net/browse/NODEJS-387).

I'll talk with the rest of the team to see if we can include it for the upcoming v3.3.0 today and let you know.
The patch is looking good, I'll do a full review soon.

jorgebay:
We will try to include it for v3.3.0 release.

guzmo:
Good catches! Will fix the review comments tonight.

jorgebay:
Landed in cb89374.
Thanks @guzmo !

guzmo:
No problem, glad to help :)

PR #240, opened 2017-09-06
Expose Client#getState()

jorgebay:


lib/metadata/client-state.js
```diff
@@ -0,0 +1,96 @@
...
+ * @return {Array<Host>}
+ */
+ClientState.prototype.getConnectedHosts = function () {
+  return this._hosts;
+};
+
+/**
+ * Gets the amount of open connections to a given host.
+ * @param {Host} host
+ * @return {Number|undefined}
```
tolbertam: One question here.  Say you have a previous reference to a host.  The host goes down, the user requests new state and then calls this method.  In this case `undefined` is returned.  Wondering if it would be better to return `0` in this case (and also for `getInFlightQueries`) or if you think `undefined` is better since it communicates to user that the host is not connected, thus there is no state for it.

test/integration/short/metadata-tests.js
```diff
@@ -9,1090 +9,1132 @@ var errors = require('../../../lib/errors');
 var vit = helper.vit;
 var vdescribe = helper.vdescribe;
 
-describe('Metadata', function () {
+describe('metadata', function () {
```
tolbertam: Are the changes in these tests mostly just for formatting?  Couldn't detect anything otherwise but thought i'd ask.

jorgebay: Its lowercased, now the structure is the following:
- metadata
-- Metadata
-- Client#getState()

lib/metadata/client-state.js
jorgebay: I was using `undefined` as in "We don't have that info", but now that I think about it, for reporting (like logging / tracing info) it would be simpler for the user if we always return a number. I will change it.

tolbertam: 👍 that was my thinking as well.  I had some code that was iterating over all hosts (not just connected ones) and encountered that, now I know longer need to special case if returns undefined so the change should work great.

PR #241, opened 2017-09-13
Test of consistency set at QueryRequest [NODEJS-370]

landim:


test/integration/short/connection-simulator-tests.js
```diff
@@ -0,0 +1,86 @@
+'use strict';
+var assert = require('assert');
+var utils = require('../../../lib/utils');
+var types = require('../../../lib/types/index');
+var helper = require('../../test-helper');
+var simulacron = require('../simulacron');
+var defaultOptions = require('../../../lib/client-options.js').defaultOptions();
+var requests = require('../../../lib/requests.js');
+
+describe('Connection', function () {
```
tolbertam: The test seems good but it doesn't validate the change for NODEJS-370 since the change was made on the `getTrace()` method right?  

It would require additional logic in simulacron to test that, by adding support for parsing the tracing flag and adding an arbitrary uuid to the response.  I think that is doable if you want to try adding that.  

An alternative way to test would be to do something like take a node down, and then call `getTrace` with CL ALL, it should fail, and then try again with CL ANY and it should succeed (or something like that).  It would probably be good to make that a long test since it would require bringing down nodes.

landim: Good. I will include that test.

landim: Having a second thoughts about this, I will include a test that check if the queries executed by getTrace set CL by getting the simulacron logged queries.

jorgebay:
#232 was merged, I'll rebase this one.

test/test-helper.js
```diff
@@ -729,6 +729,34 @@ var helper = {
...
+  },
+  newConnection: function(address, protocolVersion, options){
+    if (!address) {
+      address = helper.baseOptions.contactPoints[0];
+    }
+    if (typeof protocolVersion === 'undefined') {
+      protocolVersion = this.getProtocolVersion();
+    }
+    //var logEmitter = function (name, type) { if (type === 'verbose') { return; } console.log.apply(console, arguments);};
+    options = utils.deepExtend({logEmitter: helper.noop}, options || this.defaultOptions);
```
jorgebay: `this.defaultOptions` is not defined.

jorgebay:
Any updates about this?

PR #242, opened 2017-09-14
Simulacron: fix stop callback on windows

jorgebay:


jorgebay:
Also, removing the infrastructure "meta" tests, as those were taking more than 2 minutes on appveyor...

PR #243, opened 2017-09-15
Declared methods as non enumerable

noah-ellman:
Declared the recently added methods to Row.prototype with Object.defineProperty setting enumerable to false. Previously, resultset rows could be used like a plain Object, including with for in/of loops. The last changes broke that behavior causing backwards incompatability. This fix will remedy that without effecting anything else.  Also as I side note, it would appear that if a database has any columns named "values" or "keys",  it would break this module.  I suggest that row objects should either behavior like plain objects as they did previously, or the row data should be moved into another variable instead of being directly on the instance.

### DIFF
--- a/lib/types/row.js
+++ b/lib/types/row.js
@@ -17,49 +17,62 @@ function Row(columns) {
  * Returns the cell value.
  * @param {String|Number} columnName Name or index of the column
  */
-Row.prototype.get = function (columnName) {
-  if (typeof columnName === 'number') {
-    //its an index
-    return this[this.__columns[columnName].name];
-  }
-  return this[columnName];
-};
+Object.defineProperty(Row.prototype, "get", {
+    enumerable: false,
+    value: function (columnName) {
+        if (typeof columnName === 'number') {
+            //its an index
+            return this[this.__columns[columnName].name];
+        }
+        return this[columnName];
+    }
+});
 
 /**
  * Returns an array of the values of the row
  * @returns {Array}
  */
-Row.prototype.values = function () {
-  var valuesArray = [];
-  this.forEach(function (val) {
-    valuesArray.push(val);
-  });
-  return valuesArray;
-};
+Object.defineProperty(Row.prototype, "values", {
+        enumerable: false,
+        value: function () {
+            var valuesArray = [];
+            this.forEach(function (val) {
+                valuesArray.push(val);
+            });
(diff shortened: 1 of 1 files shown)


noah-ellman:
The appveyor tests passed the first time.  Nothing changed except eslint whitespace fixes.   Maybe run it again.

jorgebay:
Hi @noah-ellman, thanks for the proposed contribution!

I think the current behaviour for enumeration of row properties is correct:

```js
// built-in methods are not included
const columnNames = Object.keys(row);
```

```js
const columnNames = [];
for (var key in row) {
  if (row.hasOwnProperty(key)) {
    columnNames.push(key);
  } 
}
```

See:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#Iterating_over_own_properties_only
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys

These statements will be valid even if you name your columns "values" or "forEach()".
About the utility methods inside row, those are just utilities, the user can still use access column names and cell values without those (when overridden).

```js
const values = columnNames.map(name => row[name]);
```

I've find using `row['col1']` or row.col1 notation very comfortable for the user. I've tried more elegant solutions like defining those properties as getter methods, but the performance on current Node.js LTS versions was awful.

We look forward for your next contribution (code or proposal), you can check out the filed tickets on JIRA (we don't use GitHub issues): https://datastax-oss.atlassian.net/projects/NODEJS/issues/

noah-ellman:
Thanks for reviewing my submission.  However I tend to disagree with you.  The current implementation is unsafe for plain vanilla for-in and for-of loops unless the user manually checks hasOwnProperty himself, which many people do not.  Having to call keys() every time is an uneeded overhead that could be avoided by making those methods non-enumerable.  Thanks!

noah-ellman:
In response to you:
Having a column name of "keys" will absolutely break this module. 
The following code reproduces easily: 

```
cassandra.execute("select host_id as keys from system.local")
.then(
    results => {
        console.log(results.rows[0].values());  // works
        console.log(results.rows[0].keys());  // ERROR
    }
)

TypeError: results.rows[0].keys is not a function


```

PR #244, opened 2017-09-18
Speculative executions doc

jorgebay:
Include Speculative executions documentation.
Minor jsdoc fixes.

### DIFF
--- a/doc/features/speculative-executions/README.md
+++ b/doc/features/speculative-executions/README.md
@@ -99,7 +99,7 @@ const client = new Client({
 ## Enabling speculative execution 
 
 Speculative executions are controlled by an instance of `SpeculativeExecutionPolicy` provided when initializing the
-Client. This policy defines the threshold after which a new speculative execution is triggered.
+`Client`. This policy defines the threshold after which a new speculative execution is triggered.
 
 The driver provides a `ConstantSpeculativeExecutionPolicy` that schedules a given number of speculative executions,
 separated by a fixed delay, the policy is exported under the `<root>.policies.speculativeExecution` module.
--- a/doc/features/speculative-executions/README.md
+++ b/doc/features/speculative-executions/README.md
@@ -1,0 +1,214 @@
+# Speculative query execution
+
+Sometimes a server node might be experiencing difficulties (for example, long GC pause) and take longer than
+usual to reply. Queries sent to that node experience higher latencies than expected.
+
+One thing we can do to improve that is preemptively start a second execution of the query against another node, 
+before the first node has replied or errored out. If that second node replies faster, we can send the response back 
+to the client (we also cancel the first query):
+
+```
+client           driver          exec1  exec2
+--+----------------+--------------+------+---
+  | execute(query) |
+  |--------------->|
+  |                | query host1
+  |                |------------->|
+  |                |              |
+  |                |              |
+  |                |     query host2
+  |                |-------------------->|
+  |                |              |      |
+  |                |              |      |
+  |                |     host2 replies   |
+  |                |<--------------------|
+  |   complete     |              |
+  |<---------------|              |
+  |                | cancel       |
+  |                |------------->|
+```
+
+Or the first node could reply just after the second execution was started. In this case, we cancel the second execution.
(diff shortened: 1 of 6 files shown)


doc/features/speculative-executions/README.md
```diff
@@ -0,0 +1,214 @@
+# Speculative query execution
```
tolbertam: One thing was just thinking about while reading this is that these concepts are not specific to any driver per se.  I.E. we have mostly duplicated documentation between the C#, java and nodejs drivers with regards to speculative execution and query idempotence.   Wonder if going forward we could document these concepts at a higher level so we don't have to repeat ourselves amongst the docs, but not really sure if that is worth it.

doc/features/speculative-executions/README.md
```diff
@@ -0,0 +1,214 @@
...
+  queryOptions: { 
+    isIdempotent: true
+  }
+});
+```
+
+## Enabling speculative execution 
+
+Speculative executions are controlled by an instance of `SpeculativeExecutionPolicy` provided when initializing the
+Client. This policy defines the threshold after which a new speculative execution is triggered.
```
tolbertam: Small nit, Client could be `Client`.

doc/features/speculative-executions/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+This simple policy uses a constant threshold:
+
+```js
+const client = new Client({
+  contactPoints,
+  policies: {
+    speculativeExecution: new ConstantSpeculativeExecutionPolicy(
+      200, // delay before a new execution is launched
+      2) // maximum amount of additional executions
```
tolbertam: Shouldn't this be `maximum amount of executions`?  Additional to me implies "after the first".

doc/features/speculative-executions/README.md
```diff
@@ -0,0 +1,214 @@
...
+  }
+});
+```
+
+Given the configuration above, an idempotent query would be handled this way:
+
+- start the initial execution at t0
+- if no response has been received at t0 + 200 milliseconds, start a speculative execution on another node
+- if no response has been received at t0 + 400 milliseconds, start another speculative execution on a third node
```
tolbertam: Similar to comment above, if the configuration is `2`, I would only expect there to be one execution beyond the initial one.

doc/features/speculative-executions/README.md
```diff
@@ -0,0 +1,214 @@
...
+this is the case. Now you run the following query, which completes successfully:
+
+```
+delete from my_table where k = 1;
+```
+
+But now the second execution of the first query finally reaches its target node, which applies the mutation. The row
+that you've just deleted is back!
+
+**Using [query timestamps](../query-timestamps)**, which are enabled by default, prevents
```
tolbertam: That reminds me that it would be good to validate that if using client timestamps, that the same timestamp is used among all executions.  Can't imagine that not being the case, but will double check.

doc/features/speculative-executions/README.md
tolbertam: I just noticed that the java documentation is wrong from this perspective as well.  Will fix it there.

jorgebay: The second parameter of `ConstantSpeculativeExecutionPolicy` is the amount of _speculative executions_, I think the legend is correct.

The `SpeculativeExecutionPolicy` is first used to get a plan after the first execution.

tolbertam: Ack I'm wrong about that.  I get confused by that constantly.  In that case, i like `additional` as it is more explicit to mean beyond the first.

tolbertam: It would be good to use that in the `ConstantSpeculativeExecutionPolicy` jsdoc too to make it more clear that it means beyond the first.

jorgebay: will do

doc/features/speculative-executions/README.md
jorgebay: Its true that is 98% alike and its weird to maintain "copies", but the code samples and details (like further support of certain config settings) I think justifies for different doc sources.

PR #246, opened 2017-09-18
Add documentation for tuning retry policy

idanto:
- Add code example for:
   - How to override retry policy
   - How to use `IdempotenceAwareRetryPolicy`
   - How to indicate if query is `isIdempotent` in both query and client levels

### DIFF
--- a/doc/features/tuning-policies/README.md
+++ b/doc/features/tuning-policies/README.md
@@ -157,15 +157,17 @@ unexpected error, invoked in the following situations:
     - When the contacted host replies with an error, such as `overloaded`, `isBootstrapping`, `serverError`, etc. In 
     this case, the error is instance of `ResponseError`
 
-A default and base retry policy is included, along with `IdempotenceAwareRetryPolicy` that considers query idempotence. The default implementation doesn't use `IdempotenceAwareRetryPolicy`, in order to use it you should wrap the default policy with `IdempotenceAwareRetryPolicy`:
+A default and base retry policy is included, along with IdempotenceAwareRetryPolicy. The default implementation doesn't consider query idempotence. You can use the IdempotenceAwareRetryPolicy by setting it on the client options:
 
 ``` js
- client = new cassandra.Client({
-     contactPoints: contactPoints.split(','),
-     keyspace: keyspace,
-     authProvider: authProvider,
+var Client = reuiqre('cassandra-driver').Client;
+var IdempotenceAwareRetryPolicy = reuiqre('cassandra-driver').policies.retry.IdempotenceAwareRetryPolicy;
+var RetryPolicy = require('cassandra-driver').policies.retry.RetryPolicy;
+
+ client = new Client({
+     contactPoints,
      policies: {
-         retry: new cassandra.policies.retry.IdempotenceAwareRetryPolicy(new cassandra.policies.retry.RetryPolicy())
+         retry: new IdempotenceAwareRetryPolicy(new RetryPolicy())
      }
  });
 ```
--- a/doc/features/tuning-policies/README.md
+++ b/doc/features/tuning-policies/README.md
@@ -157,16 +157,173 @@ unexpected error, invoked in the following situations:
     - When the contacted host replies with an error, such as `overloaded`, `isBootstrapping`, `serverError`, etc. In 
     this case, the error is instance of `ResponseError`
 
-The [operation info][OperationInfo], passed as a parameter to the retry policy methods, exposes the `query` and query 
-`options` as properties.
+A default and base retry policy is included, along with `IdempotenceAwareRetryPolicy` that considers query idempotence. The default implementation doesn't use `IdempotenceAwareRetryPolicy`, in order to use it you should wrap the default policy with `IdempotenceAwareRetryPolicy`:
 
-A default and base retry policy is included, along with `IdempotenceAwareRetryPolicy` that considers query idempotence.
+``` js
+ client = new cassandra.Client({
+     contactPoints: contactPoints.split(','),
+     keyspace: keyspace,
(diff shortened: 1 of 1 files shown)


doc/features/tuning-policies/README.md
```diff
@@ -157,16 +157,175 @@ unexpected error, invoked in the following situations:
     - When the contacted host replies with an error, such as `overloaded`, `isBootstrapping`, `serverError`, etc. In 
     this case, the error is instance of `ResponseError`
 
-The [operation info][OperationInfo], passed as a parameter to the retry policy methods, exposes the `query` and query 
-`options` as properties.
+A default and base retry policy is included, along with `IdempotenceAwareRetryPolicy` that considers query idempotence. The default implementation doesn't use `IdempotenceAwareRetryPolicy`, in order to use it you should wrap the default policy with `IdempotenceAwareRetryPolicy`:
```
jorgebay: To avoid repeating `IdempotenceAwareRetryPolicy`, I would use:

> A default and base retry policy is included, along with `IdempotenceAwareRetryPolicy`. The default implementation doesn't consider query idempotence. You can use the `IdempotenceAwareRetryPolicy` by setting it on the client options:

doc/features/tuning-policies/README.md
```diff
@@ -157,16 +157,175 @@ unexpected error, invoked in the following situations:
...
+
+``` js
+ client = new cassandra.Client({
```
jorgebay: It would be nice to avoid using the `cassandra` qualifier / module name whenever possible, that way it can be safer for the user to copy and paste (ie: users can import `Client` directly).

doc/features/tuning-policies/README.md
```diff
@@ -157,16 +157,175 @@ unexpected error, invoked in the following situations:
...
+     contactPoints: contactPoints.split(','),
+     keyspace: keyspace,
```
jorgebay: Avoid declaring properties that are optional not directly related to the code same.

doc/features/tuning-policies/README.md
```diff
@@ -157,16 +157,175 @@ unexpected error, invoked in the following situations:
     - When the contacted host replies with an error, such as `overloaded`, `isBootstrapping`, `serverError`, etc. In 
     this case, the error is instance of `ResponseError`
 
-The [operation info][OperationInfo], passed as a parameter to the retry policy methods, exposes the `query` and query 
-`options` as properties.
+A default and base retry policy is included, along with `IdempotenceAwareRetryPolicy` that considers query idempotence. The default implementation doesn't use `IdempotenceAwareRetryPolicy`, in order to use it you should wrap the default policy with `IdempotenceAwareRetryPolicy`:
+
+``` js
+ client = new cassandra.Client({
+     contactPoints: contactPoints.split(','),
```
jorgebay: As `contactPoints` is a mandatory option, I would include it without any transformation:

```javascript
const client = new Client({
  contactPoints,
  // the property I want to show
});
```
In this case, you could use enhanced object literals https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Grammar_and_types#Enhanced_Object_literals

doc/features/tuning-policies/README.md
```diff
@@ -157,16 +157,175 @@ unexpected error, invoked in the following situations:
...
+     policies: {
+         retry: new cassandra.policies.retry.IdempotenceAwareRetryPolicy(new cassandra.policies.retry.RetryPolicy())
+     }
+ });
+```
+
+You can also override the policy implementation by inherits the default policy:
+
+``` js
+var cassandra = require('cassandra-driver');
```
jorgebay: Creating a custom retry policy is something that is not very common, I don't think its a good idea to include a complete code sample about it.

idanto: The whole point of the documentation was to give more info on the retry policy, cause when I tried to use it the documentation is very missing, also the logic of how the default policy works.
Maybe instead of put here a complete code sample we can put a link to the default policy or add some documentation on how the default one works/ how to override it without the implementation.

What do you think?

doc/features/tuning-policies/README.md
idanto: OK, thanks. 
I still want to show where those objects are in the package so I included a `require` statements. I can say that when I looked how to do those things it took me some time to find all the policies.

doc/features/tuning-policies/README.md
idanto: Thanks

doc/features/tuning-policies/README.md
idanto: Thanks

doc/features/tuning-policies/README.md
idanto: Thanks

doc/features/tuning-policies/README.md
jorgebay: Overriding the retry policy is not a very common use case, so we have to worry about good API docs and a clear reference implementation, I think linking to it would be enough.

landim: With API docs and good understanding (policy behaviors) would be sufficient to implement a custom retry policy.

jorgebay:
Thanks for your suggestions @idanto and sorry to take so long to close this PR.

Since 4.0, the retry policy is not engaged when a query is not idempotent, see:
- https://docs.datastax.com/en/developer/nodejs-driver/4.0/features/tuning-policies/#query-idempotence
- https://docs.datastax.com/en/developer/nodejs-driver/4.0/features/tuning-policies/#retry-policy
- https://docs.datastax.com/en/developer/nodejs-driver/4.0/upgrade-guide/#query-idempotency-and-retries

PR #247, opened 2017-10-30
Updating connection pooling docs

Sfeinste:
The connection pooling docs are using incorrect syntax for property assignment, `=` has been replaced with `:`.


jorgebay:
Merged 🎉  ! Thanks @Sfeinste !

PR #250, opened 2017-12-14
Call super in NoHostAvailableError constructor

jorgebay:
- Call super in NoHostAvailableError constructor
- Exclude all stack frames above new instance call

### DIFF
--- a/test/unit/error-tests.js
+++ b/test/unit/error-tests.js
@@ -48,7 +48,7 @@ function assertError(err, errorConstructor) {
   assert.ok(err.message);
   assert.ok(err.info);
   assert.ok(err.stack);
-  helper.assertInstanceOf(err, Error);
+  helper.assertInstanceOf(err, errors.DriverError);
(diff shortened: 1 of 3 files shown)


test/unit/error-tests.js
```diff
@@ -0,0 +1,55 @@
...
+var fileName = path.basename(__filename);
+
+describe('DriverError', function () {
+  it('should inherit from Error and have properties defined', function () {
+    var error = new errors.DriverError('My message');
+    assertError(error, errors.DriverError);
+  });
+});
+describe('NoHostAvailableError', function () {
+  it('should inherit from DriverError and have properties defined', function () {
```
tolbertam: To effectively ensure this inherits from `DriverError` shouldn't `assertError` also check for that?    I see `helper.assertInstanceOf(err, Error);` but not `helper.assertInstanceOf(err, errors.DriverError);`

jorgebay: My intention was to assert instance of `DriverError`, I've pushed a fix.

PR #253, opened 2017-12-19
DCAwareRoundRobinPolicy: Warn when local datacenter is not specified

tolbertam:
For [NODEJS-378](https://datastax-oss.atlassian.net/browse/NODEJS-378).

If no `localDC` was configured with `DCAwareRoundRobinPolicy` a log message like the following will be emitted at `warning` level:

> warning -- No local DC was provided with DCAwareRoundRobinPolicy.  Using discovered DC 'datacenter1' from host 127.0.0.1:9042.  Future releases will require local DC to be provided.

### DIFF
--- a/lib/policies/load-balancing.js
+++ b/lib/policies/load-balancing.js
@@ -132,10 +132,12 @@ DCAwareRoundRobinPolicy.prototype.init = function (client, hosts, callback) {
   if (!this.localDc) {
     //get the first alive local, it should be local on top
     var hostsArray = hosts.values();
+    var dcHost;
     for (var i = 0; i < hostsArray.length; i++) {
       var h = hostsArray[i];
       if (h.datacenter) {
         this.localDc = h.datacenter;
+        dcHost = h;
         break;
       }
     }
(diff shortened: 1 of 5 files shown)


lib/policies/load-balancing.js
```diff
@@ -132,17 +132,22 @@ DCAwareRoundRobinPolicy.prototype.init = function (client, hosts, callback) {
...
         this.localDc = h.datacenter;
+        dcHost = h;
         break;
       }
     }
     //this should never happen but it does not hurt
     if (!this.localDc) {
       return callback(new errors.DriverInternalError('Local datacenter could not be determined'));
     }
+    client.log('warning', 'No local DC was provided with DCAwareRoundRobinPolicy' +
```
tolbertam: Message may not be perfect, feedback welcome :)

jorgebay: I think its clear enough.
I would use the word "data center" instead of the first occurrence of DC and replace the second occurrence of "provided" with "specified".

`'No local Data Center was provided with DCAwareRoundRobinPolicy'`
`'.  Future releases will require local DC to be specified.'`

jorgebay:
I've pushed some fixes for unrelated tests that were affected by this change.

lib/policies/load-balancing.js
tolbertam: 👍 will make those changes quickly.

tolbertam:
> I've pushed some fixes for unrelated tests that were affected by this change.

Good catch!  I didn't catch that.  I noticed that a test is now failing on 0.10 because of es6 incompatibility, but that will go away when this is merged.

PR #254, opened 2017-12-20
Replace var with const / let

jorgebay:
Part of NODEJS-406, its a large change set replacing occurrences of `var` with `const/let`.

Includes 2 new eslint rules: `no-var` and `prefer-const`.

I would like to have this pr open for the least amount of time, to avoid rebasing with other changes.
I'll do some quick benchmark tomorrow morning with Node.js 6 to see if it has any significant impact.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -92,7 +92,9 @@ module.exports = {
     "no-void": "error",
     "no-with": "error",
     "no-buffer-constructor": "error",
-    radix: "error"
+    radix: "error",
+    "no-var": "error",
+    "prefer-const": "error"
   },
   "globals": {
     "Buffer": false,
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ It can be **piped** downstream and provides automatic pause/resume logic (it buf
 client.stream('SELECT time, val FROM temperature WHERE station_id=', [ 'abc' ])
   .on('readable', function () {
     // 'readable' is emitted as soon a row is received and parsed
-    var row;
+    let row;
     while (row = this.read()) {
       console.log('time %s and value %s', row.time, row.val);
     }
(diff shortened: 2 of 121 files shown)


jorgebay:
I ran some benchmarks using 0a4ae89 as baseline. There is some negative impact when using Node.js 6 and some almost negligible positive impact with Node.js 8:

Benchmarks
throughput-test, 1 connection per host, max outstanding requests: 512.

Node.js 6.12.2
- **master** Insert: 17670.86 req/s; Select: 14256.08 req/s
- **NODEJS-406** Insert: 16742.19 req/s (🔽  5.25%); Select: 13551.67 req/s (🔽  4.94%)

Node.js 8.9.3
- **master** Insert: 33112.03 req/s; Select: 24329.81 req/s
- **NODEJS-406** Insert: 33476.54 req/s (🔼  1.1%); Select: 25851.15 (🔼  4.58%) req/s

I'm comfortable making this change as:
- There is an important benefit on the development side: `const` declarations are far safer than `var` (block-scoped and it can't be reassigned). Currently `const` declarations outweigh `let` declarations by 3 to 1 on `lib`.
- There is a positive-to-no impact on the performance under Node.js 8.9, which uses the V8 6.1 with [Ignition and TurboFan](https://v8project.blogspot.com.es/2017/05/launching-ignition-and-turbofan.html), the execution pipeline going forward in V8.

Let me know what you think, if you agree, I would like to merge it today or tomorrow.

jorgebay:
> Neat that you could use eslint as a way to validate this. I also noticed that you can use it to make the changes for you (with --fix), did you end up using that or doing this all by hand?

I realize that after 3 hours in (I was getting nowhere manually replacing), the `--fix` did its magic, there were a few hiccups regarding scope and closures (no variable hoisting with let/const) but it all went well from there.

jorgebay:
Rebased, I'll wait for CI builds to merge it.

PR #255, opened 2017-12-20
Modify ResponseError messages based on error contents

tolbertam:
For [NODEJS-400](https://datastax-oss.atlassian.net/browse/NODEJS-400).

Follows the behavior of other drivers by modifying error message contents based on the values of its contents.  I chose to follow the csharp driver's error messages as they were the easiest to follow in my opinion.

### DIFF
--- a/lib/readers.js
+++ b/lib/readers.js
@@ -308,12 +308,12 @@ FrameReader.prototype.readMetadata = function (kind) {
 };
 
 // templates for derived error messages.
-var _writeTimeoutQueryMessage = 'Server timeout during write query at consistency %s (%d peer(s) acknowledged the write over %d required)';
-var _writeTimeoutBatchLogMessage = 'Server timeout during batchlog write at consistency %s (%d peer(s) acknowledged the write over %d required)';
-var _writeFailureMessage = 'Server failure during write query at consistency %s (%d responses were required but only %d replicas responded, %d failed)';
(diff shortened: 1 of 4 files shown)


lib/readers.js
```diff
@@ -306,6 +307,14 @@ FrameReader.prototype.readMetadata = function (kind) {
   return meta;
 };
 
+// templates for derived error messages.
```
tolbertam: In the jira ticket we discussed the possibility of creating constructors for each error type or using some kind of builder, but it didn't take too much extra code to add this inline so I opted to take the simple approach of adding the logic in the existing method, but can change it if requested.

jorgebay:
Implementation looks great and using unit tests at parser level is a good call.

Can we add some integration tests using simulacron? (I commented on the ticket that the estimation was way off)

tolbertam:
> Can we add some integration tests using simulacron? (I commented on the ticket that the estimation was way off)

That's a good idea and will be pretty easy to pull off,  will do!

tolbertam:
Added integration tests and also handled [NODES-409](https://datastax-oss.atlassian.net/browse/NODEJS-409) to include `keyspace` and `table` for `alreadyExists` errors.

I need to make some changes for ES6 (var->let, use arrow functions, etc.) , will take care of that later today.

tolbertam:
Thanks!  I noticed there's a small issue with my tests with Nodejs 4 that i'll fix quick and i'll also make relevant changes around var/let, should have that pushed shortly and then i'll merge.

PR #256, opened 2017-12-21
Set warmup true at defaultClientOptions

landim:


jorgebay:
We need to change the jsdoc annotations and understand why the unit tests are failing (probably related to setup).

landim:
Addressed the JSDoc update, test fixed and ran a test to double check the expected behavior after change.

PR #257, opened 2017-12-22
Use class syntax and use arrow functions

jorgebay:
[As detailed in the ticket](https://datastax-oss.atlassian.net/browse/NODEJS-406) there is a feature on ES6 that prevents us from moving public declarations to class syntax (`TypeError: Class constructor cannot be invoked without 'new'`).

I've migrated some of the internal prototypes and included the eslint rule `constructor-super` which is super useful in implementations like `Request`.
I've also changed some occurrences of functions to arrow functions and, as explained in the ticket comments, I've included the rule `"arrow-body-style": ["error", "as-needed"]`.

I'm happy with the result, specially with modules like `requests.js`, `prepare-handler.js`, `operation-state.js`. If I have more time, I would like to migrate `stream-id-stack.js`, `tokenizer.js` and `writers.js`.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -98,7 +98,6 @@ module.exports = {
     "arrow-body-style": ["error", "as-needed"],
     "arrow-spacing": "error",
     "no-confusing-arrow": ["error", { "allowParens": true }],
-    "class-methods-use-this": "error",
     "yoda": "error",
     "constructor-super": "error"
   },
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -94,7 +94,13 @@ module.exports = {
     "no-buffer-constructor": "error",
     radix: "error",
     "no-var": "error",
-    "prefer-const": "error"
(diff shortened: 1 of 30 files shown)


jorgebay:
Good input @landim, I've changed the suggested fn declarations into arrows.

I've also used class syntax for writers, tokenizers and `StreamIdStack`.

jorgebay:
As this is a large change set, I'm pinging @landim and @tolbertam for reviewing to avoid conflicts with other dev branches.

tolbertam:
> As this is a large change set, I'm pinging @landim and @tolbertam for reviewing to avoid conflicts with other dev branches.

:+1: i'll take a look at it this morning.  I think it would be pretty easy to rebase my PR after this is merged as the changes are pretty localized there, so I'm ok with this one being merged first.

jorgebay:
ok, I'll take the opportunity and merge it, thanks!

PR #258, opened 2017-12-22
Use system.peers in protocol negotiation

tolbertam:
For [NODEJS-405](https://datastax-oss.atlassian.net/browse/NODEJS-405).  Pushing ahead of holiday break.

Adds support for checking the system.peers / local `cassandra_version` column to attempt to resolve the common protocol version among them.

Some notes:

1. Introduced `VersionNumber` type for parsing version numbers into types that we can use for comparison.  This might have been overkill as simple string comparison could have sufficed, but I think it is nice to have.
2. If we detect a common protocol version that is different than the one in the use by the control connection, the control connection is disconnected and reconnected at that version.  This is done before retrieving schema metadata.
3. There is some special case logic for when we detect C* 3.0+ nodes and older nodes that don't support protocol v3.   In this case we use the common protocol version between those nodes that at least support protocol v3.  In practice this won't be something we need to worry about as C* does not support nodes with versions that are multiple major versions apart.
4. Tests are currently skipped because there are some things that need to be added to simulacron to support the tests (some issues around provisioning and protocol supported messages that i've fixed locally and am still testing).

### DIFF
--- a/lib/control-connection.js
+++ b/lib/control-connection.js
@@ -12,6 +12,7 @@ const requests = require('./requests');
 const utils = require('./utils');
 const types = require('./types');
 const f = util.format;
+const VersionNumber = types.VersionNumber;
 // eslint-disable-next-line prefer-const
 let dns = require('dns');
 
@@ -24,6 +25,11 @@ const schemaChangeTypes = {
   updated: 'UPDATED',
   dropped: 'DROPPED'
 };
+
+const V2_0_0 = VersionNumber.parse('2.0.0');
+const V2_1_0 = VersionNumber.parse('2.1.0');
+const V2_2_0 = VersionNumber.parse('2.2.0');
+const V3_0_0 = VersionNumber.parse('3.0.0');
 
 /**
  * Creates a new instance of <code>ControlConnection</code>.
@@ -260,12 +266,13 @@ ControlConnection.prototype.refreshHosts = function (newNodesUp, callback) {
     return;
   }
   const self = this;
-  if (!this.host.protocolVersion) {
-    this.hosts.forEach(h => h.setProtocolVersion(self.protocolVersion));
-  }
   this.log('info', 'Refreshing local and peers info');
   const c = this.connection;
   const host = this.host;
+  if (!self.host.protocolVersion) {
+    self.host.setProtocolVersion(self.protocolVersion);
+  }
+
   utils.series([
     function getLocalInfo(next) {
       const request = new requests.QueryRequest(selectLocal, null, null);
@@ -279,6 +286,38 @@ ControlConnection.prototype.refreshHosts = function (newNodesUp, callback) {
       c.sendStream(request, null, function (err, result) {
         self.setPeersInfo(newNodesUp, err, result, next);
       });
+    },
+    function resolveAndSetProtocolVersion(next) {
+      if (!self.initialized) {
+        // resolve protocol version from highest common version among hosts.
+        const highestCommon = self._highestCommonProtocolVersion(self.protocolVersion, self.hosts);
+        const reconnect = highestCommon !== self.protocolVersion;
+
+        // set protocol version on each host.
+        self.protocolVersion = highestCommon;
+        self.hosts.forEach(h => h.setProtocolVersion(self.protocolVersion));
+
+        // if protocol version changed, reconnect.
+        if (reconnect) {
+          self.log('info', 'Reconnecting since the protocol version changed to ' + highestCommon);
+          utils.series([
+            function closeConnection(cNext) {
+              self.host.removeFromPool(c, cNext);
+            },
+            function borrowConnection(cNext) {
+              self.borrowHostConnection(self.host, function (err, connection) {
+                if (err) {
+                  cNext(err);
+                }
+                self.connection = connection;
+                cNext();
+              });
+            }
+          ], next);
+        } else {
+          next();
+        }
+      }
     },
     function getKeyspaces(next) {
       // to acquire metadata we need to specify the cassandra version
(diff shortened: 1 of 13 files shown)


lib/host.js
```diff
@@ -251,8 +251,8 @@ Host.prototype.checkHealth = function (connection) {
  * @internal
  * @ignore
  */
-Host.prototype.removeFromPool = function (connection) {
-  this.pool.remove(connection);
+Host.prototype.removeFromPool = function (connection, cb) {
```
tolbertam: Needed to add a callback here to detect when the connection is closed.  Not quite sure if my use of this is correct of if there is a better way (I want to remove the control connection from pool and create a new one when it's done).

lib/types/version-number.js
```diff
@@ -0,0 +1,113 @@
+'use strict';
```
tolbertam: This needs docs.  Also, should it be public?  Might be generally useful, but we could also consider hiding it..

test/integration/short/control-connection-simulator-tests.js
```diff
@@ -0,0 +1,138 @@
...
+              name: "dc0",
+              nodes: nodes
+            }
+          ]
+        }, next);
+      },
+      function connect(next) {
+        client.connect(function (err, result) {
+          // TODO: Check for downgrade.
+          if (expectedProtocolVersion < 3) {
```
tolbertam: Because simulacron doesn't support protocol v1 and v2 (as native-protocol library it depends on doesn't support it), we expect an error in this case.  I'll also add a query log message in simulacron when there is an attempt with a bad protocol version so we can validate it.    I'll probaly also add an integration test with a 2 node cluster with C* 2.1 and C* 2.0 to test that downgrade case.

lib/control-connection.js
```diff
@@ -12,6 +12,7 @@ const requests = require('./requests');
 const utils = require('./utils');
 const types = require('./types');
 const f = util.format;
+const VersionNumber = types.VersionNumber;
```
landim: +1 for VersionNumber. It might be "too much" for now, but on the long run it will payoff.

lib/types/version-number.js
landim: I don't think that, from a driver client perspective, it is necessary to make it public.

test/integration/short/control-connection-simulator-tests.js
```diff
@@ -0,0 +1,138 @@
...
+    it('should downgrade to protocol v3 with versions 2.2 & 2.1', testWithNodes(['2.2.11', '2.1.17'], 3));
+    it('should downgrade to protocol v2 with versions 2.2 & 2.0', testWithNodes(['2.2.11', '2.0.17'], 2));
+    it('should downgrade to protocol v1 with versions 2.2 & 1.2', testWithNodes(['2.2.11', '1.2.19'], 1));
+    it('should downgrade to protocol v2 with versions 2.1 & 2.0', testWithNodes(['2.1.17', '2.0.17'], 2));
+    it('should downgrade to protocol v1 with versions 2.1 & 1.2', testWithNodes(['2.1.17', '1.2.19'], 1));
+    it('should downgrade to protocol v1 with versions 2.0 & 1.2', testWithNodes(['2.0.17', '1.2.19'], 1));
+    // no need to downgrade since both support protocol V4.
+    it('should not downgrade with versions 3.0 & 2.2', testWithNodes(['3.0.13', '3.0.11', '2.2.9'], 4));
+    // can't downgrade because C* 3.0 does not support protocol V2.
+    it('should not downgrade with versions 3.0 & 2.0', testWithNodes(['3.0.13', '2.0.17'], 4));
```
landim: The only difference between "Can't downgrade" test and "No need to downgrade" test, from output perspective, is the log warning. Should verify it?

tolbertam: Good idea, I can check for that.

lib/control-connection.js
```diff
@@ -773,6 +812,74 @@ ControlConnection.prototype.reset = function (callback) {
...
+ * version away from each other.
+ *
+ * @param {Number} version Maximum version to consider.
+ * @param {Array.<Host>} hosts The hosts to determine highest protocol version from.
+ * @return {Number} Highest supported protocol version among hosts.
+ */
+ControlConnection.prototype._highestCommonProtocolVersion = function(version, hosts) {
+  let maxVersion = version;
+  // whether or not protocol v3 is required (nodes detected that don't support < 3).
+  let v3Requirement = false;
```
jorgebay: We _could_ use a `minVersion` field but its maybe an overkill, because dropping old protocol versions is not something that happens too often on the server side.

lib/control-connection.js
```diff
@@ -773,6 +812,74 @@ ControlConnection.prototype.reset = function (callback) {
...
+    if (!h.cassandraVersion || h.cassandraVersion.length === 0) {
+      self.log('warning', 'Encountered host ' + h.address + ' with no cassandra version,' +
+        ' skipping as part of protocol version evaluation');
+      return;
+    }
+
+    try {
+      const cassandraVersion = VersionNumber.parse(h.cassandraVersion);
+      self.log('verbose', 'Encountered host ' + h.address + ' with cassandra version ' + cassandraVersion);
+      if (cassandraVersion.compare(V3_0_0) >= 0) {
```
jorgebay: Have you thought about how are we going to extend this behaviour on the DSE Driver?
Maybe we can move this logic to `protocol-version` module?

lib/control-connection.js
```diff
@@ -773,6 +812,74 @@ ControlConnection.prototype.reset = function (callback) {
...
+    }
+  });
+
+  if (v3Requirement && maxVersion < types.protocolVersion.v3) {
+    self.log('warning', 'Detected hosts with maximum protocol version of ' + maxVersion +
+      ' but there are some hosts that require at least version 3, will not be able to connect to these older hosts');
+    maxVersion = maxVersionWith3OrMore;
+  }
+
+  self.log('verbose', 'Resolved protocol version ' + maxVersion + ' as the highest common protocol version among hosts');
```
jorgebay: As the decimal representation of the protocol version could be misleading, we can use the hex representation instead: `'Resolved protocol version 0x' + maxVersion.toString(16)`

lib/types/version-number.js
```diff
@@ -0,0 +1,113 @@
...
+
+const _versionPattern = /(\d+)\.(\d+)(\.\d+)?(\.\d+)?([~-]\w[.\w]*(?:-\w[.\w]*)*)?(\+[.\w]+)?/;
```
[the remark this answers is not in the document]
jorgebay: Maybe something like `/(\d+)\.(\d+)(?:\.(\d+))?(?:\.(\d+)?)?(?:[-~](\w+))?/` could simplify the logic from below, groups from 1 to 5: https://regex101.com/r/XzuhD4/1

lib/control-connection.js
```diff
@@ -282,6 +287,40 @@ ControlConnection.prototype.refreshHosts = function (newNodesUp, callback) {
         self.setPeersInfo(newNodesUp, err, result, next);
       });
     },
+    function resolveAndSetProtocolVersion(next) {
+      if (!self.initialized) {
```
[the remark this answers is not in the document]
tolbertam: woops, will fix.

lib/control-connection.js
```diff
@@ -282,6 +287,40 @@ ControlConnection.prototype.refreshHosts = function (newNodesUp, callback) {
...
+        self.hosts.forEach(function (h) {
+          h.setProtocolVersion(self.protocolVersion);
+        });
+
+        // if protocol version changed, reconnect.
+        if (reconnect) {
+          self.log('info', 'Reconnecting since the protocol version changed to ' + highestCommon);
+          utils.series([
+            function closeConnection(cNext) {
+              self.host.removeFromPool(c, cNext);
```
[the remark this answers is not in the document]
tolbertam: Thanks for the suggestion, will look into that.

lib/control-connection.js
tolbertam: Ah yes, that's a much better idea, will do that.  I had initially implemented it that way but reverted the logic back to control-connecton and not sure why.

In terms the of the DSE driver, I hope it should just be a matter of comparing the `dse_version` column and maybe falling back to `cassandra_version` if that isn't present (although I think it always is).

lib/types/version-number.js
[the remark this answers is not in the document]
tolbertam: good idea, i'll give that a look as the substring logic is kinda hacky.

lib/control-connection.js
tolbertam: Will think about how to make this better.  You are right that things will likely get messy next time a protocol version is dropped (C* 4.0?) so it would be good to take care of that now if there is an elegant way of doing that.

lib/types/version-number.js
jorgebay: I think its good as is for now, inside the `types` folder but not exported in the `types` module.
You could add `@ignore` to the jsdoc when documenting, if you want to keep it this way

lib/types/version-number.js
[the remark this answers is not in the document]
tolbertam:
```
/(\d+)\.(\d+)(?:\.(\d+))?(?:\.(\d+)?)?(?:[-~](\w+))?/
```

was pretty close to what I needed.  I had to add some more to parse the multiple preRelease identifiers and build number:

```
/(\d+)\.(\d+)(?:\.(\d+))?(?:\.(\d+)?)?(?:[-~]([\w+]*(?:-\w[.\w]*)*))?(?:\+([.\w]+))?/;
```

lib/control-connection.js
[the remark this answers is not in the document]
tolbertam: As far as I can tell, removing the listener for `socketClose` is not required, but i'll look more into that.  Instead of creating another function on `Host` i just inlined it into control connection (hope that's ok).

tolbertam:
Processed all feedback except w/r/t the discussion about `minVersion` [here](https://github.com/datastax/nodejs-driver/pull/258#discussion_r159420005), will think more about that and come up with a solution.  I also would like to add some unit tests for `highestCommonVersion` and an long integration test with a ccm cluster for testing with older protocol versions.  Should finish that up today.

lib/types/protocol-version.js
```diff
@@ -127,6 +135,73 @@ const protocolVersion = {
...
+   * as those cannot be connected to.  In general this will not be a problem as C* does
+   * not support clusters with nodes that have versions that are more than one major
+   * version away from each other.
+   * @param {Connection} connection Connection hosts were discovered from.
+   * @param {Array.<Host>} hosts The hosts to determine highest protocol version from.
+   * @return {Number} Highest supported protocol version among hosts.
+   */
+  getHighestCommon: function(connection, hosts) {
+    const log = connection.log ? connection.log.bind(connection) : utils.noop;
+    let maxVersion = connection.protocolVersion;
```
tolbertam: Possible issue here, though probably not a valid use case:

control connection is to a host that supports V2 at maximum (i.e. C* 2.0, there are other hosts that require at least V3 (i.e. C* 3.0).

In this case this method will return V2, and hosts that require V3 can't be connected to.  I think the best we can do here is log this scenario, as we do at line 198 when we detect hosts that don't support V3.

tolbertam: I added some additional logging for this case . Technically it's not valid since you shouldn't be able to have a cluster with nodes separated by more than one major version, so i log something at error.

test/integration/short/control-connection-simulator-tests.js
tolbertam: I added a long integration test for this, takes about ~80seconds (as after bootstrapping a node it takes 30 seconds to join, waits for gossip to settle, etc.), but looks to be working.

lib/types/protocol-version.js
jorgebay: I think logging an error is enough, no sane cluster should have this issue.

lib/types/protocol-version.js
```diff
@@ -127,6 +135,81 @@ const protocolVersion = {
       return 0;
     }
     return version - 1;
+  },
+
+  /**
+   * Computes the highest supported protocol version collectively by the given hosts.
+   *
+   * Considers the cassandra_version of the input hosts to determine what protocol versions
+   * are supported and uses the highest common protocol version amont them.
```
jorgebay: spelling `amont`

lib/types/protocol-version.js
```diff
@@ -25,6 +32,7 @@ const protocolVersion = {
   v5: 0x05,
   maxSupported: 0x04,
   minSupported: 0x01,
+  log: utils.log,
```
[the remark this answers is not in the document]
tolbertam: ah yes, I think that was something i experimented with but then backed off, will remove, thanks.

lib/control-connection.js
tolbertam: I think for now I'd like to leave this as it is.  I think it would be better to address this next time protocol version support is removed from a C* version, and also we don't have to worry about this scenario as much because there shouldn't be nodes in the cluster separated by more than one version anyways.

tolbertam:
Removed (wip) tag as I think this should be ready to go

tolbertam:
Will rebase this shortly and if all tests pass will merge.

tolbertam:
The appveyor builds are marked as failed because of a weird issue with caches in appveyor, the tests appeared to have passed, so merging.

PR #259, opened 2018-01-04
Verify Buffer.from() is not Int8Array.from()

jorgebay:
Older versions of Node.js 4 imported Int8Array.from() function on Buffer, which is different from Buffer.from() introduced in Node.js 4.5.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -101,6 +101,7 @@ module.exports = {
     "Promise": true,
     "Symbol": false,
     "Uint16Array": false,
+    "Int8Array": false,
     "process": false,
     "setTimeout": false,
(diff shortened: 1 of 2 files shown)


jorgebay:
I've tested manually with Node.js 4.0 and tests like `test/unit/inet-address-tests.js` that used to fail now pass.

I don't think its necessary to add more tests or to change the used node.js revision on CI (which are more recent).

tolbertam:
I've noticed travis workers have been having odd issues lately (i.e. https://travis-ci.org/datastax/nodejs-driver/jobs/324993020) that seem environmental so I retriggered the travis pr build.

PR #260, opened 2018-01-08
Move Object.defineProperty outer scope

landim:


jorgebay:
hm... I think `Object.defineProperty(MyClass.prototype)` doesn't work as expected:

For example:

```javascript
function InetAddress(buffer) {
  this.buffer = buffer;
}

Object.defineProperty(InetAddress.prototype, 'length', { 
  get: () => this.buffer.length, 
  enumerable: true
});
```

Doesn't make the instance properties enumerable:

```javascript
var ip = new InetAddress(Buffer.from([127, 0, 0, 1]));
console.log(ip); // "InetAddress { buffer: <Buffer 7f 00 00 01> }"
console.log(Object.keys(ip)); // "[ 'buffer' ]"
```

landim:
Moving the defineProperty to property makes the property not enumerable for the object only the prototype. It can only be _enumerable_ if:
`
console.log(Object.keys(Object.getPrototypeOf(ip))); // "[ 'buffer', 'length', 'equals', 'getBuffer', etc.. ]"
`
From user perspective it would be nicer if the property is enumerable for debugging/logging. Having this in mind, another way would be to simple use at function constructor:
`
this.length = this.buffer.length;
` 
But would be possible to set new values and It would impact on the behavior of each client that further invoke _ip.length_ property.

Another approach would be to take advantage of Object.freeze as InetAddress, LocalDate, LocalTime and Tuple are immutable objects.
For example:
`
function InetAddress(buffer) {
  if (!(buffer instanceof Buffer) || (buffer.length !== 4 && buffer.length !== 16)) {
    throw new TypeError('The ip address must contain 4 or 16 bytes');
  }
  this.buffer = buffer;
  this.length = this.buffer.length;
  this.version = this.buffer.length === 4 ? 4 : 6;
  if (this.constructor === InetAddress) { //Simple verification to not freeze subtypes completely 
    Object.freeze(this); 
  }
}
`
The output would be: 
`
var ip = new InetAddress(Buffer.from([127, 0, 0, 1]));
console.log(ip); // "InetAddress { buffer: <Buffer 7f 00 00 01> }"
console.log(Object.keys(ip)); // [ 'buffer', 'length', 'version' ]
console.log(ip.length); // 4
ip.length = 1; //on strict mode will throw exception
console.log(ip.length); // 4
`

wdyt?

jorgebay:
The more I think about it, the more convinced I am about ditching `Object.defineProperty()` altogether.

I wanted to use `Object.defineProperty()` to elegantly expose properties as in C# / Python, but it's doesn't completely fixes the problems and is not very common on the Node.js world.

I'm comfortable in exposing both public and private properties both writable and enumerable (`this._myProp = value`), like its being done in Node.js core libraries and most libraries.
Regarding calculated properties (currently with getters), like `LocalTime.hour`, it should have been designed as methods (`getHour()`), which is more js-like instead of trying to get away by doing it with properties.

For `LocalTime` and `LocalDate` calculated properties, I would calculate them upfront to overcome this issue and move on. Most of the use case would probably call `toString()` or get the value of properties, so using a lazy approach has no significant benefits.

For new classes in the API, I would use as little properties as possible and expose most of the functionality via methods (more as its common in js and java).

landim:
Addressed:
- Removed the _Object.defineProperty_, although it was creating a design more alike Python and C#, it was different from Nodejs standard
- Definition of the properties inside function constructor: the properties are enumerable and calculated upfront (as those are calculated on top of immutable objects)

lib/types/local-date.js
```diff
@@ -64,12 +64,13 @@ function LocalDate(year, month, day) {
...
-  //If date cannot be represented yet given a valid days since epoch, track
-  //it internally.
-  const value = isNaN(this.date.getTime()) ? year : null;
-  Object.defineProperty(this, '_value', { enumerable: false, value: value });
-
-  const self = this;
+  /** 
+   * If date cannot be represented yet given a valid days since epoch, track
+   * it internally.
+   * @ignore
```
jorgebay: `@private` doclet prevents a property / class from being included, there is no need to add also `@ignore`.

lib/types/local-date.js
```diff
@@ -78,31 +79,23 @@ function LocalDate(year, month, day) {
    * @type Number
    * @memberof module:types~LocalDate#
```
jorgebay: I think that `@memberof` and `name` is not needed anymore, as the prop is defined using `this.propName` below, jsdoc will infer those.

lib/types/tuple.js
```diff
@@ -22,14 +22,21 @@ function Tuple() {
...
   }
-  Object.defineProperty(this, 'elements', { value: elements, enumerable: false, writable: false });
+ 
+  /** 
+   * Immutable elements of Tuple object.
+   * @ignore
+   * @private
+   */
+ 
+   this.elements = elements;
```
jorgebay: eslint is complaining about 3 spaces :)

landim:
Addressed the jsdocs nit

PR #261, opened 2018-01-09
Expose node token and range information

tolbertam:
For [NODEJS-95](https://datastax-oss.atlassian.net/browse/NODEJS-95).  Not much is done yet, but had some things I wanted to discuss before proceeding further.  Comments added inline.

### DIFF
--- a/lib/metadata/index.js
+++ b/lib/metadata/index.js
@@ -139,7 +139,7 @@ Metadata.prototype.buildTokens = function (hosts) {
   const tokenRanges = new Set();
   if(this.ring.length === 1) {
     // If there is only one token, return the range ]minToken, minToken]
-    const min = this.tokenizer.minToken;
+    const min = this.tokenizer.minToken();
     tokenRanges.add(new TokenRange(min, min, this.tokenizer));
   } else {
     for (let i = 0; i < this.ring.length; i++) {
@@ -310,6 +310,11 @@ Metadata.prototype.getTokenRangesForHost = function (keyspaceName, host) {
       // the keyspace was not found, the metadata should be loaded beforehand
       return null;
     }
+  }
+
+  // If the ring has only 1 token, just return the ranges as we should only have a single node cluster.
+  if (this.ring.length === 1) {
+    return this.getTokenRanges();
   }
 
   // TODO: This code is duplicated from getReplicas, consider making common method.
--- a/lib/metadata/index.js
+++ b/lib/metadata/index.js
@@ -11,7 +11,7 @@ const errors = require('../errors');
 const types = require('../types');
 const requests = require('../requests');
 const schemaParserFactory = require('./schema-parser');
-const TokenRange = require('../token-range');
+const TokenRange = require('../token').TokenRange;
 
 /**
  * @const
@@ -99,14 +99,16 @@ Metadata.prototype.buildTokens = function (hosts) {
   if (!this.tokenizer) {
     return this.log('error', 'Tokenizer could not be determined');
   }
+  const self = this;
   //Get a sorted array of tokens
   const allSorted = [];
   //Get a map of <token, primaryHost>
   const primaryReplicas = {};
   //Depending on the amount of tokens, this could be an expensive operation
   const hostArray = hosts.values();
-  const parser = this.tokenizer.parse;
-  const compare = this.tokenizer.compare;
+  const compare = function(val1, val2) {
+    return val1.compare(val2);
+  };
   const stringify = this.tokenizer.stringify;
   const datacenters = {};
   hostArray.forEach(function (h) {
@@ -114,9 +116,9 @@ Metadata.prototype.buildTokens = function (hosts) {
       return;
     }
     h.tokens.forEach(function (tokenString) {
-      const token = parser(tokenString);
+      const token = self.tokenizer.parse(tokenString);
       utils.insertSorted(allSorted, token, compare);
-      primaryReplicas[stringify(token)] = h;
+      primaryReplicas[stringify(token.getValue())] = h;
     });
     let dc = datacenters[h.datacenter];
     if (!dc) {
@@ -151,7 +153,7 @@ Metadata.prototype.buildTokens = function (hosts) {
   //Compute string versions as it's potentially expensive and frequently reused later
   this.ringTokensAsStrings = new Array(allSorted.length);
   for (let i = 0; i < allSorted.length; i++) {
-    this.ringTokensAsStrings[i] = stringify(allSorted[i]);
+    this.ringTokensAsStrings[i] = stringify(allSorted[i].getValue());
   }
   //Datacenter metadata (host length and racks)
   this.datacenters = datacenters;
@@ -296,8 +298,43 @@ Metadata.prototype.getTokenRanges = function () {
  * @param {Host} host The host.
  * @returns {Set<TokenRange>} Ranges for the keyspace on this host
  */
-Metadata.prototype.getTokenRanges = function (keyspaceName, host) {
-  // TODO
+Metadata.prototype.getTokenRangesForHost = function (keyspaceName, host) {
+  if (!this.ring) {
+    return null;
+  }
+  const self = this;
+  let keyspace;
+  if (keyspaceName) {
+    keyspace = this.keyspaces[keyspaceName];
+    if (!keyspace) {
+      // the keyspace was not found, the metadata should be loaded beforehand
+      return null;
+    }
+  }
+
+  // TODO: This code is duplicated from getReplicas, consider making common method.
+  if (!keyspace.replicas) {
+    //Calculate replicas the first time for the keyspace
+    keyspace.replicas =
+      keyspace.tokenToReplica(this.tokenizer, this.ringTokensAsStrings, this.primaryReplicas, this.datacenters);
+  }
+
+  // TODO: Consider caching/precomputing keyspace -> (host, range) as this could be an expensive operation.
+  const ranges = new Set();
+  // find tokens containing host and identify ranges beginning with that token.
+  Object.keys(keyspace.replicas).forEach((tokenStr) => {
+    const replicas = keyspace.replicas[tokenStr];
+    if(replicas.includes(host)) {
+      // identify ranges that start with this token.
+      const token = self.tokenizer.parse(tokenStr);
+      this.tokenRanges.forEach((tokenRange) => {
+        if (tokenRange.start.equals(token)) {
+          ranges.add(tokenRange);
+        }
+      });
+    }
+  });
+  return ranges;
 };
 
 Metadata.prototype.log = utils.log;
--- a/lib/metadata/index.js
+++ b/lib/metadata/index.js
@@ -114,7 +114,7 @@ Metadata.prototype.buildTokens = function (hosts) {
     h.tokens.forEach((tokenString) => {
       const token = this.tokenizer.parse(tokenString);
       utils.insertSorted(allSorted, token, (t1, t2) => t1.compare(t2));
-      primaryReplicas[stringify(token.getValue())] = h;
+      primaryReplicas[stringify(token._value)] = h;
     });
     let dc = datacenters[h.datacenter];
     if (!dc) {
@@ -149,7 +149,7 @@ Metadata.prototype.buildTokens = function (hosts) {
   //Compute string versions as it's potentially expensive and frequently reused later
   this.ringTokensAsStrings = new Array(allSorted.length);
   for (let i = 0; i < allSorted.length; i++) {
-    this.ringTokensAsStrings[i] = stringify(allSorted[i].getValue());
+    this.ringTokensAsStrings[i] = stringify(allSorted[i]._value);
   }
   //Datacenter metadata (host length and racks)
   this.datacenters = datacenters;
@@ -243,19 +243,26 @@ Metadata.prototype._getKeyspaceReplicas = function (keyspace) {
 };
 
 /**
- * Gets the host list representing the replicas that contain such partition.
+ * Gets the host list representing the replicas that contain the given partition key, token or token range.
  * <p>
  *   It uses the pre-loaded keyspace metadata to retrieve the replicas for a token for a given keyspace.
  *   When the keyspace metadata has not been loaded, it returns null.
  * </p>
  * @param {String} keyspaceName
- * @param {Buffer} tokenBuffer
+ * @param {Buffer|Token|TokenRange} token Can be Buffer (serialized partition key), Token or TokenRange
  * @returns {Array}
  */
-Metadata.prototype.getReplicas = function (keyspaceName, tokenBuffer) {
+Metadata.prototype.getReplicas = function (keyspaceName, token) {
   if (!this.ring) {
     return null;
   }
+  if (util.isBuffer(token)) {
+    token = this.tokenizer.hash(token);
+  }
+  if (token instanceof TokenRange) {
+    token = token.end;
+  }
+
   let keyspace;
   if (keyspaceName) {
     keyspace = this.keyspaces[keyspaceName];
@@ -264,7 +271,6 @@ Metadata.prototype.getReplicas = function (keyspaceName, tokenBuffer) {
       return null;
     }
   }
-  const token = this.tokenizer.hash(tokenBuffer);
   let i = utils.binarySearch(this.ring, token, (t1, t2) => t1.compare(t2));
   if (i < 0) {
     i = ~i;
@@ -323,12 +329,43 @@ Metadata.prototype.getTokenRangesForHost = function (keyspaceName, host) {
 
   // for each range, find replicas for end token, if replicas include host, add range.
   this.tokenRanges.forEach((tokenRange) => {
-    const replicasForToken = replicas[this.tokenizer.stringify(tokenRange.end.getValue())];
+    const replicasForToken = replicas[this.tokenizer.stringify(tokenRange.end._value)];
     if (replicasForToken.indexOf(host) !== -1) {
       ranges.add(tokenRange);
     }
   });
   return ranges;
+};
+
+/**
+ * Constructs a Token from the input buffer(s) or string input.  If a string is passed in
+ * it is assumed this matches the token representation reported by cassandra.
+ * @param {Array[Buffer]|Buffer|String} components 
+ * @returns {Token} constructed token from the input buffer.
+ */
+Metadata.prototype.newToken = function (components) {
+  if (!this.tokenizer) {
+    throw new Error('Partitioner not established.  This should only happen if metadata was disabled or you have not connected yet.');
+  }
+  if (util.isArray(components)) {
+    return this.tokenizer.hash(Buffer.concat(components));
+  } else if (util.isString(components)) {
+    return this.tokenizer.parse(components);
+  }
+  return this.tokenizer.hash(components);
+};
+
+/**
+ * Constructs a TokenRange from the given start and end tokens.
+ * @param {Token} start 
+ * @param {Token} end 
+ * @returns TokenRange build range spanning from start (exclusive) to end (inclusive).
+ */
+Metadata.prototype.newTokenRange = function(start, end) {
+  if (!this.tokenizer) {
+    throw new Error('Partitioner not established.  This should only happen if metadata was disabled or you have not connected yet.');
+  }
+  return new TokenRange(start, end, this.tokenizer);
 };
 
 Metadata.prototype.log = utils.log;
--- a/lib/metadata/index.js
+++ b/lib/metadata/index.js
@@ -99,7 +99,6 @@ Metadata.prototype.buildTokens = function (hosts) {
   if (!this.tokenizer) {
     return this.log('error', 'Tokenizer could not be determined');
   }
-  const self = this;
   //Get a sorted array of tokens
   const allSorted = [];
   //Get a map of <token, primaryHost>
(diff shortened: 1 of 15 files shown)


lib/token-range.js
```diff
@@ -0,0 +1,155 @@
...
+   * (referring to the number of tokens, not the actual amount of data).
+   *
+   * Splitting an empty range is not permitted.  But not that, in edge
+   * cases, splitting a range might produce one or more empty ranges.
+   *
+   * @param {Number} numberOfSplits Number of splits to make.
+   * @returns {TokenRage[]} Split ranges.
+   * @throws {Error} If splitting an empty range.
+   */
+  splitEvenly(numberOfSplits) {
```
tolbertam: I started by porting all the method definitions from the java driver as a starting point, but I was questioning if it was necessary to implement all of the methods, my thoughts:

1. splitEvently, isEmpty, isWrappedAround, unwrap, contains, equals, compare, toString:  All seem revelant, useful and easy to implement.
2. intersects, intersectWith, mergeWith:  Are useful, but complex.  It is nice for merging/combining ranges (i.e. "Give me a range that applies to replicas X, Y, Z.).  Wasn't sure if I should implement these, but wouldn't take me long.

tolbertam:
Other thoughts:

0. Should we make a specific `Token` type that wraps the raw value, or expose the raw values directly?  This could be useful for understanding what 'type' a token is, however in the the general case users will only ever work with one token type (that of their ring).
1. `Host` already has a `tokens` member which is a array of string coming from the `tokens` column of the peers table.  If we do 0, should be alter this or provide an additional member on `Host` to get its tokens?

lib/token-range.js
landim: 1: definitely +1
2: +1 as it would help on non-trivial table scans

landim:
0-1. From Design perspective its a good call, but I dont think it is necessary for now.
2. Although is not easily related, imo its easier to understand the ranges.

lib/token-range.js
jorgebay: - 1: +1
- 2: I think exposing the token info is a good step but it doesn't have to have all the utilities in the java driver, as long as we enable users to create those methods on their own if needed.
I would leave it for the end and if there is enough time, implement them.

jorgebay:
0. I think we should expose a `Token` class as part of this task. 
1. hm... I'm not sure. The API docs only specifies that is an `Array`, so we could get away of changing the underlying type to `Array<Token>`. It's a small breaking change but its not that big a deal if `Token.toString()` creates the same representation and we can document it in the changelog.
2. `tokenizers` should continue to be internal, `stringify()` is only needed to overcome the poor support for `Map` and `Set` in js. I think `Token.toString()` should provide the same info as C* but `Murmur3Tokenizer.stringify()` behaviour should not be changed (stringify: fastest unique string hash in js for the given token type). Makes sense?

lib/token-range.js
tolbertam: 👍 sounds like a good approach, if have time to get to it i’ll implement them.

tolbertam:
Agreed on everything!  Thanks for the feedback all 👍 .  Will proceed with creating a `Token` class, adjusting `Hosts.tokens` to be `Array<Token>`, and having the `Murmur3Token.toString()` implementation match that of system.peers while leaving `Murmur3Tokenizer.stringify()` in tact.  Should be able to get the bulk of that done today.

lib/metadata/index.js
```diff
@@ -262,6 +281,62 @@ Metadata.prototype.getReplicas = function (keyspaceName, tokenBuffer) {
...
+  }
+
+  // TODO: This code is duplicated from getReplicas, consider making common method.
+  if (!keyspace.replicas) {
+    //Calculate replicas the first time for the keyspace
+    keyspace.replicas =
+      keyspace.tokenToReplica(this.tokenizer, this.ringTokensAsStrings, this.primaryReplicas, this.datacenters);
+  }
+
+  // TODO: Consider caching/precomputing keyspace -> (host, range) as this could be an expensive operation.
```
tolbertam: wdyt?  Should I store something like `keyspace.hostToRanges` similar to how we do something for `keyspace.replicas`?

lib/token.js
```diff
@@ -0,0 +1,276 @@
...
+  }
+}
+
+/**
+ * Represents a token from a Cassandra ring where the partitioner
+ * is Murmur3Partitioner.
+ * 
+ * The raw token type is a varint (represented by MutableLong).
+ */
+class Murmur3Token extends Token {
```
tolbertam: Should `Murmur3Token` / `RandomToken` / `ByteOrderedToken` be exported to public?  The java driver doesn't do it, but I also don't see the harm in it and might be nice for instanceof checks (although probably not likely to be used).

lib/types/mutable-long.js
```diff
@@ -79,6 +79,10 @@ MutableLong.fromString = function fromString(str, radix) {
   return result;
 };
 
+MutableLong.toString = function toString() {
```
tolbertam: I think this is doable, but it will be challenging given that i'm not very familiar with the `MutableLong` code (and I'm not very good at bit-twiddling).  Will likely save this towards the end.

lib/token.js
jorgebay: I'd say we only export `Token` but not the actual implementations.

lib/types/mutable-long.js
jorgebay: Probably you meant `MutableLong.prototype.toString` :)
I think we could just call `this.toImmutable().toString()`. Otherwise we will have to implement division operation which is somehow complex.

tolbertam: Hah!  Yeah, that would be....much easier 😄 .   Definitely will go that route.

tolbertam:
Fleshed out `TokenRange` implementation some more and added tokenizer-specific `split` code (ByteOrdered one is currently TODO).  Needs more testing and I broke some stuff, will work on that tomorrow.

lib/tokenizer.js
```diff
@@ -227,7 +268,61 @@ class Murmur3Tokenizer extends Tokenizer {
...
+
+  minToken() {
+    if (!this._minToken) {
+      // minimum long value.
+      this._minToken = this.parse('-9223372036854775808');
+    }
+    return this._minToken;
+  }
+
+  maxToken() {
```
tolbertam: I'll likely change all of these to constants.

lib/tokenizer.js
```diff
@@ -299,15 +440,81 @@ class ByteOrderedTokenizer extends Tokenizer {
...
+
+  split(start, end, numberOfSplits) {
+    const tokenOrder = start.compare(end);
+
+    if (tokenOrder === 0 && start.equals(this.minToken())) {
+      throw new Error("Cannot split whole ring with ordered partitioner");
+    }
+
+    let startVal, endVal, range, ringLength, ringEnd;
+    const longNumberOfSplits = MutableLong.fromNumber(numberOfSplits);
```
tolbertam: Should work from `Integer` instead of `MutableLong` here, will fix.

tolbertam:
was a bit behind master so rebased

lib/tokenizer.js
```diff
@@ -295,19 +436,115 @@ class ByteOrderedTokenizer extends Tokenizer {
...
+      } else {
+        break;
+      }
+    }
+    return new token.ByteOrderedToken(value.slice(0, zeroIndex), this);
   }
 
   stringify(value) {
-    return value.toString('hex');
+    return value.toString('hex').toUpperCase();
```
tolbertam: To match C*, although not really important.

lib/tokenizer.js
```diff
@@ -295,19 +436,115 @@ class ByteOrderedTokenizer extends Tokenizer {
   }
 
   /**
-   * @param {Buffer|Array} value
+   * @param {Buffer} value
    * @returns {Buffer}
    */
   hash(value) {
-    return value;
+    // strip any trailing zeros as tokens with trailing zeros are equivalent
```
tolbertam: I noticed the java driver does this.  I don't think it's required, but if it wasn't done here, would have to update compare/equals methods to account for it

lib/tokenizer.js
```diff
@@ -295,19 +436,115 @@ class ByteOrderedTokenizer extends Tokenizer {
...
   }
 
   stringify(value) {
-    return value.toString('hex');
+    return value.toString('hex').toUpperCase();
   }
 
   parse(value) {
-    return utils.allocBufferFromString(value);
+    return this.hash(utils.allocBufferFromString(value, 'hex'));
```
tolbertam: I think this might have been previously broken, the tokens returned from C* are hexstrings so should be parsed that way:

```
Datacenter: datacenter1
==========
Address    Rack        Status State   Load            Owns                Token                                       
                                                                          56713727820156410577229101238628035242      
127.0.0.1  rack1       Up     Normal  46.66 KB        66.67%              00                                          
127.0.0.3  rack1       Up     Normal  46.69 KB        66.67%              0113427455640312821154458202477256070484    
127.0.0.2  rack1       Up     Normal  46.69 KB        66.67%              56713727820156410577229101238628035242  
```

lib/metadata/index.js
```diff
@@ -98,24 +99,26 @@ Metadata.prototype.buildTokens = function (hosts) {
   if (!this.tokenizer) {
     return this.log('error', 'Tokenizer could not be determined');
   }
+  const self = this;
```
jorgebay: Just a suggestion, you can use arrow functions below and avoid the need of a `self` field.

lib/metadata/index.js
jorgebay: I don't think it would be likely for a user to call `getTokenRangesForHost()` with the same parameters more than once.

lib/metadata/index.js
```diff
@@ -262,6 +281,67 @@ Metadata.prototype.getReplicas = function (keyspaceName, tokenBuffer) {
...
+  return this.tokenRanges;
+};
+
+/**
+ * Gets the token ranges that are replicated on the given host, for
+ * the given keyspace.
+ *
+ * @param {String} keyspaceName The name of the keyspace to get ranges for.
+ * @param {Host} host The host.
+ * @returns {Set<TokenRange>} Ranges for the keyspace on this host
```
jorgebay: We can include that we return `null` when the keyspace is not found in the driver metadata.

lib/token.js
```diff
@@ -0,0 +1,285 @@
...
+   */
+  toBuffer(encoder) {
+    return encoder.encode(this.getValue(), this.getType());
+  }
+
+  toString() {
+    return this._tokenizer.stringify(this.getValue());
+  }
+
+  /**
```
jorgebay: It could be useful to provide an `inspect()` method (used by log providers / console.log / util.inspect()) to have a human-readable string, that returns something like `this.constructor.name + ' { ' + this.toString() + ' }'`.

lib/token.js
```diff
@@ -0,0 +1,285 @@
...
+ * Represents a token on the Cassandra ring.
+ *
+ * @property value The raw value of the token.  Type depends on
+ *                 partitioner implementation. 
+ * 
+ */
+class Token {
+  constructor(value, tokenizer) {
+    this.value = value;
+    Object.defineProperty(this, '_tokenizer', { value: tokenizer, enumerable: false});
```
jorgebay: I think that, instead of `Token#compare()` calling `Tokenizer#compare()` each `Token` implementation should have a `compare()` implementation.
We can remove `Tokenizer#compare()` as it's not being used anymore by `buildTokens()`.

Besides that, I think we should avoid mixing `Token#toString()` with `Tokenizer#stringify()`: `Token#toString()` is a string representation of a token (same as C* tools) and `Tokenizer#stringify()` is a workaround only present on the nodejs driver suited for allowing string hashes of tokens (object properties can only by strings). I propose we implement `Token#toString()` and the base tokenizer uses it default implementation for `stringify()` but each `Tokenizer` implementation can override that behavior to provide the fastest hash for that token (most notably on murmur3).

If we implement those changes, we can get rid of the `Tokenizer` dependency in the `Token` constructor.

lib/tokenizer.js
```diff
@@ -227,7 +268,61 @@ class Murmur3Tokenizer extends Tokenizer {
    * @returns {MutableLong}
```
jorgebay: Now returns a `Murmur3Token`

tolbertam:
Thanks for reviewing!  I'll process all of your feedback today and finish up by adding some integration tests.

lib/token.js
tolbertam: Both great ideas, will do!

lib/token.js
tolbertam: ah yes, that would work great, the current output isn't currently too readable.

tolbertam:
I encountered a bug ([NODEJS-419](https://datastax-oss.atlassian.net/browse/NODEJS-419)) while testing, and also realized my implementation of `getTokenRangesForHost` was completely wrong.   I'm working on some tests that exercise token ranges for all partitioners with a 3 node cluster (w/ and without vnodes) that hopefully I can finish up tomorrow.

lib/metadata/index.js
```diff
@@ -254,12 +284,88 @@ Metadata.prototype.getReplicas = function (keyspaceName, tokenBuffer) {
...
-      keyspace.tokenToReplica(this.tokenizer, this.ringTokensAsStrings, this.primaryReplicas, this.datacenters);
+
+  const replicas = this._getKeyspaceReplicas(keyspace);
+  return replicas[closestToken];
+};
+
+/**
+ * Gets the token ranges that define data distribution in the ring.
+ *
+ * @returns {Set<TokenRange>} The ranges of the ring.
```
tolbertam: After using the API a bit I think it might be confusing that some methods return Array and others Set, should I change these to return Array?

lib/metadata/index.js
```diff
@@ -131,10 +130,26 @@ Metadata.prototype.buildTokens = function (hosts) {
   this.primaryReplicas = primaryReplicas;
   //All the tokens in ring order
   this.ring = allSorted;
+
+  // Build TokenRanges.
+  const tokenRanges = new Set();
```
jorgebay: I would avoid using a `Set` as ES6 doesn't provide a way to override element comparison / hash code, so its not very useful with instances of type object.

lib/metadata/index.js
```diff
@@ -218,20 +233,36 @@ Metadata.prototype._refreshKeyspacesCb = function (waitReconnect, callback) {
...
- * @param {Buffer} tokenBuffer
+ * @param {Buffer|Token|TokenRange} token Can be Buffer (serialized partition key), Token or TokenRange
  * @returns {Array}
  */
-Metadata.prototype.getReplicas = function (keyspaceName, tokenBuffer) {
+Metadata.prototype.getReplicas = function (keyspaceName, token) {
   if (!this.ring) {
     return null;
   }
+  if (util.isBuffer(token)) {
```
jorgebay: Minor nit: `util.isBuffer()` is deprecated, use `Buffer.isBuffer()` instead or `token instanceof Buffer` (no difference really).

lib/metadata/schema-parser.js
```diff
@@ -1027,14 +1027,17 @@ function getTokenToReplicaSimpleMapper(replicationFactor) {
...
         let nextReplicaIndex = i + j;
         if (nextReplicaIndex >= ringLength) {
           //circle back
           nextReplicaIndex = nextReplicaIndex % ringLength;
         }
         const nextReplica = primaryReplicas[ringTokensAsStrings[nextReplicaIndex]];
-        tokenReplicas.push(nextReplica);
+        // In the case of vnodes, consecutive sections of the ring can be assigned to the same host.
+        if (tokenReplicas.indexOf(nextReplica) === -1) {
+          tokenReplicas.push(nextReplica);
```
jorgebay: Nice catch and fix!
We could maybe benefit from using ES6 `Set` here but it's out of scope of this feature. I've created a separate ticket for it to be considered in the future : NODEJS-420.

lib/tokenizer.js
```diff
@@ -40,26 +42,48 @@ class Tokenizer {
...
-    if (val1 > val2) {
-      return 1;
-    }
-    if (val1 < val2) {
-      return -1;
+  split(start, end, numberOfSplits) {
+    throw new Error('You must implement a split function for the tokenizer');
+  }
+
+  splitBase(start, range, ringEnd, ringLength, numberOfSplits) {
```
jorgebay: Can we add jsdoc comments to include the param types (`Integer`) for future reference?

lib/tokenizer.js
jorgebay: Makes sense

lib/metadata/index.js
jorgebay: The more I think about, I think both `Array` / `Set` are valid return types, so it's ok for me to leave it as is.

lib/metadata/index.js
tolbertam: ah, right!  I'll fix that

lib/metadata/index.js
```diff
@@ -104,18 +105,16 @@ Metadata.prototype.buildTokens = function (hosts) {
...
       return;
     }
-    h.tokens.forEach(function (tokenString) {
-      const token = parser(tokenString);
-      utils.insertSorted(allSorted, token, compare);
-      primaryReplicas[stringify(token)] = h;
+    h.tokens.forEach((tokenString) => {
+      const token = this.tokenizer.parse(tokenString);
+      utils.insertSorted(allSorted, token, (t1, t2) => t1.compare(t2));
+      primaryReplicas[stringify(token._value)] = h;
```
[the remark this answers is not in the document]
tolbertam: `getValue()` for murmur 3 returns `Long` where `_value` is `MutableLong`, so I think I still want to use `_value` for that. but I agree we should minimize the usage of that.  Alternatively we could change the `stringify()` implementation to return the token strings instead of the array contents, but that would be slower.

lib/metadata/index.js
tolbertam: 👍 , will stick with `Set` then.  My initial thought was it's useful as a means of communicating that would won't get duplicates (in terms of strict equality (`===`)), however there is a slight cost in that the APIs between arrays and sets are different so it could be annoying for users who want to interact between the two, but that doesn't seem too troubling.

tolbertam:
Thanks for reviewing, I think I've processed all feedback but let me know if I missed anything.

jorgebay:
I've pushed a commit with small jsdoc fixes and a change to return the raw value on `Murmur3Token#getValue()`, feel free to discard those changes if u prefer.

test/unit/token-tests.js
```diff
@@ -0,0 +1,322 @@
...
+        [ hex('b'), hex('c') ],
+        [ hex('c'), hex('d') ]
+      ]));
+      // ring start should be 0x64 (d), end should be 0x61 (a) spanned over a length of 253 (2^8 - 3 byte distance between d and a)
+      // Each range should be 256 / 3 = 84 tokens, with 1 remainder assigned to first range.
+      // Expect 3 ranges, 85, 84, and 84.
+      // 100 (0x64) + 85 == 185 (0xB9)  : ]64,B9]
+      // 0xFF - 0xB9 = 70, this leaves 14 tokens including 0, which gives us 13 (0xD): ]B9,0D]
+      // 0x61 - 0x0D == 84 tokens. ]0D,61]
+      it('should split range that wraps around the ring', tester.rangeSplitTester(hex('d'), hex('a'), 3, [
```
landim: Really liked how the tests are clear and documented. +1

lib/token.js
```diff
@@ -0,0 +1,282 @@
...
+  }
+
+  /**
+   * Encodes the underlying value into a buffer that can be
+   * used for routing.
+   * 
+   * @param {Encoder} encoder Encode for encoding to buffer.
+   * @returns {Buffer} buffer data for the token.
+   */
+  toBuffer(encoder) {
```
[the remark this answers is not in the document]
tolbertam: Yeah the initial idea was to provide it as a convenience function for those looking to convert Tokens into Buffers so they may be fed into `getReplicas`, but since then I've allowed `getReplicas` to take `Token` as an input, so it's no longer needed, so I think we can remove it (will do).

tolbertam:
> I've pushed a commit with small jsdoc fixes and a change to return the raw value on Murmur3Token#getValue(), feel free to discard those changes if u prefer.

Thanks for cleaning everything up.   My initial concern about that was if a user wanted to get at `getValue()`, it would return a `MutableLong`, but now that I think about it, it's not really a problem as I don't expect users to act on `getValue` anyways.   Now that `getValue` simply returns the `_value` property, what do you think about dropping `getValue` and renaming `_value` to `value`?

I'll do some quick testing and drop `toBuffer`, will push it up shortly.

jorgebay:
> what do you think about dropping `getValue` and renaming `_value` to `value`?

I think `getValue()` is ok, I don't have a strong opinion either way.

tolbertam:
Ah, I now remember why someone would access the token value,  for using it in CQL queries like `token > ?`.  It looks like what you did in encoder would account for that.

> I think getValue() is ok, I don't have a strong opinion either way.

Cool, i'll leave that as is then. I pushed up the change to drop `toBuffer` 👍

jorgebay:
lgtm

tolbertam:
> lgtm

Awesome thanks, will squash and merge after tests complete :tada:

tolbertam:
One appveyor run failed because tracing was present after 5 attempts and a CCM startup failure, considering ok to merge.

PR #263, opened 2018-01-11
Refactor routing key generation and set routing key for batches

jorgebay:
- Use routing key from first statement for Batches
- Use create methods for BATCH, EXECUTE and QUERY requests
- Add integration test with routing indexes

Its a 400 loc refactoring, it's hard to review on a diff so it's probably better to see the end result instead.

### DIFF
--- a/test/unit/client-tests.js
+++ b/test/unit/client-tests.js
@@ -145,41 +145,60 @@ describe('Client', function () {
       });
     });
     it('should build the routingKey based on routingNames', function (done) {
-      const requestHandlerMock = function (request, options) {
-        this._options = options;
+      const requestHandlerMock = {
+        send: function (request, options, client, cb) {
+          assert.ok(options);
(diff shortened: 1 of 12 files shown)


test/unit/client-tests.js
```diff
@@ -674,135 +695,6 @@ describe('Client', function () {
       });
     });
   });
-  describe('#_setQueryOptions()', function () {
```
tolbertam: Aren't these tests still applicable? (Other than the method names changing), or are things adequately covered by other tests?

jorgebay: I think unit test of private methods should be used for functionality that is really hard to test, but this is not the case, unit tests on `test/unit/client-tests.js` and `test/unit/encoder-tests.js`and integration tests on `test/integration/short/load-balancing-tests.js` already cover the functionality.

PR #266, opened 2018-01-29
Use a fixed limit for connection writing queue and handle network stream backpressure

jorgebay:
~~Still a work in progress but most of the logic is in.~~

The request handler is responsible to check for busy connections (above thresholds) and move to the next connection if possible.

The connection can still enqueue above stream id (rare, specially on modern protocol versions), that way it doesn't affect heart beats and control connection queries. The timeout of those enqueued requests is started immediately as they are added to the queue.

### DIFF
--- a/lib/client-options.js
+++ b/lib/client-options.js
@@ -6,6 +6,26 @@ const types = require('./types');
 const utils = require('./utils');
 const errors = require('./errors');
 
+/** Core connections per host for protocol versions 1 and 2 */
+const coreConnectionsPerHostV2 = {
+  [types.distance.local]: 2,
+  [types.distance.remote]: 1,
+  [types.distance.ignored]: 0
+};
+
+/** Core connections per host for protocol version 3 and above */
+const coreConnectionsPerHostV3 = {
+  [types.distance.local]: 1,
+  [types.distance.remote]: 1,
+  [types.distance.ignored]: 0
+};
+
+/** Default maxRequestsPerConnection value for protocol v1 and v2 */
+const maxRequestsPerConnectionV2 = 128;
+
+/** Default maxRequestsPerConnection value for protocol v3+ */
+const maxRequestsPerConnectionV3 = 2048;
+
 /**
  * @returns {ClientOptions}
  */
@@ -33,10 +53,7 @@ function defaultOptions () {
     },
     pooling: {
       heartBeatInterval: 30000,
-      warmup: true,
-      // TODO: Make it protocol version dependent
-      // TODO: Use min value between set by the user and the protocol maximum
-      maxRequestsPerConnection: 2048
+      warmup: true
     },
     socketOptions: {
       connectTimeout: 5000,
@@ -45,7 +62,7 @@ function defaultOptions () {
       keepAliveDelay: 0,
       readTimeout: 12000,
       tcpNoDelay: true,
-      coalescingThreshold: 8000
+      coalescingThreshold: 65536
     },
     authProvider: null,
     maxPrepared: 500,
@@ -279,22 +296,25 @@ function getTimestamp(client, userOptions, defaultValue) {
 }
 
 /**
- * Core connections per host for protocol versions 1 and 2
+ * Sets the default options that depend on the protocol version.
+ * @param {ClientOptions} options
+ * @param {Number} version
  */
-const coreConnectionsPerHostV2 = {};
-coreConnectionsPerHostV2[types.distance.local] = 2;
-coreConnectionsPerHostV2[types.distance.remote] = 1;
-coreConnectionsPerHostV2[types.distance.ignored] = 0;
-/**
- * Core connections per host for protocol version 3 and above
- */
-const coreConnectionsPerHostV3 = {};
-coreConnectionsPerHostV3[types.distance.local] = 1;
-coreConnectionsPerHostV3[types.distance.remote] = 1;
-coreConnectionsPerHostV3[types.distance.ignored] = 0;
+function setProtocolDependentDefaults(options, version) {
+  let coreConnectionsPerHost = coreConnectionsPerHostV3;
+  let maxRequestsPerConnection = maxRequestsPerConnectionV3;
+  if (!types.protocolVersion.uses2BytesStreamIds(version)) {
+    coreConnectionsPerHost = coreConnectionsPerHostV2;
+    maxRequestsPerConnection = maxRequestsPerConnectionV2;
+  }
+  options.pooling = utils.deepExtend({}, { coreConnectionsPerHost, maxRequestsPerConnection }, options.pooling);
+}
 
 exports.extend = extend;
 exports.defaultOptions = defaultOptions;
 exports.coreConnectionsPerHostV2 = coreConnectionsPerHostV2;
 exports.coreConnectionsPerHostV3 = coreConnectionsPerHostV3;
 exports.createQueryOptions = createQueryOptions;
+exports.maxRequestsPerConnectionV2 = maxRequestsPerConnectionV2;
+exports.maxRequestsPerConnectionV3 = maxRequestsPerConnectionV3;
+exports.setProtocolDependentDefaults = setProtocolDependentDefaults;
--- a/lib/client-options.js
+++ b/lib/client-options.js
@@ -35,6 +35,7 @@ function defaultOptions () {
       heartBeatInterval: 30000,
       warmup: true,
       // TODO: Make it protocol version dependent
+      // TODO: Use min value between set by the user and the protocol maximum
       maxRequestsPerConnection: 2048
     },
     socketOptions: {
--- a/lib/client-options.js
+++ b/lib/client-options.js
@@ -33,7 +33,9 @@ function defaultOptions () {
     },
     pooling: {
       heartBeatInterval: 30000,
-      warmup: true
+      warmup: true,
+      // TODO: Make it protocol version dependent
+      maxRequestsPerConnection: 2048
     },
     socketOptions: {
       connectTimeout: 5000,
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -486,6 +486,8 @@ Connection.prototype._idleTimeoutHandler = function () {
     }, this.options.pooling.heartBeatInterval);
     return;
   }
+  // TODO: if there isn't an stream id available, check last read time
+
   this.log('verbose', 'Connection idling, issuing a Request to prevent idle disconnects');
   this.sendingIdleQuery = true;
   this.sendStream(new requests.QueryRequest(idleQuery), utils.emptyObject, function (err) {
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -124,7 +124,7 @@ Connection.prototype.open = function (callback) {
   const self = this;
   this.log('info', 'Connecting to ' + this.address + ':' + this.port);
   if (!this.options.sslOptions) {
-    this.netClient = new net.Socket();
+    this.netClient = new net.Socket({ highWaterMark: this.options.socketOptions.coalescingThreshold });
     this.netClient.connect(this.port, this.address, function connectCallback() {
       self.log('verbose', 'Socket connected to ' + self.address + ':' + self.port);
       self.bindSocketListeners();
@@ -139,6 +139,8 @@ Connection.prototype.open = function (callback) {
       self.bindSocketListeners();
       self.startup(callback);
     });
+    // TLSSocket will validate for values from 512 to 16K (depending on the SSL protocol version)
+    this.netClient.setMaxSendFragment(this.options.socketOptions.coalescingThreshold);
   }
   this.netClient.once('error', function socketError(err) {
     self.errorConnecting(err, false, callback);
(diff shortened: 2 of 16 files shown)


lib/writers.js
```diff
@@ -194,63 +202,56 @@ class WriteQueue extends events.EventEmitter {
...
-          setImmediate(next);
-        });
-      },
-      function loopFinished() {
-        // The queue is now empty
-        self.isRunning = false;
+    this.netClient.cork();
+    let totalLength = 0;
+    let willEmitDrain = false;
+    while (this.queue.length > 0 && totalLength < this.coalescingThreshold && !this.error) {
```
jorgebay: `this.error` check should probably not go here, I'll revisit the logic.

lib/client-options.js
```diff
@@ -33,7 +33,10 @@ function defaultOptions () {
     },
     pooling: {
       heartBeatInterval: 30000,
-      warmup: true
+      warmup: true,
+      // TODO: Make it protocol version dependent
+      // TODO: Use min value between set by the user and the protocol maximum
+      maxRequestsPerConnection: 2048
```
tolbertam: It would be nice if this were set to 1024 for v3+ to match the java driver (unless you have a good reason for it to be 2048)

Another thing the java driver allows you to do is to set max requests per connection by host distance.  To be honest, I don't find that very useful and I don't think we should allow that here.

lib/connection.js
```diff
@@ -407,18 +407,24 @@ Connection.prototype.prepareOnce = function (query, callback) {
 };
 
 /**
- * Uses the frame writer to write into the wire
+ * Queues the operation to be written to the wire and invokes the callback once the response was obtained or with an
+ * error (socket error or Opera
```
tolbertam: unfinished comment here i think

lib/connection.js
```diff
@@ -440,20 +446,15 @@ Connection.prototype._write = function (operation, streamId) {
   const self = this;
   this.writeQueue.push(operation, function writeCallback (err) {
     if (err) {
+      // There was a serialization error or the operation has already timed out or was cancelled
+      self._streamIds.push(streamId);
```
tolbertam: In this case do we know if the write actually happened?  If not it makes sense to return the stream id, otherwise we shouldn't return it unless we get a response.

lib/writers.js
```diff
@@ -194,63 +202,56 @@ class WriteQueue extends events.EventEmitter {
...
-            self.setWriteError(err);
-          }
-          // Allow IO between writes
-          setImmediate(next);
-        });
-      },
-      function loopFinished() {
-        // The queue is now empty
-        self.isRunning = false;
+    this.netClient.cork();
```
tolbertam: This seems like a good idea, especially since we bound what we are writing on `coalescingThreshold`

lib/writers.js
```diff
@@ -194,63 +202,56 @@ class WriteQueue extends events.EventEmitter {
...
       }
-    );
+      totalLength += data.length;
+      // Mark it as written to avoid race conditions
+      writeItem.callback();
+      willEmitDrain = !this.netClient.write(data, err => this.setWriteError(err));
+    }
+    // Flush all data
+    this.netClient.uncork();
+    if (!willEmitDrain) {
```
tolbertam: For my benefit, the idea here is that if successive calls to `this.netClient.write` all return true, we won't get a drain event because the stream is not getting backed up.  In this case we call `run` immediately to continue consuming the write queue.  Otherwise, we wait for the `drain` event (line 176) which is our signal to know that we can continue consuming the write queue and writing.

lib/writers.js
```diff
@@ -194,63 +202,56 @@ class WriteQueue extends events.EventEmitter {
 
   run() {
     if (!this.isRunning) {
-      this.process();
+      this.isRunning = true;
+      // Use nextTick to allow the queue to build up on the current phase
```
tolbertam: This comment was really useful in helping me understand why `nextTick`.  We want to allow executing code to finish to allow queue to build up, instead of invoking right away each time.  Seems like could be introduce a small improvement.

test/integration/short/pool-simulator-tests.js
```diff
@@ -0,0 +1,106 @@
...
+
+const assert = require('assert');
+const simulacron = require('../simulacron');
+const helper = require('../../test-helper');
+const errors = require('../../../lib/errors');
+const types = require('../../../lib/types');
+
+const Client = require('../../../lib/client');
+
+describe('pool', function () {
```
tolbertam: Nice tests!

tolbertam:
Have had this running for about 6 hours now against 3 node C* clusters.  Each cluster has a client running against it that is submitting up to 256 simultaneous requests, with speculative executions enabled with a count of 3 and interval of 100ms.  Given this, you could expect up to 768 (3*256) simultaneous requests for inflight requests, and also an unknown amount of timed out speculative requests that we haven't gotten responses for yet but another request for that execution has completed.

The C* nodes are running on [n1-standard-1 GCE instances](https://cloud.google.com/compute/docs/machine-types) which are much smaller than someone should be running in production (4GB ram) and are being restarted at random intervals.  When a node comes up I do still see a spike of in flight requests as I did in [NODEJS-422](https://datastax-oss.atlassian.net/browse/NODEJS-422), but not nearly as dramatic.  This is as expected, as the permitted number of requests per connection is bound at 2048 now, so we don't see a huge spike of possibly unbounded requests (as requests were queued when running out of stream ids).  The anticipated max for a Host is 16386 (2048 * 8) as we allow up to 8 connections per host.   In the typical case, a user would only have 1 connection per host, so you can think of this as an extreme case because:

* We have a consistent constant load (256 concurrent requests possibly being speculatively executed)
* The C* nodes are under-provisioned (4gb ram)
* We have 8 connections per host

As we seen previously, C* 2.1 was much more sensitive to accepting load as soon as a node came up, where C* 3.0 is not nearly as sensitive, but you do see a spike on a host as it comes up.

## 2.1.19 Cluster In Flight Stats

6hrs_2_1_19

## 3.0.15 Cluster In Flight Stats
6hrs_3_0_15

lib/connection.js
tolbertam: I looked at the code and as far as I can tell the callback is invoked before we actually do the write, so if it were to fail it would be something like an encoding error so the write wouldn't have happened, so makes sense to return the stream id.

jorgebay:
@tolbertam nice data and thanks for the analysis!

lib/connection.js
jorgebay: will fix

lib/writers.js
jorgebay: I've simplified the logic a little bit and renamed the variable to `fitsInStreamBuffer` to avoid negating it.

Check it out and let me know what you think.

The idea is that the queue is marked as "running" when its actually writing (sync) and when its waiting for the `drain` event.

When there are enough items in the queue, the flow is:
`process() -> 'drain' event -> process() -> 'drain' event -> ...`

jorgebay:
ok, it's now ready to be reviewed.

The write queue is changed, it's based mostly on the `highWaterMark` of the socket stream. 

I've spent more time than I should trying to figure out how to change the buffer size for `TLSSocket`  which is 16KB by default, to later understand that is limited by the max size of a TLS record...

Regarding the default value for `maxRequestsPerConnection`, I've chosen `2048` to avoid surprises for the users when upgrading. I was thinking about the defaults for a `Client` instance, with a small 3-node cluster, and 6K max outstanding requests seemed safer than 3K.

I'm more concerned protocol v2 defaults, which has configured 2 connections per host as default , that, calculating again with a 3 node cluster, it will be 768 max outstanding requests (2 connections * 128 * 3 hosts), but hopefully the amount of users for C* 2.0 is not significant and C* 2.0 reached EOL a long time ago.

lib/client-options.js
```diff
@@ -42,7 +62,7 @@ function defaultOptions () {
       keepAliveDelay: 0,
       readTimeout: 12000,
       tcpNoDelay: true,
-      coalescingThreshold: 8000
+      coalescingThreshold: 65536
```
jorgebay: I've changed to 64K to make less system calls by default, specially after KPTI.

tolbertam:
> Regarding the default value for maxRequestsPerConnection, I've chosen 2048 to avoid surprises for the users when upgrading. I was thinking about the defaults for a Client instance, with a small 3-node cluster, and 6K max outstanding requests seemed safer than 3K.

> I'm more concerned protocol v2 defaults, which has configured 2 connections per host as default , that, calculating again with a 3 node cluster, it will be 768 max outstanding requests (2 connections * 128 * 3 hosts), but hopefully the amount of users for C* 2.0 is not significant and C* 2.0 reached EOL a long time ago.

That rings a bell for me now.  The reason why 1024 was chosen for Protocol V3+ in the java driver I believe is based on the pooling settings.  With V2, the default max connections per host was 8, with 128 requests per connection gives 1024 simultaneous requests (8*128).  As V3 defaults to 1 connection per host, 1024 was the value that was landed on to match V2.

I think 2048 is fine, but I think a C* host is only going to handle so many native protocol requests at a time and queue the rest (`native_transport_max_threads` default == 128) although that may change in the future.  So the difference between 1024 and 2048 shouldn't be that substantial, other than maybe throwing `BusyConnectionException` at a higher threshold.

I agree that we shouldn't worry too much about 1.2 & 2.0, they are both EOL'd in Apache C* and DataStax Enterprise, and since this is configurable if it gives users problems they can work around it.  It's better than changing the pooling size in my opinion.

landim:
I am also +1 on not to be too much concerned about <2.0 as they reached EOL, especially if there as workaround for "possible" issues.

lib/connection.js
```diff
@@ -139,6 +139,8 @@ Connection.prototype.open = function (callback) {
       self.bindSocketListeners();
       self.startup(callback);
     });
+    // TLSSocket will validate for values from 512 to 16K (depending on the SSL protocol version)
+    this.netClient.setMaxSendFragment(this.options.socketOptions.coalescingThreshold);
```
tolbertam: It looks like in the case of coalescingThreshold > 16K, the setting just isn't applied?   So this would not be useful unless the coalescingThreshold < 16k.

lib/writers.js
```diff
@@ -194,63 +202,52 @@ class WriteQueue extends events.EventEmitter {
...
+        data = writeItem.operation.request.write(this.encoder, writeItem.operation.streamId);
+      }
+      catch (err) {
+        writeItem.callback(err);
+        continue;
       }
-    );
+      // Mark it as written to avoid race conditions
+      writeItem.callback();
+      fitsInStreamBuffer = this.netClient.write(data, err => this.setWriteError(err));
```
tolbertam: `fitsInStreamBuffer` is a much better name and easier to understand than `willEmitDrain` 👍

lib/writers.js
```diff
@@ -194,63 +202,52 @@ class WriteQueue extends events.EventEmitter {
...
+    }
+    // Flush all data
+    this.netClient.uncork();
+
+    if (fitsInStreamBuffer) {
```
tolbertam: I was initially thinking you would still want to call `process` or `run` here, but I realized its not necessary because if `fitsInStreamBuffer` is true, than the queue is empty or we are in an error state, so we don't need to schedule run (until more are queued).

lib/writers.js
```diff
@@ -194,63 +202,52 @@ class WriteQueue extends events.EventEmitter {
...
+      this.isRunning = false;
+    }
   }
 
   /**
    * Emits the 'error' event and callbacks items that haven't been written and clears them from the queue.
    * @param err
    */
   setWriteError(err) {
+    if (!err) {
```
tolbertam: In what case would err not be set here?  Just wondering if there is a use case or if this is just for protecting against an invalid empty input.

test/integration/short/ssl-tests.js
```diff
@@ -28,6 +29,15 @@ describe('Client', function () {
           helper.finish(client, done)();
         });
       });
+      it('should callback in error when rejecting unauthorized', function (done) {
```
tolbertam: We expect this to fail since C* is using a self-signed certificate with CCM is that right?

lib/connection.js
jorgebay: correct

test/integration/short/ssl-tests.js
jorgebay: yep :)

jorgebay:
After investigating a regression found in the new write queue, I went back to use `socket.write()` callback to throttle writes and manual coalescing, which seems to perform way better than `cork`/`uncork` and `'drain'` event (it caused ~10% lower throughput).

I removed the `whilst()` call and used a `while` loop.

I'm about to finish running the benchmarks.

test/unit/basic-tests.js
```diff
@@ -756,7 +756,7 @@ describe('writers', function () {
           buffers.push(buf);
           totalLength += buf.length;
           if (cb) {
-            setTimeout(cb, 50);
+            setTimeout(cb, 20);
```
landim: Is there a reason to change to 20? I think a code comment would be good to be more clear about this.

jorgebay: I saw a failure on a CI run related to the final timeout on this test and the solution was to use `helper.setIntervalUntil()` to avoid race conditions.
I've reduced the time on this timeout to make it faster, any timeout (not `setImmediate()`) would work.
I could add a comment :)

PR #267, opened 2018-02-05
Destroy socket if close event not received within connect timeout

tolbertam:
For [NODEJS-428](https://datastax-oss.atlassian.net/browse/NODEJS-428).

Rather than `destroy`ing the connection always, attempt to `end` first, and if server hasn't responded within `SocketOptions.connectTimeout` then destroy it.

Tested this out locally and it seems to work good.   Attempted writing a test with this using simulacron but while I can configure it to accept connections and ignore `STARTUP` messages, it will still close the socket when it receives a FIN from the driver.   I think I can write a simple test using `ccm pause` so will push that tomorrow.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -626,7 +626,15 @@ Connection.prototype.close = function (callback) {
   this.isSocketOpen = false;
   this.log('verbose', 'Closing connection to ' + this.endpoint);
   const self = this;
+
+  // If server doesn't acknowledge the half-close within connection timeout, destroy the socket.
+  let endTimeout = setTimeout(() => {
+    self.log('info', this.endpoint + ' did not respond to connection close within ' + this.options.socketOptions.connectTimeout + 'ms, destroying connection');
+    this.netClient.destroy();
+  }, this.options.socketOptions.connectTimeout);
+
   this.netClient.once('close', function (hadError) {
+    clearTimeout(endTimeout);
     if (hadError) {
       self.log('info', 'The socket closed with a transmission error');
     }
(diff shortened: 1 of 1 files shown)


lib/connection.js
```diff
@@ -626,7 +626,15 @@ Connection.prototype.close = function (callback) {
   this.isSocketOpen = false;
   this.log('verbose', 'Closing connection to ' + this.endpoint);
   const self = this;
+
+  // If server doesn't acknowledge the half-close within connection timeout, destroy the socket.
+  const endTimeout = setTimeout(() => {
```
jorgebay: `callback` should be invoked on the timeout handler, that way we always invoke it.

lib/connection.js
```diff
@@ -626,7 +626,15 @@ Connection.prototype.close = function (callback) {
...
+    self.log('info', this.endpoint + ' did not respond to connection close within ' + this.options.socketOptions.connectTimeout + 'ms, destroying connection');
```
jorgebay: Minor: mix use of `self` and `this`.

tolbertam: D'oh!  That's a bad habit of mine, will fix :)

lib/connection.js
tolbertam: `netClient.destroy` causes the `close` event to happen which triggers `callback` so it shouldn't be needed here unless I am mistaken.

tolbertam:
Thanks for reviewing!

> I think it would be very hard to cover from an integration tests. I'm comfortable with the existent coverage we get from endurance test runs.

Sounds good.  It's pretty easy for me to validate manually, so comfortable not having a specific test for this.

lib/connection.js
jorgebay: ah, I learned something :), sounds good.

jorgebay:
Sorry, I've forgot my review was in "requested changes" state.

PR #268, opened 2018-02-14
Add additional validation in Encoder.setRoutingKeyFromUser

tolbertam:
For [NODEJS-429](https://datastax-oss.atlassian.net/browse/NODEJS-429)

Only Array<Buffer>, Buffer, Token, and TokenRange values should be
acceptable for options.routingKey.  Throws TypeError when unexpected
type provided.

### DIFF
--- a/test/integration/short/load-balancing-tests.js
+++ b/test/integration/short/load-balancing-tests.js
@@ -145,10 +145,10 @@ context('with a reusable 3 node cluster', function () {
       }, helper.finish(client, done));
     });
     it('should target the correct replica using user-provided Buffer routingKey', function (done) {
-      testWithQueryOptions((client) => {
-        // Use [0] which should map to host 1.
-        return { routingKey: Buffer.from([0]) };
-      }, '1', done);
+      // Use [0] which should map to node 1
+      testWithQueryOptions((client) => ({
+        routingKey: Buffer.from([0])
+      }), '1', done);
     });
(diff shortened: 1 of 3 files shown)


test/integration/short/load-balancing-tests.js
```diff
@@ -144,9 +144,69 @@ context('with a reusable 3 node cluster', function () {
...
+      }, '3', done);
+    });
+    it('should throw TypeError if invalid routingKey type is provided', function () {
+      const client = new Client({
+        policies: { loadBalancing: new TokenAwarePolicy(new RoundRobinPolicy()) },
+        keyspace: 'ks_network_rp1',
+        contactPoints: helper.baseOptions.contactPoints
+      });
+      const query = 'select * from system.local';
+      return client.execute(query, [], { routingKey: 'this is not valid' })
```
jorgebay: I think we need to use the callback-based interface here, as using promises will catch errors that are unexpectedly thrown along with rejected promises.

We should test that `client.execute()` will not sync throw an error
```
// client.execute() call should not throw
client.execute(query, [], { routingKey: 'this is not valid' }, err => {
  helper.assertInstanceOf(err, TypeError);
  done();
});
```

jorgebay:
I've just validated that `client.execute()` call in "should throw TypeError if invalid routingKey type is provided" would throw a sync error (unexpected error) without the fix.

Lgtm +1

PR #270, opened 2018-03-02
Add NO_COMPACT startup option

tolbertam:
Need to investigate if there is a way to test this at integration level as would be nice to have a test for it if we could.  I documented a process for testing this manually in datastax/java-driver#915 which I verified works with these changes.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -98,6 +98,16 @@ const warmupLimit = 32;
  * @property {Number} protocolOptions.maxVersion When set, it limits the maximum protocol version used to connect to
  * the nodes.
  * Useful for using the driver against a cluster that contains nodes with different major/minor versions of Cassandra.
+ * @property {Boolean} protocolOptions.noCompact When set to true, enables the NO_COMPACT startup option.
+ * <p>
+ * When this option is supplied <code>SELECT</code>, <code>UPDATE</code>, <code>DELETE</code>, and <code>BATCH</code>
+ * statements on <code>COMPACT STORAGE</code> tables function in "compatibility" mode which allows seeing these tables
+ * as if they were "regular" CQL tables.
+ * </p>
+ * <p>
+ * This option only effects interactions with interactions with tables using <code>COMPACT STORAGE</code> and is only
+ * support by C* 4.0+ and DSE 6.0+.
+ * </p>
  * @property {Object} socketOptions
  * @property {Number} socketOptions.connectTimeout Connection timeout in milliseconds. Default: 5000.
  * @property {Number} socketOptions.defunctReadTimeoutThreshold Determines the amount of requests that simultaneously
--- a/lib/client.js
+++ b/lib/client.js
@@ -106,7 +106,7 @@ const warmupLimit = 32;
  * </p>
  * <p>
  * This option only effects interactions with interactions with tables using <code>COMPACT STORAGE</code> and is only
- * support by C* 4.0+ and DSE 6.0+.
+ * supported by C* 3.0.16+, 3.11.2+, 4.0+ and DSE 6.0+.
  * </p>
  * @property {Object} socketOptions
  * @property {Number} socketOptions.connectTimeout Connection timeout in milliseconds. Default: 5000.
--- a/lib/requests.js
+++ b/lib/requests.js
@@ -216,16 +216,21 @@ class PrepareRequest extends Request {
   }
 }
 class StartupRequest extends Request {
-  constructor(cqlVersion) {
+  constructor(cqlVersion, noCompact) {
     super();
     this.cqlVersion = cqlVersion || '3.0.0';
+    this.noCompact = noCompact || null;
   }
 
   write(encoder, streamId) {
     const frameWriter = new FrameWriter(types.opcodes.startup);
-    frameWriter.writeStringMap({
+    const startupOptions = {
       CQL_VERSION: this.cqlVersion
-    });
+    };
+    if(this.noCompact) {
+      startupOptions['NO_COMPACT'] = 'true';
+    }
+    frameWriter.writeStringMap(startupOptions);
     return frameWriter.write(encoder.protocolVersion, streamId);
   }
 }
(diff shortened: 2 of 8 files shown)


lib/client.js
```diff
@@ -98,6 +98,16 @@ const warmupLimit = 32;
...
  * Useful for using the driver against a cluster that contains nodes with different major/minor versions of Cassandra.
+ * @property {Boolean} protocolOptions.noCompact When set to true, enables the NO_COMPACT startup option.
+ * <p>
+ * When this option is supplied <code>SELECT</code>, <code>UPDATE</code>, <code>DELETE</code>, and <code>BATCH</code>
+ * statements on <code>COMPACT STORAGE</code> tables function in "compatibility" mode which allows seeing these tables
+ * as if they were "regular" CQL tables.
+ * </p>
+ * <p>
+ * This option only effects interactions with interactions with tables using <code>COMPACT STORAGE</code> and is only
+ * support by C* 4.0+ and DSE 6.0+.
```
tolbertam: Actually, the option was added in 3.0.16 & 3.11.2 as well, so will update this comment.

tolbertam:
Without noCompact, `column1` / `value` are missing:

```
... (ccm) atolbert-rmbp:nodejs-driver andrew.tolbert$ node
> const cassandra = require('./index.js');
undefined
> const client = new cassandra.Client({ contactPoints: ['127.0.0.1'] });
undefined
> client.execute('select * from simple.foo where key=0xc0').then(result => console.log(result.rows[0]));
Promise {
  <pending>,
  domain: 
   Domain {
     domain: null,
     _events: 
      { removeListener: [Function: updateExceptionCapture],
        newListener: [Function: updateExceptionCapture],
        error: [Function: debugDomainError] },
     _eventsCount: 3,
     _maxListeners: undefined,
     members: [] } }
> Row {
  key: <Buffer c0>,
  bar: 10,
  baz: Uuid: 33cb65d4-6721-4ca8-854f-1f020c5353cb }
```

with noCompact = true, the columns are present:

```
> (ccm) atolbert-rmbp:nodejs-driver andrew.tolbert$ node
> const cassandra = require('./index.js');
undefined
> const client = new cassandra.Client({ contactPoints: ['127.0.0.1'], protocolOptions: { noCompact: true } });
undefined
> client.execute('select * from simple.foo where key=0xc0').then(result => console.log(result.rows[0]));
Promise {
  <pending>,
  domain: 
   Domain {
     domain: null,
     _events: 
      { removeListener: [Function: updateExceptionCapture],
        newListener: [Function: updateExceptionCapture],
        error: [Function: debugDomainError] },
     _eventsCount: 3,
     _maxListeners: undefined,
     members: [] } }
> Row {
  key: <Buffer c0>,
  column1: 'yak',
  bar: 10,
  baz: Uuid: 33cb65d4-6721-4ca8-854f-1f020c5353cb,
  value: 'cafedead' }
```

tolbertam:
After experimenting, there is a way to test this at integration level (see [here](https://github.com/datastax/java-driver/pull/915#issuecomment-370067152)).  Will work on that.

tolbertam:
Added tests and removed wip label.

test/integration/short/client-execute-tests.js
```diff
@@ -539,6 +539,55 @@ describe('Client', function () {
         done();
       });
     });
+    vdescribe('3.0.16', 'with noCompact', function () {
+      before(function (done) {
+        // C* 4.0 does not support compact storage
+        if (helper.isCassandraGreaterThan('4.0')) {
+          this.skip();
```
jorgebay: I think you should exit after skipping.

lib/client.js
jorgebay: Missing change :)

test/integration/short/client-execute-tests.js
```diff
@@ -539,6 +539,55 @@ describe('Client', function () {
...
+        }
+        const client = newInstance({keyspace: keyspace, protocolOptions: { noCompact: true }});
+        utils.series([
+          client.connect.bind(client),
+          helper.toTask(client.execute, client, 'CREATE TABLE tbl_cs (key blob PRIMARY KEY, bar int, baz uuid) WITH COMPACT STORAGE'),
+          helper.toTask(client.execute, client, "INSERT INTO tbl_cs (key, bar, baz, column1, value) values (0xc0, 10, 33cb65d4-6721-4ca8-854f-1f020c5353cb, 'yak', 0xcafedead)"),
+        ], done);
+      });
+      it('set to true should reveal non-schema columns', () => {
```
jorgebay: Its awesome that you found a way to cover it using integration tests (I will use the same for C# driver as well, <insert_burns_meme_here> mwahahaha).

lib/requests.js
```diff
@@ -216,16 +216,21 @@ class PrepareRequest extends Request {
   }
 }
 class StartupRequest extends Request {
-  constructor(cqlVersion) {
+  constructor(cqlVersion, noCompact) {
     super();
     this.cqlVersion = cqlVersion || '3.0.0';
+    this.noCompact = noCompact || null;
```
jorgebay: nit: `|| null` portion can be removed.
If `this.noCompact` is `truthy`, you insert `'NO_COMPACT': 'true'` which will work either way as you don't use the actual value.

test/integration/short/client-execute-tests.js
jorgebay: I've looked it up in the doc:

> Best practice: To avoid confusion, do not execute further instructions in a test or hook after calling `this.skip()`.

It probably doesn't work throwing an error as other test frameworks, but I haven't looked it up.

lib/client.js
tolbertam: D'oh!  I'll fix that :)

test/integration/short/client-execute-tests.js
tolbertam: ah!  TIL.  I'll add returns where applicable 👍

jorgebay:
Changes lgtm! 👍

PR #273, opened 2018-03-08
Add jsdoc for each response error code

tolbertam:
For [NODEJS-418](https://datastax-oss.atlassian.net/browse/NODEJS-418).

### DIFF
--- a/lib/types/index.js
+++ b/lib/types/index.js
@@ -234,6 +234,25 @@ const protocolEvents = {
 
 /**
  * Server error codes returned by Cassandra
+ * @type {Object}
+ * @property {Number} serverError Something unexpected happened.
+ * @property {Number} protocolError Some client message triggered a protocol violation.
+ * @property {Number} badCredentials Authentication was required and failed.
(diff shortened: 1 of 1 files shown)


lib/types/index.js
```diff
@@ -234,6 +234,25 @@ const protocolEvents = {
 
 /**
  * Server error codes returned by Cassandra
+ * @type {Object}
+ * @property {Number} serverError Something unexpected happened.
+ * @property {Number} protocolError Some client message triggered a protocol violation.
+ * @property {Number} badCredentials Authentication was required and failed.
+ * @property {Number} unavailableException Raised when coordinator knows there is not enough replicas alive to perform a query with the requested consistency level.
```
tolbertam: Wasn't sure whether it was appropriate for me to also document the properties (beyond `message`) that would come along with errors of each type.  i.e. for `unavailableException`, we would expect `consistencies`, `required` and `alive`).  wdyt?

jorgebay: I see what you mean... I think this makes a good case for using dedicated error prototypes. Just to be able to document them, we could still provide a error code (for conditions, mappings, etc). Maybe we can do that on a separate patch.

PR #274, opened 2018-03-08
Log driver version on Client.connect

tolbertam:
For [NODEJS-426](https://datastax-oss.atlassian.net/browse/NODEJS-426).  Since `Client` emits log events, it made the most sense to emit a log message on connect instead of `Client` instantiation, as in that case there would be no opportunity for user to register a log event.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -9,6 +9,8 @@ const ProfileManager = require('./execution-profile').ProfileManager;
 const requests = require('./requests');
 const clientOptions = require('./client-options');
 const ClientState = require('./metadata/client-state');
+const description = require('../package.json').description;
+const version = require('../package.json').version;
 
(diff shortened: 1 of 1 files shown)


lib/client.js
```diff
@@ -401,6 +403,7 @@ Client.prototype._connectCb = function (callback) {
   }
   this.connecting = true;
   const self = this;
+  this.log('info', util.format("Connecting to cluster using '%s' version %s", description, version));
```
tolbertam: Will log something like:

> Connecting to cluster using 'DataStax Node.js Driver for Apache Cassandra' version 3.4.1


I surrounded the package description in quotes to remove ambiguity, as without them it's possible that one may interpret the version as the Apache Cassandra version.

PR #275, opened 2018-03-09
Reject promise for schema metadata methods when not connected

tolbertam:
For [NODEJS-412](https://datastax-oss.atlassian.net/browse/NODEJS-412).

Added an `initialized` flag to `Metadata` that gets set upon the first completion of `refeshKeyspaces` (which is done as part of connection intiailization).  

This flag is only applicable if `isMetadataSyncEnabled` is true.  

In the case when a user calls `getUdt`, `getTable`, `getFunctions`, `getFunction`, `getAggregates` or `getMaterializedView` and `isMetadataSyncEnabled` is true.  The callback should be invoked in error immediately.  If `isMetadataSyncEnabled` is false, it continues down an existing path, which will raise an Error indicating that a connection cannot be acquired, i.e.:

```
{ OperationTimedOutError: A connection could not be acquired before timeout.
    at Timeout.waitTimeout [as _onTimeout] (/Users/andrew.tolbert/Documents/Projects/nodejs-driver/lib/control-connection.js:708:14)
    at ontimeout (timers.js:458:11)
    at tryOnTimeout (timers.js:296:5)
    at Timer.listOnTimeout (timers.js:259:5)
  name: 'OperationTimedOutError',
  info: 'Represents a client-side error that is raised when the client did not hear back from the server within socketOptions.readTimeout',
  message: 'A connection could not be acquired before timeout.' }
```

Wasn't sure if that behavior was ok or not.  We could optimize that by failing right away in a similar fashion.

Other metadata methods I didn't change:

* `newToken()` and `newTokenRange()` : already preemptively fail if `this.tokenizer` isn't set.
* `getReplicas()`, `getTokenRanges()`, `getTokenRangesForHost()`, currently return null if never have connected.  Should I change this?   It's not asynchronous so wasn't sure.
* `getTrace()`, `refreshKeyspaces()` attempts to acquire connection which will fail.  Should it fail earlier?

### DIFF
--- a/lib/metadata/index.js
+++ b/lib/metadata/index.js
@@ -177,6 +177,9 @@ Metadata.prototype.refreshKeyspace = function (name, callback) {
  * @private
  */
 Metadata.prototype._refreshKeyspaceCb = function (name, callback) {
+  if (!this.initialized) {
+    return callback(this._uninitializedError(), null);
+  }
   this.log('info', util.format('Retrieving keyspace %s metadata', name));
   const self = this;
   this._schemaParser.getKeyspace(name, function (err, ksInfo) {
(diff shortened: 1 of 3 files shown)


jorgebay:
> `getReplicas()`, `getTokenRanges()`, `getTokenRangesForHost()`, currently return `null`

We can include a jsdoc mentioning the expected behaviour when not connected.

jorgebay:
> `getTrace()`, `refreshKeyspaces()` attempts to acquire connection which will fail. Should it fail earlier?

I think we have to prevent that and fail immediately.

lib/metadata/index.js
```diff
@@ -434,7 +448,7 @@ Metadata.prototype.getUdt = function (keyspaceName, name, callback) {
 Metadata.prototype._getUdtCb = function (keyspaceName, name, callback) {
   let cache;
   if (this.options.isMetadataSyncEnabled) {
-    const keyspace = this.keyspaces[keyspaceName];
+    const keyspace = this._getKeyspaceFromCache(keyspaceName);
```
jorgebay: We can't throw an error on an callback-based method, we should manually check, invoke the callback and exit.

tolbertam: D'oh, good point!  Don't know why that didn't occur to me at the time.  Also erroring out the callback should also reject the promise, so still solves that for me.  I'll fix it.

tolbertam:
> I think we have to prevent that and fail immediately.

Sounds good.  The schema methods also currently behave the same if `isMetadataSyncEnabled` is false, so will make sure those fail immediately too.

tolbertam:
Thanks for reviewing!

> We could include a test making sure that, when using callbacks, the error is defined as a parameter (and not thrown).

Will do and then merge 👍

PR #279, opened 2018-05-15
Add node.js 10 to CI matrices and update tests to account for changes in node.js 10

tolbertam:
Adds node.js 10 to our test matrices.

Also makes a number of test changes:

- Use 192.0.2.1 in EC2MultiRegionTranslator test.  `127.100.100.1` resolves by default in Ubuntu 17.04+ so using a reserved address seemed more reliable.
- Fixed assert.throws use as nodejs 10 will now fail when using `string` for validating expectation.
- Account for node.js 10 change to defer `readable` events to `nextTick`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: node_js
 node_js:
+  - "10"
   - "8"
   - "6"
   - "4"
--- a/build.yaml
+++ b/build.yaml
@@ -6,8 +6,8 @@ schedules:
       exclude:
         # Exclude all builds for 6.x.
         - nodejs: ['6']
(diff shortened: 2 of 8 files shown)


jorgebay:
I imagine finding the underlying issue for stream-based test was a tough one :) https://github.com/nodejs/node/pull/17979

Great fixes! +1

tolbertam:
Indeed it was!  I ultimately found the notice about it in the [release notes](https://nodejs.org/en/blog/release/v10.0.0/):

* The 'readable' event is now always deferred with nextTick. [1e0f3315c7]

But it was definitely puzzling at first 😄

PR #280, opened 2018-07-16
Introduce RequestTracker interface and RequestLogger implementation

jorgebay:
Fixes NODEJS-366 and NODEJS-220.

### DIFF
--- a/lib/tracker/request-logger.js
+++ b/lib/tracker/request-logger.js
@@ -1,21 +1,92 @@
 'use strict';
 
+const events = require('events');
 const RequestTracker = require('./request-tracker');
+const errors = require('../errors');
+const format = require('util').format;
+
+const nanosToMillis = 1000000;
+const defaultMessageMaxQueryLength = 500;
+const defaultMaxParameterValueLength = 50;
+const defaultMaxErrorStackTraceLength = 200;
 
 /**
  * A request tracker that logs the requests executed through the session, according to a set of
  * configurable options.
  * @implements {module:tracker~RequestTracker}
  * @alias module:tracker~RequestLogger
+ * @example <caption>Logging slow queries</caption>
+ * const requestLogger = new RequestLogger({ slowThreshold: 1000 });
+ * requestLogger.emitter.on('show', message => console.log(message));
+ * // Add the requestLogger to the client options
+ * const client = new Client({ contactPoints, requestTracker: requestLogger });
  */
 class RequestLogger extends RequestTracker {
+
+  /**
+   * Creates a new instance of {@link RequestLogger}.
+   * @param {Object} options
+   * @param {Number} [options.slowThreshold] The threshold in milliseconds beyond which queries are considered 'slow'
+   * and logged as such by the driver.
+   * @param {Number} [options.requestSizeThreshold] The threshold in bytes beyond which requests are considered 'large'
+   * and logged as such by the driver.
+   * @param {Boolean} [options.logNormalRequests] Determines whether it should emit 'normal' events for every
+   * EXECUTE, QUERY and BATCH request executed successfully, useful only for debugging. This option can be modified
+   * after the client is connected using the property {@link RequestLogger#logNormalRequests}.
+   * @param {Boolean} [options.logErroredRequests] Determines whether it should emit 'failure' events for every
+   * EXECUTE, QUERY and BATCH request execution that resulted in an error. This option can be modified
+   * after the client is connected using the property {@link RequestLogger#logErroredRequests}.
+   * @param {Number} [options.messageMaxQueryLength] The maximum amount of characters that are logged from the query
+   * portion of the message. Defaults to 500.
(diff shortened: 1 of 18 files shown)


tolbertam:
Awesome!  Should hopefully get some time to look at this later this week.

lib/tracker/request-logger.js
```diff
@@ -0,0 +1,281 @@
...
+   * @param {Boolean} [options.logNormalRequests] Determines whether it should emit 'normal' events for every
+   * EXECUTE, QUERY and BATCH request executed successfully, useful only for debugging. This option can be modified
+   * after the client is connected using the property {@link RequestLogger#logNormalRequests}.
+   * @param {Boolean} [options.logErroredRequests] Determines whether it should emit 'failure' events for every
+   * EXECUTE, QUERY and BATCH request execution that resulted in an error. This option can be modified
+   * after the client is connected using the property {@link RequestLogger#logErroredRequests}.
+   * @param {Number} [options.messageMaxQueryLength] The maximum amount of characters that are logged from the query
+   * portion of the message. Defaults to 500.
+   * @param {Number} [options.messageMaxParameterValueLength] The maximum amount of characters of each query parameter
+   * value that will be included in the message.. Defaults to 50.
```
tolbertam: Very small nit, there are two periods here.

lib/tracker/request-logger.js
```diff
@@ -0,0 +1,281 @@
...
+    /**
+     * The object instance that emits <code>'slow'</code>, <code>'large'</code>, <code>'normal'</code> and
+     * <code>'failure'</code> events.
+     * @type {EventEmitter}
+     */
+    this.emitter = new events.EventEmitter();
+  }
+
+  /**
+   * Logs a message when a request execution was deemed slow or too large.
```
tolbertam: Shouldn't these comments be:

`onSuccess`:  Logs message if request execution was deemed too slow, large or if normal requests are logged.
`OnError:`  Logs message if request execution was too large and/or encountered an error.

lib/tracker/request-logger.js
```diff
@@ -0,0 +1,281 @@
...
+    if (this._options.requestSizeThreshold > 0 && requestLength > this._options.requestSizeThreshold) {
+      this._logLargeErrorRequest(host, query, parameters, queryOptions, requestLength, err, latency);
+    }
+    else if (this.logErroredRequests) {
+      this._logErrorRequest(host, query, parameters, queryOptions, requestLength, err, latency);
+    }
+  }
+
+  _logSlow(host, query, parameters, queryOptions, requestLength, responseLength, latency) {
+    const message = format('[%s] Slow request, took %d ms (%s): %s', host.address, ~~toMillis(latency),
```
tolbertam: Wasn't initially clear to me that what bitwise not operation `~~` was used for, is to round down effectively right?

lib/tracker/request-logger.js
```diff
@@ -0,0 +1,281 @@
+'use strict';
```
tolbertam: Really dig the design of `RequestLogger` and `RequestTracker`, nice and clean, good documentation and easy to follow :+1:

lib/tracker/request-tracker.js
```diff
@@ -0,0 +1,58 @@
...
+   * @param {Array|Object|null} parameters In the case of prepared or unprepared query executions, the provided
+   * parameters.
+   * @param {QueryOptions} queryOptions The query options used to execute the request.
+   * @param {Number} requestLength Length of the body of the request.
+   * @param {Number} responseLength Length of the body of the response.
+   * @param {Array<Number>} latency An array containing [seconds, nanoseconds] tuple, where nanoseconds is the
+   * remaining part of the real time that can't be represented in second precision (see <code>process.hrtime()</code>).
+   */
+  onSuccess(host, query, parameters, queryOptions, requestLength, responseLength, latency) {
+    throw new Error('Not implemented');
```
tolbertam: This is a small thing, but what do you think about having these do nothing instead of throwing errors?  That way implementations only need to  implement one of `onSuccess` / `onError` .

test/integration/short/tracker-simulator-tests.js
```diff
@@ -0,0 +1,249 @@
+'use strict';
```
tolbertam: Awesome integration and unit tests :+1:

lib/tracker/request-logger.js
jorgebay: Current message makes no sense... I'll fix it!

lib/tracker/request-logger.js
jorgebay: 👍 `Math.floor()` is more clear, will change.

lib/tracker/request-tracker.js
jorgebay: I think it makes more sense to avoid throwing an error as its likely that a user would like to implement just 1 of them ✔️

PR #281, opened 2018-07-19
Add 'host' query option to enable host targeting

tolbertam:
For [NODEJS-443](https://datastax-oss.atlassian.net/browse/NODEJS-443).

Adds a new 'host' query option for host targeting, i.e.:

```javascript
const host = client.hosts.get('127.0.0.2:9042');
client.execute('select rpc_address from system.local', [], { host: host })
  .then((result) => console.log('Queried host:', result.info.queriedHost, '\nRows:', result.rows));
```

emits:

    Queried host: 127.0.0.2:9042
    Rows: [ Row { rpc_address: InetAddress: 127.0.0.2 } ]

### DIFF
--- a/lib/client-options.js
+++ b/lib/client-options.js
@@ -222,6 +222,10 @@ function createQueryOptions(client, userOptions, rowCallback, logged) {
   userOptions = (!userOptions || typeof userOptions === 'function') ? utils.emptyObject : userOptions;
   const defaultQueryOptions = client.options.queryOptions;
 
+  if (userOptions.host !== null && !(userOptions.host instanceof Host)) {
+    throw new TypeError('host must be an instance of Host');
+  }
+
   // Using fixed property names is 2 order of magnitude faster than dynamically shallow clone objects
   const result = {
     autoPage: ifUndefined(userOptions.autoPage, defaultQueryOptions.autoPage),
@@ -320,3 +324,5 @@ exports.createQueryOptions = createQueryOptions;
 exports.maxRequestsPerConnectionV2 = maxRequestsPerConnectionV2;
 exports.maxRequestsPerConnectionV3 = maxRequestsPerConnectionV3;
 exports.setProtocolDependentDefaults = setProtocolDependentDefaults;
+
+const Host = require('./host').Host;
--- a/lib/client-options.js
+++ b/lib/client-options.js
@@ -231,6 +231,7 @@ function createQueryOptions(client, userOptions, rowCallback, logged) {
     executionProfile: profile,
     fetchSize: ifUndefined(userOptions.fetchSize, defaultQueryOptions.fetchSize),
     hints: userOptions.hints,
+    host: userOptions.host,
     isIdempotent: ifUndefined(userOptions.isIdempotent, defaultQueryOptions.isIdempotent),
     keyspace: userOptions.keyspace,
     logged: ifUndefined(userOptions.logged, logged),
--- a/lib/client-options.js
+++ b/lib/client-options.js
@@ -222,7 +222,7 @@ function createQueryOptions(client, userOptions, rowCallback, logged) {
   userOptions = (!userOptions || typeof userOptions === 'function') ? utils.emptyObject : userOptions;
   const defaultQueryOptions = client.options.queryOptions;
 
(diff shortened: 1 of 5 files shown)


lib/client-options.js
```diff
@@ -318,4 +323,6 @@ exports.coreConnectionsPerHostV3 = coreConnectionsPerHostV3;
 exports.createQueryOptions = createQueryOptions;
 exports.maxRequestsPerConnectionV2 = maxRequestsPerConnectionV2;
 exports.maxRequestsPerConnectionV3 = maxRequestsPerConnectionV3;
-exports.setProtocolDependentDefaults = setProtocolDependentDefaults;
\ No newline at end of file
+exports.setProtocolDependentDefaults = setProtocolDependentDefaults;
+
+const Host = require('./host').Host;
```
tolbertam: delayed require until after `exports.defaultOptions` otherwise we create cyclic dependency with `host` -> `host-connection-pool` -> `defaultOptions`

lib/client-options.js
```diff
@@ -222,6 +222,10 @@ function createQueryOptions(client, userOptions, rowCallback, logged) {
   userOptions = (!userOptions || typeof userOptions === 'function') ? utils.emptyObject : userOptions;
   const defaultQueryOptions = client.options.queryOptions;
 
+  if (userOptions.host !== null && !(userOptions.host instanceof Host)) {
```
tolbertam: Wasn't sure if this validation is necessary or not.

lib/request-handler.js
```diff
@@ -153,15 +155,31 @@ class RequestHandler {
...
-      if (err) {
-        return callback(err);
-      }
-      self._hostIterator = iterator;
-      self._callback = callback;
+    if (this.requestOptions.host) {
+      // if host is configured bypass load balancing policy and use
+      // a single host plan.
+      let hostProvided = false; 
+      self._hostIterator = {
```
jorgebay: We can reuse existing `utils.arrayIterator([ host ])` or you can write a small generator function (Node.js 4 supports it):

```javascript
function* singleHostIterator(h) {
  yield h;
}

self._hostIterator = singleHostIterator(host);
```

test/integration/short/client-execute-simulator-tests.js
```diff
@@ -0,0 +1,92 @@
...
+      // since node 3 is not included in our LBP, the request should fail as we have no
+      // connectivity to that node.
+      const node = cluster.node(3);
+      const host = client.hosts.get(node.address);
+      return client.execute(query, [], { host: host })
+        .catch((err) => {
+          helper.assertInstanceOf(err, errors.NoHostAvailableError);
+          // no hosts should have been attempted.
+          assert.deepEqual(Object.keys(err.innerErrors), []);
+        });
```
jorgebay: We should also verify that `catch()` was executed, e.g.: `p.catch(...).then(() => assert.ok(catchedError))`

test/integration/short/client-execute-simulator-tests.js
```diff
@@ -0,0 +1,92 @@
...
+          alive: 0,
+          required: 1,
+          consistency_level: 'LOCAL_ONE'
+        }
+      }, () => {
+        client.execute(query, [], { host: host }, (err, result) => {
+          assert.ok(err);
+          helper.assertInstanceOf(err, errors.NoHostAvailableError);
+          assert.deepEqual(Object.keys(err.innerErrors), [node.address]);
+          const nodeError = err.innerErrors[node.address];
```
jorgebay: nit: assert that `innerErrors` contains just 1 key/value.

lib/request-handler.js
tolbertam: :+1: that's much better, will change it.

tolbertam:
> I liked the idea about creating a generic simulacron test suite for execute() method. Wdyt about including a test that verifies what happens when a syntax error / invalid response error is obtained.

I started looking into doing that, but then realized we already have tests for that in `test/integration/short/error-simulator-tests.js` :).

jorgebay:
Hopefully we will be able to reduce the amount of unnecessary allocations of `createQueryOptions()` method on NODEJS-450, I'm sick of that method...

PR #282, opened 2018-07-24
Parse Virtual Keyspace Metadata

tolbertam:
Adds support for parsing virtual keyspace metadata.  Like the java driver implementation (datastax/java-driver#1050) virtual keyspaces and tables are exposed in the same way as regular keyspaces and tables.

In cases where where a user requests an explicit keyspace (`metadata.refreshKeyspaces`) or table (`metadata.getTable`) that weren't previously cached, we first query the `system_schema` keyspace.  If no keyspace/table is found, we then query the `system_virtual_schema` keyspace.  If we are somehow able to discern the keyspace/table are virtual ahead of time (i.e. if querying a table and we know its tied to a virtual keyspace) we query `system_virtual_schema` directly.   This logic only takes place for C* 4.0+ where I introduced a `SchemaParserV3`.

### DIFF
--- a/lib/metadata/schema-parser.js
+++ b/lib/metadata/schema-parser.js
@@ -1047,26 +1047,28 @@ SchemaParserV3.prototype.getKeyspaces = function (waitReconnect, callback) {
 
 /** @override */
 SchemaParserV3.prototype.getKeyspace = function (name, callback) {
-  const self = this;
-  this.cc.query(util.format(_selectSingleKeyspaceV2, name), function (err, result) {
+  this._getKeyspace(_selectSingleKeyspaceV2, name, false, (err, ks) => {
+    if (err) {
+      return callback(err);
+    }
+    if (!ks) {
+      // if not found, attempt to retrieve as virtual keyspace.
+      return this._getKeyspace(_selectSingleVirtualKeyspace, name, true, callback);
+    }
+    return callback(null, ks);
+  });
+};
+
+SchemaParserV3.prototype._getKeyspace = function (query, name, virtual, callback) {
+  this.cc.query(util.format(query, name), (err, result) => {
     if (err) {
       return callback(err);
     }
     const row = result.rows[0];
     if (!row) {
-      // if not found try virtual keyspace
-      return self.cc.query(util.format(_selectSingleVirtualKeyspace, name), function (err, result) {
-        if (err) {
-          return callback(err);
-        }
-        const row = result.rows[0];
-        if (!row) {
-          return callback(null, null);
-        }
-        callback(null, self._parseKeyspace(row, true));
-      });
+      return callback(null, null);
     }
-    callback(null, self._parseKeyspace(row, false));
+    callback(null, this._parseKeyspace(row, virtual));
   });
 };
 
--- a/lib/metadata/schema-parser.js
+++ b/lib/metadata/schema-parser.js
@@ -31,6 +31,12 @@ const _selectAggregatesV1 = "SELECT * FROM system.schema_aggregates WHERE keyspa
 const _selectAggregatesV2 = "SELECT * FROM system_schema.aggregates WHERE keyspace_name = '%s' AND aggregate_name = '%s'";
 const _selectMaterializedViewV2 = "SELECT * FROM system_schema.views WHERE keyspace_name = '%s' AND view_name = '%s'";
 
+const _selectAllVirtualKeyspaces = "SELECT * FROM system_virtual_schema.keyspaces";
+const _selectSingleVirtualKeyspace = "SELECT * FROM system_virtual_schema.keyspaces where keyspace_name = '%s'";
+const _selectVirtualTable = "SELECT * FROM system_virtual_schema.tables where keyspace_name = '%s' and table_name='%s'";
+const _selectVirtualColumns = "SELECT * FROM system_virtual_schema.columns where keyspace_name = '%s' and table_name='%s'";
+
+
 /**
  * @abstract
  * @param {ControlConnection} cc
@@ -45,6 +51,8 @@ function SchemaParser(cc) {
   this.selectUdt = null;
   this.selectAggregates = null;
   this.selectFunctions = null;
+  this.selectVirtualTable = null;
+  this.selectVirtualColumns = null;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


lib/metadata/schema-parser.js
```diff
@@ -732,15 +766,69 @@ SchemaParserV2.prototype._parseKeyspace = function (row) {
...
     row['durable_writes'],
     strategy,
-    strategyOptions);
+    strategyOptions,
+    virtual);
 };
 
 /** @override */
-SchemaParserV2.prototype._parseTableOrView = function (tableInfo, tableRow, columnRows, indexRows, callback) {
+SchemaParserV2.prototype._parseTableOrView = function (tableInfo, tableRow, columnRows, indexRows, virtual, callback) {
```
tolbertam: Could have implemented a `SchemaParserV3.prototype._parseTableOrView` override, but figured it would be fine to do here as older C* versions would never pass `virtual = true` anyways.

lib/metadata/schema-parser.js
```diff
@@ -941,6 +1004,79 @@ SchemaParserV2.prototype._parseUdt = function (udtInfo, row, callback) {
...
+ * keyspaces.
+ *
+ * @param {ControlConnection} cc The control connection to be used
+ * @param {Function} udtResolver The function to be used to retrieve the udts.
+ * @constructor
+ * @ignore
+ */
+function SchemaParserV3(cc, udtResolver) {
+  SchemaParserV2.call(this, cc, udtResolver);
+  this.selectVirtualTable = _selectVirtualTable;
```
jorgebay: NIT: We can use the `const` directly without the need of a instance level field.

lib/metadata/schema-parser.js
```diff
@@ -45,23 +51,27 @@ function SchemaParser(cc) {
...
  */
-SchemaParser.prototype._createKeyspace = function (name, durableWrites, strategy, strategyOptions) {
+SchemaParser.prototype._createKeyspace = function (name, durableWrites, strategy, strategyOptions, virtual) {
   const ksInfo = {
     name: name,
     durableWrites: durableWrites,
     strategy: strategy,
     strategyOptions: strategyOptions,
     tokenToReplica: null,
+    virtual: virtual === true ? true : false,
```
jorgebay: Minor nit: can be simplified to `virtual: virtual === true` or `virtual: !!virtual`

lib/metadata/schema-parser.js
jorgebay: Good call 👍

lib/metadata/schema-parser.js
```diff
@@ -941,6 +1004,79 @@ SchemaParserV2.prototype._parseUdt = function (udtInfo, row, callback) {
...
+    const row = result.rows[0];
+    if (!row) {
+      // if not found try virtual keyspace
+      return self._getVirtualKeyspace(name, callback);
+    }
+    callback(null, self._parseKeyspace(row, false));
+  });
+};
+
+SchemaParserV3.prototype._getVirtualKeyspace = function (name, callback) {
```
jorgebay: We could simplify this and the above logic into a single method that receives the query and whether or not it's virtual.

tolbertam:
Thanks for reviewing!  I'll make updates to address the feedback and push later today.

lib/metadata/schema-parser.js
tolbertam: :+1: , made that change, but also introduced a `supportsVirtual` instance field, as I still needed a way of communicating that a parser supported virtual tables (might be a better way though).

lib/metadata/schema-parser.js
jorgebay: To clarify, I meant using an instance method that calls `cc.query()` and parses the keyspace, in order to make `getKeyspace()` method be 2 nested calls:

```javascript
SchemaParserV3.prototype.getKeyspace = function (name, callback) {
  // Get normal (non-virtual) keyspace
  this._getKeyspace(_selectSingleKeyspaceV2, name, false, (err, ks) => {
    if (err) return callback(err);
    if (!ks) {
      // Try to get virtual ks
      return this._getKeyspace(_selectSingleKeyspaceV2, name, true, callback);
    }
    return callback(err, ks);
  });
};
```

tolbertam: Ah, I see.  I felt that there was something I was missing when I made that change.  I'll change it to your suggestion since that makes sense to me :+1:

lib/metadata/schema-parser.js
```diff
@@ -114,9 +124,11 @@ SchemaParser.prototype.getTable = function (keyspaceName, name, cache, callback)
...
   const self = this;
+  let virtualTable = virtual;
   utils.series([
     function getTableRow(next) {
-      const query = util.format(self.selectTable, keyspaceName, name);
+      const selectTable = virtualTable ? _selectVirtualTable : self.selectTable; 
+      const query = util.format(selectTable, keyspaceName, name);
       self.cc.query(query, function (err, response) {
         if (err) {
           return next(err);
```
tolbertam: Unlike other cases we error out here for virtual tables because in this case we know we are dealing with a table in a virtual keyspace, so propagating any errors seems sensible.

tolbertam:
Pushed 58873c1 for allowing the driver to behave gracefully when the server reports C* 4.0 but doesn't support virtual tables.   The only known case of this I'm aware of is DSE 6.0.  In this case, we simply ignore any errors.   This isn't easy to integration test since we don't involve DSE in our integration test suite for this driver, but I verified the following manually with DSE 6.0.2:

* With `isSchemaMetadataEnabled: true` 
    * Client is able to connect and successfully collects keyspace metadata (previously connected but keyspace metadata was empty and log emitted `keyspace system_virtual_schema does not exist`)
    * `Client.metadata.refreshKeyspaces` works
    * `Client.metadata.refreshKeyspace` returns empty result instead of error when requesting non existent keyspace.
    * `Client.metadata.getTable` returns empty result instead of error when requesting non existent table in non existent keyspace.
* With `isSchemaMetadataEnabled: false` 
    * `Client.metadata.refreshKeyspaces` works
    * `Client.metadata.refreshKeyspace` returns empty result instead of error when requesting non existent keyspace.
    * `Client.metadata.getTable` returns empty result instead of error when requesting non existent keyspace.

jorgebay:
I've run `metadata-tests.js` using DSE 6.0.0 and the latest fix works: virtual ks/tables can't be retrieved but it doesn't fail.

lgtm, :shipit:

PR #283, opened 2018-07-25
Bypass retry logic when query is marked as non-idempotent

tolbertam:
For [NODEJS-331](https://datastax-oss.atlassian.net/browse/NODEJS-331).

### DIFF
--- a/lib/policies/retry.js
+++ b/lib/policies/retry.js
@@ -155,7 +155,8 @@ RetryPolicy.retryDecision = {
  * [RetryPolicy]{@link module:policies/retry~RetryPolicy} as child policy.
  * @extends module:policies/retry~RetryPolicy
  * @constructor
- * @deprecated since 4.0 non-idempotent operations are never tried for write timeout or request error.
+ * @deprecated since 4.0 non-idempotent operations are never tried for write timeout or request error, use the default
+ * retry policy instead.
  */
 function IdempotenceAwareRetryPolicy(childPolicy) {
   this._childPolicy = childPolicy || new RetryPolicy();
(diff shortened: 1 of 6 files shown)


lib/policies/retry.js
```diff
@@ -155,6 +155,7 @@ RetryPolicy.retryDecision = {
  * [RetryPolicy]{@link module:policies/retry~RetryPolicy} as child policy.
  * @extends module:policies/retry~RetryPolicy
  * @constructor
+ * @deprecated since 4.0 non-idempotent operations are never tried for write timeout or request error.
```
jorgebay: Good call on deprecating and not removing it right away. Can we include a message to point to the correct policy, like "Use the default retry policy instead"?

lib/request-execution.js
```diff
@@ -149,7 +149,10 @@ class RequestExecution {
     };
     const self = this;
     function onRequestError() {
-      return self._parent.retryPolicy.onRequestError(operationInfo, self._request.consistency, err);
+      if(operationInfo.options.isIdempotent) {
```
jorgebay: I was going to suggest to short circuit above, at the beginning of `_getDecision()` method, but I see that when `err.isSocketError && err.requestNotWritten` it can be safely retried on next host... 👍

jorgebay:
Not related to this patch: The test `should failover when a node goes down with some outstanding requests` is failing intermittently, we should keep an eye in case we see it failing on master branch and if we can understand what is happening under the hood (most likely the test is poorly implemented by me).

lib/policies/retry.js
tolbertam: will do :+1:

jorgebay:
I've merged it to `master`, in order to pick up the changes NODEJS-450.

I'll add the breaking changes to the upgrade guide as part of NODEJS-450 as well.

PR #284, opened 2018-08-24
shutdown on readme

lfundaro:
**Problem**

- Someone who has never used this library will be puzzled why a simple code like:
```
client = new cassandra.Client(....)
client.execute(query)
```
will never let your node process exit. 
- Even though, doing a `shutdown()` it's something one should do for proper cleanup, when starting up development of an application it's not obvious that one needs to do this. 

**Proposal** 

- Add the shutdown command 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ const client = new cassandra.Client({ contactPoints: ['h1', 'h2'], keyspace: 'ks
 const query = 'SELECT name, email FROM users WHERE key = ?';
 client.execute(query, [ 'someone' ])
   .then(result => console.log('User with email %s', result.rows[0].email));
+client.shutdown()
 ```
 
 Alternatively, you can use the callback-based execution for all asynchronous methods of the API.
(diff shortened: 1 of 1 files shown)


jorgebay:
I think I understand what you mean. My main concern about including a `shutdown()` call on "Basic Usage" is that the user might think he/she needs to shutdown after each execution (like `connect()`/`shutdown()` or `open()`/`close()`).

Additionally, doing

```javascript
client.execute(query, [ 'someone' ])
  .then(result => console.log('User with email %s', result.rows[0].email));
client.shutdown();
```

Might lead to race conditions as `shutdown()` call is made before `execute()` completes.

There's some information on the [FAQ](https://docs.datastax.com/en/developer/nodejs-driver/3.5/faq/#should-i-create-one-client-instance-per-module-in-my-application) ([source](https://github.com/datastax/nodejs-driver/blob/master/doc/faq/README.md)), do you think it would be better to include an entry on the FAQ, like:

>How do I close all long-lived connections?
You should call `shutdown()` when you don't want to use the `Client` instance anymore and/or you want to clean up open handlers to exit the process.

lfundaro:
Hi @jorgebay , agree with you. 

Better to put this on the FAQ. Feel free to close this PR.

Cheers 😄

jorgebay:
Cheers @lfundaro !

PR #286, opened 2018-09-11
Support callback parameter in TimeUuid.now() and TimeUuid.fromDate()

jorgebay:
Expose optional callback parameter on `TimeUuid.now()` and
`TimeUuid.fromDate()` for non-blocking creation of random portions.

Additionally, I've made a minor improvement to Uuid equality comparison
and exposed TimeUuid#getClockId() method.

### DIFF
--- a/lib/types/time-uuid.js
+++ b/lib/types/time-uuid.js
@@ -19,6 +19,12 @@ const _unixToGregorian = 12219292800000;
  * @private
  */
 const _ticksInMs = 10000;
+
+const minNodeId = utils.allocBufferFromString('808080808080', 'hex');
+const minClockId = utils.allocBufferFromString('8080', 'hex');
+const maxNodeId = utils.allocBufferFromString('7f7f7f7f7f7f', 'hex');
+const maxClockId = utils.allocBufferFromString('7f7f', 'hex');
+
 /**
  * Counter used to generate up to 10000 different timeuuid values with the same Date
  * @private
@@ -37,6 +43,7 @@ let _ticksForCurrentTime = 0;
  * @type {number}
  */
 let _lastTimestamp = 0;
+
 /**
  * Creates a new instance of Uuid based on the parameters provided according to rfc4122.
  * If any of the arguments is not provided, it will be randomly generated, except for the date that will use the current date.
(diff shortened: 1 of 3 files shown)


lib/types/time-uuid.js
```diff
@@ -19,6 +19,12 @@ const _unixToGregorian = 12219292800000;
  * @private
  */
 const _ticksInMs = 10000;
+
+const minNodeId = utils.allocBufferFromString('808080808080', 'hex');
```
tolbertam: Good idea making these constants

lib/types/time-uuid.js
```diff
@@ -112,9 +167,25 @@ TimeUuid.max = function (date, ticks) {
...
+ * TimeUuid.now(function (err, timeuuid) {
+ *   // do something with the generated timeuuid
+ * });
+ * @example <caption>Generate a TimeUuid based on the current date (might block)</caption>
+ * const timeuuid = TimeUuid.now();
  */
-TimeUuid.now = function (nodeId, clockId) {
-  return new TimeUuid(null, null, nodeId, clockId);
+TimeUuid.now = function (nodeId, clockId, callback) {
+  return TimeUuid.fromDate(null, null, nodeId, clockId, callback);
```
tolbertam: Nice code reuse to centralize the callback handling in one place!

lib/types/time-uuid.js
```diff
@@ -76,9 +83,59 @@ util.inherits(TimeUuid, Uuid);
  * If not provided, a random nodeId will be generated.
  * @param {String|Buffer} [clockId] A 2-length Buffer or string of 6 ascii characters representing the clock identifier.
  * If not provided a random clockId will be generated.
+ * @param {Function} [callback] An optional callback to be invoked with the error as first parameter and the created
+ * <code>TimeUuid</code> as second parameter. When a callback is provided, the random portions of the
+ * <code>TimeUuid</code> instance are created asynchronously.
+ * <p>
+ *   When nodeId and/or clockId portions are not provided, this method will generate them using
```
tolbertam: We should also indicate this in the `TimeUuid` constructor since that can be used directly, and we could indicate to use `fromDate` or `now` if you desire for these values to be created asynchronously.

jorgebay: Great catch! I'll add it to the constructor as well.

PR #288, opened 2018-09-21
Introduce a wrapper around QueryOptions

jorgebay:
https://datastax-oss.atlassian.net/browse/NODEJS-450

Introduce `ExecutionInfo` as a base class with getters to obtain the options with defaults.
Execution methods like `Client#execute()`, `Client#batch()`, ... along with metadata retrieval API is not affected.

Other changes:

- Consolidate an Upgrade Guide and include the changes NODEJS-331.
- Remove `retryOnTimeout` query option.


test/integration/short/client-pool-tests.js
```diff
@@ -740,7 +740,7 @@ describe('Client', function () {
       ], done);
     });
     it('should failover when a node goes down with some outstanding requests', function (done) {
-      const options = utils.extend({}, helper.baseOptions);
+      const options = utils.extend({ queryOptions: { isIdempotent: true } }, helper.baseOptions);
```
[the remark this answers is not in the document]
jorgebay: It looks like it, I've merged that fix to `master`: 486abc749e299584432bf6751b5a3643bcb614cc

jorgebay:
I've removed `meta` property from `ResultSet` (and included in the upgrade guide) and rebased against `master`.

doc/upgrade-guide/README.md
```diff
@@ -0,0 +1,85 @@
...
+- `options` property was replaced by `executionInfo` which is an instance of `ExecutionInfo`.
+
+### Removed `meta` property from `ResultSet`
+
+On earlier versions of the driver, the `ResultSet` exposed the property `meta` which contained the raw result metadata.
+This property was removed in the latest version.
+
+---
+
+## 2.0
```
jorgebay: There's a trivial breaking change in 3.0 related to cql functions / aggregations IIRC that it would be nice to add here.

jorgebay:
I've just rebased it.

lib/execution-info.js
```diff
@@ -0,0 +1,526 @@
...
+
+const utils = require('./utils');
+const types = require('./types');
+const errors = require('./errors');
+
+/**
+ * A base class that represents a wrapper around the user provided query options with getter methods and proper
+ * default values.
+ */
+class ExecutionInfo {
```
tolbertam: Maybe i am nitpicking too much but `ExecutionInfo`  does not feel descriptive enough about what it is.  These are really options / configuration and "Info" to me implies that it's metadata/informational.  What do you think about `ExecutionOptions` or `ExecutionConfig`?    Another possibility is referring to things as a `Request` i.e. `RequestOptions` / `RequestConfig`.

`ExecutionInfo` could also be seen as overloaded as other drivers ([c#](https://docs.datastax.com/en/latest-csharp-driver-api/html/T_Cassandra_ExecutionInfo.htm), [java](https://docs.datastax.com/en/drivers/java/3.6/com/datastax/driver/core/ExecutionInfo.html)).

lib/execution-info.js
```diff
@@ -0,0 +1,526 @@
...
+
+  /**
+   * Determines whether the driver must retrieve the following result pages automatically.
+   * <p>
+   *   This setting is only considered by the [Client#eachRow()]{@link Client#eachRow} method.
+   * </p>
+   * @abstract
+   * @returns {Boolean}
+   */
+  getIsAutoPage() {
```
tolbertam: Not sure if there is a convention, but would it make sense for these boolean-based property getters to be simply `isAutoPage()`, `isBatchCounter`, `isBatchLogged`, etc.?

lib/execution-info.js
```diff
@@ -0,0 +1,526 @@
...
+   * Determines whether query tracing is enabled for the execution.
+   * @abstract
+   * @returns {Boolean}
+   */
+  getIsQueryTracing() {
+
+  }
+
+  /**
+   * Gets the keyspace for the query.
```
tolbertam: For properties that might be null like this one and paging state, should we also indicate that they may return null if unset?

doc/upgrade-guide/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+```javascript
+const client = new Client({ contactPoints, queryOptions: { isIdempotent: true }})
+```
+
+Previously, a similar behaviour was available using `IdempotenceAwareRetryPolicy`, that is now marked as deprecated.
+
+### Removed `retryOnTimeout` property of `QueryOptions`
+
+`retryOnTimeout`, the property that controlled whether a request should be tried when a response wasn't obtained
```
tolbertam: Nice, makes a lot of sense to just give control in `RetryPolicy`

doc/upgrade-guide/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+The retry policy methods takes [`OperationInfo`][op-info] as a parameter. Some `OperationInfo` properties changes or 
+were removed.
+
+- Deprecated properties `handler`, `request` and `retryOnTimeout` were removed.
+- `options` property was replaced by `executionInfo` which is an instance of `ExecutionInfo`.
+
+### Removed `meta` property from `ResultSet`
+
+On earlier versions of the driver, the `ResultSet` exposed the property `meta` which contained the raw result metadata.
```
tolbertam: I know that this was mostly exposed for sharing paging state, but could users have been relying on meta to access other data like result types?

lib/execution-info.js
jorgebay: I think its a valid concern, specially given that `ExecutionInfo` is already taken by other drivers.

`ExecutionOptions` is a nice alternative: no collision and clear meaning. I'll change it.

lib/execution-info.js
jorgebay: I was going for `getX()` and `setX()` methods but maybe is not a good idea for booleans. Given that set methods are only internal, the user wont see this (get/set) convention, I'll change them to `isX()` 👍

lib/execution-info.js
jorgebay: I've included a note at class level (possible `undefined` return values) and some info regarding what it means for keyspace to be `undefined`.

Most values depend on the default values, that can be defined by the user when creating the `Client` instance,  so I find it hard to document possible return values... :)

TLDR return values can be `undefined` with few exceptions (lbp, retry policy, some flags, ...)

doc/jsdoc.json
```diff
@@ -7,6 +7,7 @@
       "lib/metadata",
       "lib/client.js",
       "lib/execution-profile.js",
+      "lib/execution-info.js",
```
tolbertam: Should be `lib/execution-options.js` now

lib/client.js
```diff
@@ -518,8 +519,15 @@ Client.prototype.execute = function (query, params, options, callback) {
     params = typeof params !== 'function' ? params : null;
   }
   return utils.promiseWrapper.call(this, this.options, callback, function handler(cb) {
-    options = clientOptions.createQueryOptions(this, options);
-    this._innerExecute(query, params, options, cb);
+    let info;
```
tolbertam: nit, rename `info` to `executionOptions`  or something like that.  (There are a few other instances of this)

doc/jsdoc.json
jorgebay: Good catch!

I've created a ticket to generate docs in Jenkins for this type of validation: https://datastax-oss.atlassian.net/browse/NODEJS-478

lib/client.js
jorgebay: I thought I replaced them all :/
I'll fix these ones and double check others.

PR #289, opened 2018-09-21
Set the distance on newly added nodes

jorgebay:
https://datastax-oss.atlassian.net/browse/NODEJS-472

Fixes an issue where connections attempts were made to newly added nodes without considering the host distance.

### DIFF
--- a/lib/host.js
+++ b/lib/host.js
@@ -56,33 +56,38 @@ function Host(address, protocolVersion, options, metadata) {
   this._distance = types.distance.ignored;
   Object.defineProperty(this, '_metadata', { value: metadata, enumerable: false });
   this.reconnectionSchedule = this.options.policies.reconnection.newSchedule();
+  this.reconnectionDelay = 0;
 }
 
 util.inherits(Host, events.EventEmitter);
 
 /**
- * Marks this host as not available for query coordination.
+ * Marks this host as not available for query coordination, when the host was previously marked as UP, otherwise its
+ * a no-op.
  * @internal
  * @ignore
  */
 Host.prototype.setDown = function() {
-  // multiple events signaling that a host is failing could cause multiple calls to this method
+  // Multiple events signaling that a host is failing could cause multiple calls to this method
   if (this.setDownAt !== 0) {
     // the host is already marked as Down
     return;
   }
+
   if (this.pool.isClosing()) {
     // the pool is being closed/shutdown, don't mind
     return;
   }
+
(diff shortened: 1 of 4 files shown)


test/integration/short/pool-simulator-tests.js
```diff
@@ -87,6 +88,89 @@ describe('pool', function () {
...
+      cluster = new simulacron.SimulacronCluster();
+      cluster.register([3, 3], null, done);
+    });
+
+    afterEach(done => cluster.unregister(done));
+
+    it('should not attempt to make a connection on a new node that is on an ignored/remote DC', () => {
+      // For the purpose of this test, we will use an address translator that will create the impression
+      // that on the second (ignored) DC, there are 2 nodes instead of 3
+      // Then, the address translator will show 3 nodes, giving the impression that the third node is "new".
```
landim: I might be missing something, but imho it would be easier to use the `DELETE` method of  `/connections/{clusterId}/{dataCenterId}/{nodeId}` of simulacron.

lib/host.js
```diff
@@ -266,16 +271,21 @@ Host.prototype._checkPoolState = function () {
...
   }
-  if (this._distance !== types.distance.ignored &&
-      this.pool.connections.length === 0 &&
-      this.pool.coreConnectionsLength > 0) {
+
+  if (
+    this._distance !== types.distance.ignored &&
+    this.pool.connections.length === 0 &&
+    this.pool.coreConnectionsLength > 0
+  ) {
```
landim: I would not change the code style in this `if`.

test/integration/short/pool-simulator-tests.js
```diff
@@ -5,16 +5,17 @@ const simulacron = require('../simulacron');
...
 
 describe('pool', function () {
 
-  this.timeout(5000);
+  this.timeout(20000);
   before(done => simulacron.start(done));
   after(done => simulacron.stop(done));
 
-  describe('with a 3-node simulated cluster', function () {
+  context('with a 3-node simulated cluster', function () {
```
landim: Minor change, but makes the easier to read the test. 👍  on using `context`

lib/host.js
jorgebay: I'm not a big fan of any multiline if code style, I've changed to use airbnb's convention: https://github.com/airbnb/javascript/issues/1380

I think the clearest way would be to assign it to a variable, i.e., `shouldHaveConnections`.

PR #292, opened 2018-09-28
Metrics API

jorgebay:
Expose a metrics interface and create a default implementation that emits those metrics as events.

### DIFF
--- a/lib/client-options.js
+++ b/lib/client-options.js
@@ -116,7 +116,7 @@ function extend(baseOptions, userOptions) {
   }
 
   if (!(options.metrics instanceof metrics.ClientMetrics)) {
-    throw new TypeError('metrics must be an instance of ClientMetrics or null');
+    throw new TypeError('metrics must be an instance of ClientMetrics');
   }
 
   validatePoliciesOptions(options.policies);
--- a/lib/client-options.js
+++ b/lib/client-options.js
@@ -6,6 +6,7 @@ const types = require('./types');
 const utils = require('./utils');
 const errors = require('./errors');
 const tracker = require('./tracker');
+const metrics = require('./metrics');
 
 /** Core connections per host for protocol versions 1 and 2 */
 const coreConnectionsPerHostV2 = {
@@ -68,6 +69,7 @@ function defaultOptions () {
     },
     authProvider: null,
     requestTracker: null,
+    metrics: new metrics.DefaultMetrics(),
     maxPrepared: 500,
     refreshSchemaDelay: 1000,
     isMetadataSyncEnabled: true,
@@ -111,6 +113,10 @@ function extend(baseOptions, userOptions) {
 
   if (options.requestTracker !== null && !(options.requestTracker instanceof tracker.RequestTracker)) {
     throw new TypeError('requestTracker must be an instance of RequestTracker');
+  }
+
+  if (!(options.metrics instanceof metrics.ClientMetrics)) {
+    throw new TypeError('metrics must be an instance of ClientMetrics or null');
   }
 
   validatePoliciesOptions(options.policies);
--- a/lib/metrics/client-metrics.js
+++ b/lib/metrics/client-metrics.js
@@ -53,24 +53,28 @@ class ClientMetrics {
 
   /**
    * Method invoked when an execution is retried as a result of a client-level timeout.
+   * @param {Error} e The error that caused the retry.
    */
-  onClientTimeoutRetry() {}
+  onClientTimeoutRetry(e) {}
 
   /**
    * Method invoked when an execution is retried as a result of a read timeout from the server (coordinator to replica).
+   * @param {Error} e The error that caused the retry.
    */
-  onReadTimeoutRetry() {}
+  onReadTimeoutRetry(e) {}
 
   /**
    * Method invoked when an execution is retried as a result of an unavailable error from the server.
+   * @param {Error} e The error that caused the retry.
    */
-  onUnavailableRetry() {}
+  onUnavailableRetry(e) {}
 
   /**
    * Method invoked when an execution is retried as a result of a write timeout from the server (coordinator to
    * replica).
+   * @param {Error} e The error that caused the retry.
    */
-  onWriteTimeoutRetry() {}
+  onWriteTimeoutRetry(e) {}
 
   /**
    * Method invoked when an error (other than a server or client timeout) is retried.
@@ -80,6 +84,7 @@ class ClientMetrics {
 
   /**
    * Method invoked when an error is marked as ignored by the retry policy.
+   * @param {Error} e The error that was ignored by the retry policy.
    */
   onIgnoreError(e) {}
 
@@ -87,6 +92,13 @@ class ClientMetrics {
    * Method invoked when a speculative execution is started.
    */
   onSpeculativeExecution() {}
+
+  /**
+   * Method invoked when a response is obtained successfully.
+   * @param {Array<Number>} latency The latency represented in a <code>[seconds, nanoseconds]</code> tuple
+   * Array, where nanoseconds is the remaining part of the real time that can't be represented in second precision.
+   */
+  onSuccessfulResponse(latency) {}
 }
 
 module.exports = ClientMetrics;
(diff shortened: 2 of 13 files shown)


lib/metrics/client-metrics.js
```diff
@@ -0,0 +1,103 @@
...
+   * Method invoked when a speculative execution is started.
+   */
+  onSpeculativeExecution() {}
+
+  /**
+   * Method invoked when a response is obtained successfully.
+   * @param {Array<Number>} latency The latency represented in a <code>[seconds, nanoseconds]</code> tuple
+   * Array, where nanoseconds is the remaining part of the real time that can't be represented in second precision.
+   */
+  onSuccessfulResponse(latency) {}
```
tolbertam: I think it would be nice to report also on any response (successful or not) as that would give a better view of how long it takes for the Client to be surfaced a response in general (similar to java-driver's [`Metrics.getRequestsTimer()`](https://docs.datastax.com/en/drivers/java/3.6/com/datastax/driver/core/Metrics.html#getRequestsTimer--).

lib/metrics/default-metrics.js
```diff
@@ -0,0 +1,175 @@
...
+ * </p>
+ * @implements {module:metrics~ClientMetrics}
+ * @alias module:metrics~DefaultMetrics
+ * @example <caption>Listening to events emitted</caption>
+ * defaultMetrics.errors.on('increment', err => totalErrors++);
+ * defaultMetrics.errors.clientTimeout.on('increment', () => clientTimeoutErrors++);
+ * defaultMetrics.speculativeRetries.on('increment', () => specExecsCount++);
+ * defaultMetrics.successfulResponses.on('increment', latency => specExecsCount++);
+ */
+class DefaultMetrics extends ClientMetrics {
```
tolbertam: This is a neat design because it almost serves as a Dual API.   If a user wants, they can just subscribe on the event emitters, or they can implement their own `ClientMetrics` interface.

lib/client-options.js
```diff
@@ -68,6 +69,7 @@ function defaultOptions () {
     },
     authProvider: null,
     requestTracker: null,
+    metrics: new metrics.DefaultMetrics(),
```
tolbertam: Do we expect much of an added performance cost for emitting events with the `DefaultMetrics` implementation?  I imagine at least initially, users will not use metrics, so if emitting does have a performance cost, maybe we default to a `no-op` implementation where the `onXX` methods don't do anything.

lib/client-options.js
```diff
@@ -113,6 +115,10 @@ function extend(baseOptions, userOptions) {
     throw new TypeError('requestTracker must be an instance of RequestTracker');
   }
 
+  if (!(options.metrics instanceof metrics.ClientMetrics)) {
+    throw new TypeError('metrics must be an instance of ClientMetrics or null');
```
tolbertam: technically, wouldn't it being null be an issue as we invoke the `onXXX` methods without null checking?  Should it be strictly non-null or a `ClientMetrics` instance?

lib/metrics/default-metrics.js
```diff
@@ -0,0 +1,175 @@
...
+ *   An instance of [DefaultMetrics]{@link module:metrics~DefaultMetrics} is configured by default in the client,
+ *   you can access this instance using [Client#metrics]{@link Client#metrics} property.
+ * </p>
+ * @implements {module:metrics~ClientMetrics}
+ * @alias module:metrics~DefaultMetrics
+ * @example <caption>Listening to events emitted</caption>
+ * defaultMetrics.errors.on('increment', err => totalErrors++);
+ * defaultMetrics.errors.clientTimeout.on('increment', () => clientTimeoutErrors++);
+ * defaultMetrics.speculativeRetries.on('increment', () => specExecsCount++);
+ * defaultMetrics.successfulResponses.on('increment', latency => specExecsCount++);
```
tolbertam: This should probably be `responseCount++` or maybe `totalLatencySeconds += latency[0]`

test/integration/short/retry-simulator-tests.js
```diff
@@ -0,0 +1,325 @@
+'use strict';
+
+const assert = require('assert');
```
tolbertam: Excellent test approach!

lib/metrics/client-metrics.js
```diff
@@ -0,0 +1,103 @@
+'use strict';
+
+/**
+ * Represents a base class that is used to measure events from the server and the client as seen by the driver.
+ * @alias module:metrics~ClientMetrics
+ * @interface
+ */
+class ClientMetrics {
```
tolbertam: I was thinking maybe it would be nice to provide more information as parameters (like Host where error was encountered), but then I think we would start veering into conflating with the Request Tracker implementation and I remembered your l[comment on NODEJS-300](https://datastax-oss.atlassian.net/browse/NODEJS-300?focusedCommentId=50558&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-50558) which covers this.

I think if anyone wants to do something very specific, such as Node-level metrics, they could just use `RequestTracker`, so I like this approach where the API is very simple as it makes it very easy for users to wire in their own functionality.

tolbertam: Just had a realization.  Currently, all metrics are based on some kind of activity, what about state-based metrics?  For example both the python and java driver expose inflight request counters, total number of connections, connected host count, etc.  Technically those could be event based too, but wonder if sharing come kind of state as a metric is something we should consider.

tolbertam:
I wrote my own implementation of `ClientMetrics` and it was super easy and works well!

lib/client-options.js
jorgebay: The cost of emitting events is only the function call, [when there are no listeners to the event it can be considered no-op](https://github.com/nodejs/node/blob/v10.13.0/lib/events.js#L176-L179).

lib/client-options.js
jorgebay: Good catch, initially I was setting the default instance later but I changed the behaviour and not the error message.
I'll change it.

lib/metrics/client-metrics.js
jorgebay: State info is currently exposed via `client.getState()`. I was also wondering whether we should expose it in the metrics but I didn't see it clearly (should this be refreshed on each call / periodically ?) and the user can always implement a provider specific gauge using `ClientState` information.

lib/metrics/client-metrics.js
jorgebay: Good idea!
I'll add it.

lib/metrics/default-metrics.js
jorgebay: 🤦

jorgebay:
I think I've addressed all the concerns / suggestions.

Regarding response meters both here and the `RequestTracker`, I found a bug: responses obtained after a speculative execution already "won" are not included or tracked. I've filed [NODEJS-479](https://datastax-oss.atlassian.net/browse/NODEJS-479) for it and I was tempted to try to address it in this PR but I've started digging and it's more complex than it seems. We can tackle it before this release or leave it as a known issue.

tolbertam:
Thanks!  I'll give it a final review pass.

```
Regarding response meters both here and the RequestTracker, I found a bug: responses obtained after a speculative execution already "won" are not included or tracked. I've filed NODEJS-479 for it and I was tempted to try to address it in this PR but I've started digging and it's more complex than it seems. We can tackle it before this release or leave it as a known issue.
```

It's a bit subjective, but I think that might be ok.  I believe the java-driver behaves the same way, after a response is received, all subsequent responses are discarded and not recorded, although I do see the point that it's an incomplete view as you may have a slow node and without recording anything your metrics aren't a complete representation.

lib/metrics/client-metrics.js
tolbertam: > the user can always implement a provider specific gauge using ClientState information.

Just realized that's precisely what I'm doing already :man_facepalming: .  I think I agree, lets keep `ClientMetrics` to only activity-based things, and users can use `Client.getState` to get client & host state.

PR #294, opened 2018-09-28
Fix RangeError on empty buffer in Encoder.decodeMap

Gaen:
I've encountered an error similar to the one described here: https://stackoverflow.com/questions/42119104/rangeerror-index-out-of-range-when-connect-to-cassandra-databse-in-node-js

Null value is retrieved from map as empty buffer. Looks like the condition I've modified was intended to handle this situation properly but in fact it does not, leading to RangeError while attempting to read from empty buffer.

This fix helped me get rid of RangeError.


Gaen:
Force-pushed the update with more explicit implementation and tests

Gaen:
Back from despair, all tests green. Yes we can!

jorgebay:
Landed in 0deac79bf06f0ac5689c0b4d7123a3e9c867d45e, thanks @Gaen !

PR #298, opened 2018-10-18
Borrow a connection from the pool before retrying on the same host

jorgebay:
https://datastax-oss.atlassian.net/browse/NODEJS-474

Use a different connection from the pool (if possible) when retrying on the same host.
It contains a minor improvement to include the host address for `OperationTimedOutError`.

### DIFF
--- a/lib/policies/retry.js
+++ b/lib/policies/retry.js
@@ -197,6 +197,10 @@ IdempotenceAwareRetryPolicy.prototype.onWriteTimeout = function (info, consisten
  * [retryDecision]{@link module:policies/retry~RetryPolicy.retryDecision}.
  * @property {Number} [consistency] The [consistency level]{@link module:types~consistencies}.
  * @property {useCurrentHost} [useCurrentHost] Determines if it should use the same host to retry the request.
+ * <p>
+ *   In the case that the current host is not available anymore, it will be retried on the next host even when
+ *   <code>useCurrentHost></code> is set to <code>true</code>.
+ * </p>
  */
 
 /**
--- a/lib/policies/retry.js
+++ b/lib/policies/retry.js
@@ -199,7 +199,7 @@ IdempotenceAwareRetryPolicy.prototype.onWriteTimeout = function (info, consisten
  * @property {useCurrentHost} [useCurrentHost] Determines if it should use the same host to retry the request.
  * <p>
  *   In the case that the current host is not available anymore, it will be retried on the next host even when
- *   <code>useCurrentHost></code> is set to <code>true</code>.
+ *   <code>useCurrentHost</code> is set to <code>true</code>.
  * </p>
  */
 
--- a/lib/request-execution.js
+++ b/lib/request-execution.js
@@ -7,7 +7,7 @@ const retry = require('./policies/retry');
 const types = require('./types');
 const utils = require('./utils');
 
-const retryOnNextCurrentHost = Object.freeze({
+const retryOnCurrentHost = Object.freeze({
   decision: retry.RetryPolicy.retryDecision.retry,
   useCurrentHost: true,
   consistency: undefined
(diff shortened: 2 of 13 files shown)


jorgebay:
hm... it looks like there is a test fail on the windows builds. I can't reproduce it locally, so I'll use appveyor's output.

jorgebay:
It looks like its now fixed, I've started several builds on appveyor and the tests pass with the latest changes.

lib/policies/retry.js
```diff
@@ -197,6 +197,10 @@ IdempotenceAwareRetryPolicy.prototype.onWriteTimeout = function (info, consisten
  * [retryDecision]{@link module:policies/retry~RetryPolicy.retryDecision}.
  * @property {Number} [consistency] The [consistency level]{@link module:types~consistencies}.
  * @property {useCurrentHost} [useCurrentHost] Determines if it should use the same host to retry the request.
+ * <p>
+ *   In the case that the current host is not available anymore, it will be retried on the next host even when
+ *   <code>useCurrentHost></code> is set to <code>true</code>.
```
landim: Extra `>` at `userCurrentHost>`

test/integration/short/pool-simulator-tests.js
```diff
@@ -86,6 +91,98 @@ describe('pool', function () {
...
+
+          // The pool will be busy after this request, so the retry will occur on the next host
+          const p2 = client.execute('SELECT * FROM paused_on_first', null, { readTimeout: 50, isIdempotent: true });
+          promises.push(p2);
+          return p2;
+        })
+        // The query will be retried on the next host
+        .then(rs => assert.strictEqual(rs.info.queriedHost, client.hosts.values()[1].address))
+        .then(() => Promise.all(promises))
+        .then(() => assert.strictEqual(retryPolicyCalled, 1))
```
landim: It would be good to have an server side verification. e.g. Checking the logs at simulacron to verify if the request was made a different (next) host. Wdyt?

tolbertam: I think that the code is able to validate the queried host is probably good enough for the intent of the test.

lib/request-execution.js
```diff
@@ -193,22 +193,38 @@ class RequestExecution {
...
+
     if (useCurrentHost !== false) {
-      // Use existing host (default)
-      return this._sendOnConnection();
+      // Use existing host (default).
+      // Reusing the existing connection is suitable for the most common scenarios, like server read timeouts that
+      // will be fixed with a new request.
+      // To cover all scenarios (e.g., where a different connection to the same host might mean something different),
+      // we obtain a new connection from the host pool.
+      return this._host.borrowConnection((err, connection) => {
```
tolbertam: Might this just return the same connection that was previously used?  Should we add some logic to `borrowConnection` to allow passing existing connection(s) which communicates 'give me a connection that isn't any of these'?

lib/policies/retry.js
jorgebay: will fix!

lib/request-execution.js
jorgebay: I was considering that the connection was likely to be removed from the pool by then (`removeFromPool()` is called on `_handleError()` method), but I haven't consider that the retry might be triggered by errors that don't cause the connection to be removed (e.g., `overloaded`).

Great remark! I'll change it to try to avoid using the same connection when possible (more than 1 open connection):
- When there's 2 or more connections, use another one.
- When there's only 1 connection, use the same one.

tolbertam: This could be a good optimization for the other drivers as well.  The current state of this PR behaves exactly like the java driver.  Maybe a good confab discussion.

jorgebay:
While trying to fix the issue related to borrowing the same connection, I've found the following issues:
- When there's a socket error and the request was not written (was queued but never got to the socket), the driver doesn't rely on the `RetryPolicy` and blindly retries. Previously, it retried on the next host, I've changed to retry on the same host first (will fallback to next host, if there isn't a connection available).
- Calling `host.borrowConnection()` might return a `Connection` that is not on the current keyspace (e.g., recently opened connection). I've changed by moving the logic of switching the keyspace to the pool.

jorgebay:
@tolbertam I've pushed the latest fixes, if you want to take another pass at it.

tolbertam:
:+1: , taking a look now.

PR #301, opened 2018-11-13
Host in-flight counters and health checks

jorgebay:
Add a host-level internal counter for the amount of in-flights and use it for `ClientState`.

Additionally, expose an internal method to get the amount of responses dequeued in the last interval on all connections to the host.


jorgebay:
There is an issue on the tests: `Host#shutdown()` is not being called sometimes, causing the event loop to have queued handlers and preventing mocha to exit.

I'll fix them.

tolbertam:
> There is an issue on the tests: Host#shutdown() is not being called sometimes, causing the event loop to have queued handlers and preventing mocha to exit.

Is that because `Client` instances aren't always closed in tests?

jorgebay:
> Is that because `Client` instances aren't always closed in tests?

The culprit was the new `_healthResponseCountTimer` created at `Host` level. It requires that we call `Host#shutdown()` (which already existed and is called by `Client#shutdown()` ) to clear the timer but on the tests were `Host` instances where created `shutdown()` method was not invoked.

PR #302, opened 2018-11-15
Add required 'localDataCenter' Client option

tolbertam:
For [NODEJS-379](https://datastax-oss.atlassian.net/browse/NODEJS-379).

Adds a new option 'localDataCenter' that is required when using
DCAwareRoundRobinPolicy.

Deprecates the localDc argument in DCAwareRoundRobinPolicy.  If
specified it is still used, even in favor of 'localDataCenter'.

Update tests and examples to use localDataCenter where applicable.

Changes ClientOptions.defaultOptions to not set a default noop
logEmitter.  This was overridding the default Client logEmitter.
Instead, on utils.log if no logEmitter is present (which should
only be the case in tests), noop is used.


examples/basic/basic-connect.js
```diff
@@ -11,5 +15,5 @@ client.connect()
   })
   .catch(function (err) {
     console.error('There was an error when connecting', err);
-    return client.shutdown();
+    return client.shutdown().then(() => process.exit(1));
```
tolbertam: I noticed that runner.js was marking examples as passed even though they were actually failing.  Registering a non zero exit code in these cases did the trick.

lib/client.js
```diff
@@ -311,7 +316,7 @@ const warmupLimit = 32;
  * @extends EventEmitter
  * @param {ClientOptions} options The options for this instance.
  * @example <caption>Creating a new client instance</caption>
- * const client = new Client({ contactPoints: ['192.168.1.100'] });
+ * const client = new Client({ contactPoints: ['192.168.1.100'], localDataCenter: 'datacenter1' });
```
tolbertam: I chose to specify `datacenter1` in both the examples and docs because this is what Cassandra uses with default settings, and also what ccm uses for single data center.

lib/policies/load-balancing.js
```diff
@@ -99,7 +99,9 @@ RoundRobinPolicy.prototype.newQueryPlan = function (keyspace, executionOptions,
  * after the local nodes. In other words, this policy guarantees that no
  * host in a remote data center will be queried unless no host in the local
  * data center can be reached.
- * @param {?String} [localDc] local datacenter name.
+ * @param {?String} [localDc] Deprecated: local datacenter name.  This value is
```
tolbertam: Thinking about this some more, I suspect maybe we shouldn't mark this deprecated.   One use case where you might want to use this is when using execution profiles and you want to have different requests going to different data centers.  Instead proposing:

> local datacenter name.  This value overrides the 'localDataCenter' Client option and is useful for cases where you have multiple execution profiles that you intend on using for routing requests to different data centers.

lib/client-options.js
```diff
@@ -102,9 +102,6 @@ function extend(baseOptions, userOptions) {
         'the following values are valid contact points: ipAddress, hostName or ipAddress:port', i, hostName));
     }
   }
-  if (!options.logEmitter) {
```
tolbertam: I noticed that when using `extends` logging stopped working because of this.  Felt it'd be better instead to not do this so the default behavior is maintained when the user doesn't provide a `logEmitter`

tolbertam:
Another thing that should probably be done.  If a DC name is provided for which there are no hosts to connect to, we should raise a specific error.  Currently we surface an error such as `All host(s) tried for query failed.` which isn't very clear.  Will look into this.

lib/policies/load-balancing.js
tolbertam: I ended up undeprecating this, but let me know if I should re-evaluate.

test/integration/short/pool-simulator-tests.js
```diff
@@ -232,6 +232,7 @@ describe('pool', function () {
 
       const client = new Client({
         contactPoints: cluster.getContactPoints(),
+        localDataCenter: 'dc1',
```
tolbertam: its not clear to me why simulacron provisions dc1 as the first dc in one case, and dc0 in another..will see if I can wrap my head around that tomorrow.

lib/control-connection.js
```diff
@@ -685,6 +694,14 @@ ControlConnection.prototype.setPeersInfo = function (initializing, err, result,
       });
       self.hosts.removeMultiple(toRemove);
     }
+    if (initializing && self.options.localDataCenter) {
+      const localDc = self.options.localDataCenter;
+
+      if (!foundDataCenters.has(localDc)) {
+        return callback(new errors.ArgumentError('localDataCenter was configured as \'' + localDc + '\', but only found' +
```
tolbertam: Wasn't sure if this is the right place for this, or if we should error out or warn.  The way this currently behaves is if no hosts are found using the specified local DC, the code errors out with:

```
NoHostAvailableError: All host(s) tried for query failed. First host tried, 127.0.0.101:9042: ArgumentError: localDataCenter was configured as 'dc2', but only found hosts in data centers: [dc1]. See innerErrors.
      at whilstEnded (lib/control-connection.js:231:25)
      at next (lib/utils.js:825:14)
      at Object.whilst (lib/utils.js:819:3)
      at ControlConnection.borrowAConnection (lib/control-connection.js:205:9)
      at Array.getConnectionTask (lib/control-connection.js:383:12)
      at /home/tolbertam/Documents/Projects/nodejs-driver/lib/utils.js:692:21
      at process._tickCallback (internal/process/next_tick.js:61:11)
```

test/integration/short/pool-simulator-tests.js
tolbertam: In some tests we were explicitly provisioning dc0, have since fixed this.

lib/control-connection.js
jorgebay: Good call, validating it here prevents duplicating the logic on the LBPs.

lib/policies/load-balancing.js
```diff
@@ -128,25 +130,21 @@ DCAwareRoundRobinPolicy.prototype.init = function (client, hosts, callback) {
...
-        break;
-      }
-    }
-    //this should never happen but it does not hurt
-    if (!this.localDc) {
-      return callback(new errors.DriverInternalError('Local datacenter could not be determined'));
+
+  if (client && client.options) {
+    if (this.localDc && !client.options.localDataCenter) {
+      this.client.log('warning', 'Local data center \'' + this.localDc + '\' was provided as an argument to' +
```
jorgebay: I'm not sure if this should be logged at "warning" level or at "info" level: users might be using 2 execution profiles, e.g. "search" and "oltp" with 2 different dcs.

I think info level is a better match for conditional statements, like "its preferable to specify the local dc using ... _when your application is targeting a single datacenter_"

lib/policies/load-balancing.js
```diff
@@ -128,25 +130,21 @@ DCAwareRoundRobinPolicy.prototype.init = function (client, hosts, callback) {
...
-    client.log('warning', 'No local Data Center was provided with DCAwareRoundRobinPolicy' +
-      '.  Using discovered DC \'' + this.localDc + '\' from host ' + dcHost.address +
-      '.  Future releases will require local DC to be specified.');
+
+    // If localDc is unset, use value set in client options.
+    this.localDc = this.localDc || client.options.localDataCenter;
+  }
+
+  if (!this.localDc) {
+    return callback(new errors.DriverInternalError('\'localDataCenter\' is not defined in Client options and also was' +
```
jorgebay: We can use `ArgumentError` as this can only be specified by the user

```suggestion
    return callback(new errors.ArgumentError('\'localDataCenter\' is not defined in Client options and also was' +
```

test/unit/load-balancing-tests.js
```diff
@@ -450,9 +451,27 @@ describe('DCAwareRoundRobinPolicy', function () {
...
+    delete options.localDataCenter;
+    const client = new Client(options);
+    const logEvents = [];
+    client.on('log', function(level, className, message, furtherInfo) {
+      logEvents.push({level: level, className: className, message: message, furtherInfo: furtherInfo});
+    });
+    const hosts = new HostMap();
+    hosts.set('1', createHost('1', client.options));
+    policy.init(client, hosts, (err) => {
+      helper.assertInstanceOf(err, errors.DriverInternalError);
```
jorgebay: If we change the error class used, we should change it here also.

doc/upgrade-guide/README.md
```diff
@@ -9,6 +9,12 @@ If you have any questions or comments, you can [post them on the mailing list][m
 
 The following is a list of changes made in version 4.0 of the driver that are relevant when upgrading from version 3.x.
 
+### localDataCenter is now a required Client option
```
jorgebay: 👍 on adding this info first in the upgrade guide.

I'll adapt it on the DSE driver side.

examples/basic/basic-connect.js
```diff
@@ -1,7 +1,11 @@
 "use strict";
 const cassandra = require('cassandra-driver');
 
-const client = new cassandra.Client({ contactPoints: ['127.0.0.1']});
+const client = new cassandra.Client({ contactPoints: ['127.0.0.1'], localDataCenter: 'datacenter1' });
+client.on('log', function(level, className, message, furtherInfo) {
```
jorgebay: NIT: probably unintended.

PR #304, opened 2018-11-16
Doc improvements

alexott:
This PR includes several things:

- attempt to refer to the API doc of the same version, not the latest
- expand some topics, like, use of batches, etc.
- add links to API classes
- fix small typos/errors


doc/faq/README.md
```diff
@@ -2,7 +2,7 @@
 
 ### Which versions of Cassandra does the driver support?
 
-The latest version of the driver supports any Cassandra version greater than 1.2.
+The latest version of the driver supports any Cassandra version starting with 1.2.
```
tolbertam: In our compatibility matrix we don't list C* 1.2 or 2.0 any more since they are considered EOL'd, so maybe we should just change this to 2.1+, even if C* 1.2 happens to work we aren't testing against it specifically this release.

jorgebay:
It's looking good! I'll test it locally with documentor and give my +1.

The only issue with relative links to API docs, e.g., `/api/module.metadata/`,  is that they will not work when viewing the docs from GitHub IU, but maybe that's not that important compared to linking to the same version within the generated docs (and not having weird version jumps like you mentioned).

alexott:
I can convert them into relative links - it's not a big task... I thought that document adjust them...

jorgebay:
> The only issue with relative links to API docs, e.g., /api/module.metadata/, is that [...]

I meant relative to the driver doc home, as you propose in this patch (as opposed to adding the absolute url as in `https://docs.datastax.com/en/developer/nodejs-driver/3.5/api/...`

PR #305, opened 2018-11-19
Randomize contact points to prevent hotspots

tolbertam:
For [NODEJS-104](https://datastax-oss.atlassian.net/browse/NODEJS-104).


tolbertam:
Awesome, thanks for reviewing and adding a test!

PR #309, opened 2018-12-07
Fix performance regression in 4.x

jorgebay:
- Invoke the user callback in the next tick to allow stack unwinding.
- Revert changes in `StreamIdStack` #301. The host in-flight count is now the sum of all the `inUse` streamId in all connections.

With this change, performance seems up to par with 3.6.0:

### Write throughput
boxplot-write-throughput

### Read throughput
boxplot-read-throughput

### DIFF
--- a/lib/host-connection-pool.js
+++ b/lib/host-connection-pool.js
@@ -49,7 +49,6 @@ class HostConnectionPool extends events.EventEmitter {
     this._newConnectionTimeout = null;
     this._creating = false;
     this._state = state.initial;
-    this.inFlight = 0;
     this.responseCounter = 0;
     this.options = host.options;
     this.protocolVersion = protocolVersion;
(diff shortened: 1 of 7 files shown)


lib/request-execution.js
```diff
@@ -102,49 +103,50 @@ class RequestExecution {
   }
 
   _sendOnConnection() {
-    const self = this;
-    const startTime = process.hrtime();
+    this._startTime = process.hrtime();
```
jorgebay: I didn't find enough evidence that this refactor (use an instance member instead of a closure) has an impact but in any case, having `this` context is less error prone.

lib/request-handler.js
```diff
@@ -234,7 +240,15 @@ class RequestHandler {
           this.request.query || 'batch'));
       }, this);
     }
-    this._callback(null, result);
+
+    // Invoke the callback in the next tick allowing stack unwinding, that way we can continue
```
tolbertam: Excellent explanation!  Will be nice to have for context in the future.

lib/host-connection-pool.js
```diff
@@ -78,6 +77,19 @@ class HostConnectionPool extends events.EventEmitter {
     });
   }
 
+  getInFlight() {
```
tolbertam: I agree that changing this from passively managing a running count to recalculating on every invocation is the right thing to do because:

1. This method isn't called anywhere by the driver itself, so unless someone uses it explicitly, the cost is minimal.  We expect it to be used for host selection (choose least busy host), so it may happen on each request but it's still probably cheaper than managing the running count (since something needed to be incremented/decremented previously to manage it anyways).
2. If you only have 1 connection per host (which should be typical), we don't need to sum things up.

PR #310, opened 2018-12-21
Send driver name and driver version in the STARTUP message

jorgebay:
https://datastax-oss.atlassian.net/browse/NODEJS-491

This patch adds the following functionality:
- Support for CASSANDRA-14275: Driver information in the STARTUP message
- Expose policy level options to improve observability and allow monitoring tools to extract the policy parameter information.
- Expose the control connection resolved hosts as an internal method.

### DIFF
--- a/lib/control-connection.js
+++ b/lib/control-connection.js
@@ -843,7 +843,7 @@ ControlConnection.prototype.reset = function (callback) {
 };
 
 /**
- * Gets a Map containing the original contact points and
+ * Gets a Map containing the original contact points and the addresses that each one resolved to.
  */
 ControlConnection.prototype.getResolvedContactPoints = function () {
   return this._resolvedContactPoints;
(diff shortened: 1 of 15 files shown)


lib/control-connection.js
```diff
@@ -822,6 +842,13 @@ ControlConnection.prototype.reset = function (callback) {
   });
 };
 
+/**
+ * Gets a Map containing the original contact points and
```
jorgebay:
```suggestion
 * Gets a Map containing the original contact points and the addresses that each one resolved to.
```

lib/requests.js
```diff
@@ -246,21 +247,33 @@ class PrepareRequest extends Request {
...
+    this.options = options || {};
   }
 
   write(encoder, streamId) {
     const frameWriter = new FrameWriter(types.opcodes.startup);
+
     const startupOptions = {
-      CQL_VERSION: this.cqlVersion
+      CQL_VERSION: this.options.cqlVersion || '3.0.0',
+      DRIVER_NAME: packageInfo.description,
```
tolbertam: nice idea pulling this from package.json :)

PR #312, opened 2019-01-14
Test fix: await for promise

jorgebay:


jorgebay:
I'll include other test fixes and submit a new pull request.

PR #313, opened 2019-01-15
Test fixes

jorgebay:
I've included fixes for integration tests that we see failing often and for tests that are likely to be affected by upcoming versions of C* and DSE.

I've replaced two pool tests that were originally using 3-node ccm clusters with simulacron-based tests.

Also, I've replaced brittle tests that were using query traces to assert the correct replicas with fixed key-to-replicas maps (based on the default tokens and partitioner).


jorgebay:
Also, build time goes from 25m to 23m on Jenkins and from 43m to 38m on AppVeyor :)

PR #316, opened 2019-02-20
Defer use of host map until metadata retrieval

jorgebay:
https://datastax-oss.atlassian.net/browse/NODEJS-500
https://datastax-oss.atlassian.net/browse/NODEJS-501

Several changes to the internal behaviour of the `ControlConnection`:

- Create initial connection without the notion of `Host`
- Small refactor for resolution and minor improvements
- Include `hostId` in `Host` metadata
- Integration tests using simulacron to validate reconnection behaviour

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -435,7 +435,6 @@ Connection.prototype.changeKeyspace = function (keyspace, callback) {
 /**
  * Prepares a query on a given connection. If its already being prepared, it queues the callback.
  * @param {String} query
- * @param {String} keyspace
  * @param {function} callback
  */
 Connection.prototype.prepareOnce = function (query, callback) {
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -30,14 +30,34 @@ let net = require('net');
 function Connection(endpoint, protocolVersion, options) {
   events.EventEmitter.call(this);
   this.setMaxListeners(0);
-  if (!endpoint || endpoint.indexOf(':') < 0) {
+
+  if (!options) {
+    throw new Error('options is not defined');
+  }
+
+  /**
+   * Gets the ip and port of the server endpoint.
+   * @type {String}
+   */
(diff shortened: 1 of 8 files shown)


lib/control-connection.js
```diff
@@ -75,102 +77,118 @@ function ControlConnection(options, profileManager, context) {
...
     }
   }
 
-  utils.series([
-    function resolveNames(next) {
-      utils.each(self.options.contactPoints, function eachResolve(name, eachNext) {
-        if (name.indexOf('[') === 0 && name.indexOf(']:') > 1) {
-          // IPv6 host notation [ip]:port (RFC 3986 section 3.2.2)
-          const index = name.lastIndexOf(']:');
-          return addHost(name.substr(1, index - 1), name.substr(index + 2), name, true, eachNext);
```
tolbertam: To make sure I understand right:

* Previously, every resolved name for a contact point would create a `Host` entry in `HostMap`.
* With this change, `Host`s will only be created when resolving data from the system.local and peers tables, meaning we will only see `Host` entries for what C* nodes report are nodes.

This makes complete sense to me.  I always thought it was weird that we created hosts for contact points (most other drivers do this as well).

lib/connection.js
```diff
@@ -407,6 +435,7 @@ Connection.prototype.changeKeyspace = function (keyspace, callback) {
 /**
  * Prepares a query on a given connection. If its already being prepared, it queues the callback.
  * @param {String} query
+ * @param {String} keyspace
```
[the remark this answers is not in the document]
jorgebay: yes, thanks for catching that... It's an inconsistency between C* and DSE drivers.
Protocol DSE v2 supports sending the keyspace and C* protocol v4 doesn't.
I'll remove it from here.

lib/control-connection.js
jorgebay: yes! that's correct.

PR #321, opened 2019-03-20
Support empty buffers as result cell values

jorgebay:
https://datastax-oss.atlassian.net/browse/NODEJS-475

When an empty buffer is obtained from C*, it returns `null` for most types except for ones that are supported: text types, `blob` and `custom`.

It basically generalizes the solution in #294.


gim-m:
Awesome! Thanks for the fix Jorge! 👍

PR #329, opened 2019-06-29
Adds ability to read payload of SUPPORTED response 

SoyYoRafa:
Adds ability to read payload of SUPPORTED response. It and fetches supported options by the server prior to sending startup message. This enables for a future compression PR.

### DIFF
--- a/lib/readers.js
+++ b/lib/readers.js
@@ -194,6 +194,24 @@ FrameReader.prototype.readBytesMap = function () {
 };
 
 /**
+ * Reads an associative array of strings as keys and string lists as values
+ * @returns {Object}
+ */
+FrameReader.prototype.readStringMultiMap = function () {
+  //A [short] n, followed by n pair <k><v> where <k> is a
+  //[string] and <v> is a [string[]].
+  const length = this.readShort();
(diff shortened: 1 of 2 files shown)


SoyYoRafa:
Hi, I will be addressing the failed tests in control-connection-simulator-tests.js and adding tests in the next coming days. I don't know what the datastax job-creator failure is since I can't access it.

jorgebay:
Hi @SoyYoRafa ! Thanks for the patch, implementation looks clean and straightforward.

Currently in the DataStax drivers we don't have a real use case for the OPTIONS/SUPPORTED messages (even for compression, you can always map supported compression algorithms for protocol version). For that reason, I don't think we need to send options before STARTUP message for now, even if the protocol allows it, as it increase the time it takes for a connection to be open.

I think we can leave the logic in `readers.js` and `streams.js`, for callers to use `OPTIONS` message in the future, if needed, and discard the changes in `connection.js`, to make it a less disruptive change and merge it, sounds good?

SoyYoRafa:
Hi, sounds good. I have reset the version for `connection.js` and pushed. All tests pass.

PR #330, opened 2019-07-11
Fix for PreparedQueries.prototype._validateOverflow

BryanDonovan:
Hi,

The ```PreparedQueries#_validateOverflow``` function has two bugs that can lead to a memory leak:
1. The comparison operator when comparing keys to this._maxPrepared should be ```>=```, not ```<```.
2. The key used to delete the entry from ```this._mapByKey``` should be generated using ```this._getKey()``` instead of using the query directly.


BryanDonovan:
No problem!  In case anyone comes across this pull request, I should note that this bug really only manifests when you exceed the limit of cached prepared queries.  In our case, we had a bug where we were preparing queries that shouldn't have been, e.g, queries with `WHERE IN` clauses that had a variable number of parameters, which resulted in thousands of prepared queries.

PR #336, opened 2019-09-16
TypeScript declaration files

jorgebay:
Add TypeScript declaration files to the repo and include an index declaration file.

https://datastax-oss.atlassian.net/browse/NODEJS-464

TODO: 

- [x] Test generating members in a javascript script that is later compiled by tsc.
- [x] Add more manual compilation tests
- [x] Build TypeScript tests in CI

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ test/localConfig.json
 out/
 /docs
 **/typescript/*.js
+**/typescript/*.js.map
+**/typescript/generated.ts
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ test/localConfig.json
 .DS_Store
 out/
 /docs
+**/typescript/*.js
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,40 @@
 language: node_js
-node_js:
-  - "10"
-  - "8"
-  - "6"
-  - "4"
 sudo: false
 cache:
   directories:
     - node_modules
-script:
-  - npm install -g eslint@4; npm run eslint;
-  - npm test
+
+jobs:
+  include:
+    - stage: "tests"
+      name: "Run eslint"
+      script: "npm install -g eslint@4; npm run eslint;"
+      node_js: "10"
+    - name: "TypeScript 2.9 generation and compilation tests"
+      node_js: "10"
+      script:
+        - npm install -g typescript@2.9;
+        - pushd test/unit/typescript/
+        - tsc -p .
+        - node -e "require('./api-generation-test').generate()" > generated.ts
+        - tsc generated.ts
+    - name: "Unit tests - Node.js 10"
+      node_js: "10"
+      script: "npm test"
+    - name: "Unit tests - Node.js 8"
+      node_js: "8"
+      script: "npm test"
+    - name: "Unit tests - Node.js 6"
+      node_js: "6"
+      script: "npm test"
+    - name: "Unit tests - Node.js 4"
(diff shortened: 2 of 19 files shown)


jorgebay:
@Svjard @pc-jedi @michal-b-kaminski 
We've decided to include TypeScript definitions for the driver in the project and keep them up to date for each release.

I wanted to thank you for all your work maintaining the community definitions at DefinitelyTyped. I've learned a LOT from your work. I would like to include a note thanking you in the upcoming DataStax dev blog post, including your names and handles, let me know if that's OK with you.

Also, feel free to review this PR and/or contributing to the DataStax driver!

pc-jedi:
Nice to here that you are including definitions directly in the repo. :slightly_smiling_face:
This will help new users to simpler get started.
Will you also ad some examples in how a common scenarios will look like in TypeScript?

From my site it is OK to be mentioned.

Regarding the PR, I'll hopefully have some time in the next days to do a full review on it.

Do you have any plans on how you can guaranty that the definitions are always aligned with the js-source? A common issue is that an update of the definitions is forgotten in a PR.

jorgebay:
> Do you have any plans on how you can guaranty that the definitions are always aligned with the js-source?

Good point, the idea is to include a js script as a test that generates properties into a ts file and see if it compiles, e.g. iterate through the dataTypes and generate a check per each.

jorgebay:
Thanks Michał!

>  many types are exported as interfaces, so they can not be derived from or instantiated in TS code

That's intended, most of those js prototypes are part of metadata and that should be exposed but not instantiated. In the same direction of other strongly typed languages, those interfaces can be mocked if needed by the user but not directly instantiated.

michal-b-kaminski:
What about `LoadBalancingPolicy` - the documentation at https://docs.datastax.com/en/developer/nodejs-driver/4.1/features/tuning-policies/#implementing-a-custom-load-balancing-policy shows an example, where custom load-balancing policy derives base `LoadBalancingPolicy`? Of course you can work around it being an interface in TS, but I think if it is designed to be derived from in JS, it should be the same in TS.

jorgebay:
hm, I see what u mean. It even has a default logic, like an abstract class: https://github.com/datastax/nodejs-driver/blob/master/lib/policies/load-balancing.js#L41-L53

Make sense to change to a `class`, I'll change it! thanks 👍

jorgebay:
I've pushed a commit to make `LoadBalancingPolicy` an abstract class, I'll push the tests by tomorrow.

jorgebay:
It's ready to be reviewed.

I'll add a couple of more manual compilation tests before eod.

jorgebay:
I've added auto-generation API tests that are currently building and compiling typescript sources in TravisCI: https://travis-ci.com/datastax/nodejs-driver/builds/128315662

jorgebay:
Besides the TypeScript tests included in the repo, I've tested it with a sample project coming from the `@types` definitions and it works (I've switched to `strict: true` and added some fixes lately).

I'll merge this one and start the process of releasing v4.2.0.

You probably didn't have the time to do a full review, thanks in any case. Feel to contribute to the TypeScript declarations or to the rest of the driver codebase.

jorgebay:
Blog post with acknowledgements :)

https://academy.datastax.com/content/typescript-support-datastax-nodejs-drivers

We have some bug reports on the first release, hopefully those are not a showstopper for ts users: https://datastax-oss.atlassian.net/browse/NODEJS-556?jql=labels%20%3D%20%22typescript%22

PR #337, opened 2019-09-17
Query operator contains

koriwi:
had this changing lying around in my fork for quite some time.
I need to use this every day :)

### DIFF
--- a/lib/mapping/q.js
+++ b/lib/mapping/q.js
@@ -76,6 +76,8 @@ class QueryAssignment {
  * Contains functions that represents operators in a query.
  * @alias module:mapping~q
  * @type {Object}
+ * @property {function} contains Represents the CQL operator "CONTAINS".
+ * @property {function} containsKey Represents the CQL operator "CONTAINS KEY".
  * @property {function} in_ Represents the CQL operator "IN".
  * @property {function} gt Represents the CQL operator greater than ">".
  * @property {function} gte Represents the CQL operator greater than or equals to ">=" .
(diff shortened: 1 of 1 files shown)


jorgebay:
oh good one, thanks for the patch!

I've created a ticket to track this: https://datastax-oss.atlassian.net/browse/NODEJS-553

I would like to include it in the upcoming release (in the next few days / week), we should add a unit test for this feature in `test/unit/mapping/model-mapper-select-tests.js`.

Could you add one or two tests? Otherwise, I'll try to do it myself.

jorgebay:
Now that I think about it (after your post in the mailing list), I think supporting `CONTAINS` operator without secondary indexes doesn't make much sense, the only other way to use `CONTAINS` in CQL outside secondary indexes is using `ALLOW FILTERING` queries that we explicitly don't want to support.

From the user perspective, I would recommend to use [custom queries](https://docs.datastax.com/en/developer/nodejs-driver/latest/features/mapper/queries/#custom-queries) on the mapper.

About the contribution, I would say we can focus in supporting indexes in the Mapper w/ this operators. As I mentioned in the mailing list, its a meaty feature that requires several unit and integration tests to make sure we properly support it.

koriwi:
Cool. As i said in the mailing list, i will try to implement the needed test for secondary indexes. I can just update this pull request so we have on pull request which contains CONTAINS, CONTAINS KEY and secondary index support :)
Got some open source time from my company

jorgebay:
glad to hear!

yeah, let's keep this one open.

absurdfarce:
Closing this out due to lack of activity.  If anybody wants to pursue these changes going forward please open a new ticket.

Thanks!

PR #338, opened 2019-09-26
feat(ts): add generic to mapper

koriwi:
may i propose these changes?
I modified the DefinetlyTyped typings before and used them with 100% success.
Tried to port them to your official typings now.

### DIFF
--- a/lib/mapping/index.d.ts
+++ b/lib/mapping/index.d.ts
@@ -148,7 +148,7 @@ export namespace mapping {
       query: string,
       paramsHandler: (doc: any) => any[],
       executionOptions?: string | MappingExecutionOptions
-    ): (doc: any, executionOptions?: string | MappingExecutionOptions) => Promise<Result>;
+    ): (doc: any, executionOptions?: string | MappingExecutionOptions) => Promise<Result<T>>;
   }
 
   namespace q {
--- a/lib/mapping/index.d.ts
+++ b/lib/mapping/index.d.ts
@@ -128,7 +128,7 @@ export namespace mapping {
     update(doc: any, docInfo?: UpdateDocInfo): ModelBatchItem;
   }
 
-  interface ModelMapper<T> {
+  interface ModelMapper<T = { [key: string]: any }> {
     name: string;
     batching: ModelBatchMapper;
 
--- a/lib/mapping/index.d.ts
+++ b/lib/mapping/index.d.ts
@@ -43,10 +43,10 @@ export namespace mapping {
     newObjectInstance(): any;
   }
 
-  interface Result<T = { [key: string]: any }> extends Iterator<T> {
+  interface Result<T = any> extends Iterator<T> {
     wasApplied(): boolean;
 
-    first(): T;
+    first(): T | null;
 
     forEach(callback: (currentValue: T, index: number) => void, thisArg?: any): void;
 
@@ -72,7 +72,7 @@ export namespace mapping {
 
     batch(items: ModelBatchItem[], executionOptions?: string | MappingExecutionOptions): Promise<Result>;
 
-    forModel<T = { [key: string]: any }>(name: string): ModelMapper<T>;
+    forModel<T = any>(name: string): ModelMapper<T>;
   }
 
   type MappingOptions = {
@@ -128,7 +128,7 @@ export namespace mapping {
     update(doc: any, docInfo?: UpdateDocInfo): ModelBatchItem;
   }
 
-  interface ModelMapper<T = { [key: string]: any }> {
+  interface ModelMapper<T = any> {
     name: string;
     batching: ModelBatchMapper;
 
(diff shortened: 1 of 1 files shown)


jorgebay:
I think its a great idea!

To prevent breaking changes, we should introduce new interfaces instead of modifying existing ones.

Introduce a new `ModelMapperGeneric<T>` (nvm the name), that contains all the properties:

```typescript
interface ModelMapperGeneric<T> {
  get(doc: T, docInfo?: { fields?: string[] }, executionOptions?: string | MappingExecutionOptions): Promise<T>;
  //...
}
```

Make `ModelMapper` extend `ModelMapperGeneric<any>`:

```typescript
interface ModelMapper extends ModelMapperGeneric<any> {
}
```

Create an overload of `forModel()` (one generic and the other leave as is):

```typescript
  forModel(name: string): ModelMapper;

  forModel<T>(name: string): ModelMapperGeneric<T>;
```

lib/mapping/index.d.ts
```diff
@@ -128,27 +128,27 @@ export namespace mapping {
     update(doc: any, docInfo?: UpdateDocInfo): ModelBatchItem;
   }
 
-  interface ModelMapper {
+  interface ModelMapper<T = { [key: string]: any }> {
```
jorgebay: I think we should continue to use `any` that is less restrictive than `{ [key: string]: any }`.

jorgebay:
I've created a ticket to track this: https://datastax-oss.atlassian.net/browse/NODEJS-562

lib/mapping/index.d.ts
koriwi: I can't think of any case where you need any.
If you want backwards compatibility, we can just change it to:
`interface ModelMapper<T = any }> {`

koriwi:
Have you read my answer about `any` or `{ [key: string]: any }` ?

lib/mapping/index.d.ts
jorgebay: I prefer not to be restrictive, users can use it for whatever type they choose, but tbh I don't know which cases are not going to be covered by `T = { [key: string]: any }`

koriwi: We should use `T = { [key: string]: any }` because the driver always returns null or an object for .get() and equivalent for .find(). 
We are not being restrictive by not using `any`. We are just doing what typescript is intended for in my opinion :)

jorgebay: I think its OK to use `{ [key: string]: any }`, my concern is that it was already implemented using `any`, so restricting it now might be a breaking change.

Let's continue addressing the other issues (like `ModelMapperGeneric<T>`) and come back to this one if needed.

koriwi: ModelMapperGeneric and this point is the same in my opinion.
If we use `any` it will be fully backwards compatible and can keep ModelMapper as the only interface. Best solution right now I think.

jorgebay: Backwards compatibility is very important as a library.

Note that if somebody is using the following:

```typescript
const modelMapper: ModelMapper = mapper.forModel(x);
```

With this new change, it will break.

Would it be possible to provide additional generic interfaces and make the existing ones work as currently do?

lib/mapping/index.d.ts
```diff
@@ -128,27 +128,27 @@ export namespace mapping {
...
     name: string;
     batching: ModelBatchMapper;
 
-    get(doc: any, docInfo?: { fields?: string[] }, executionOptions?: string | MappingExecutionOptions): Promise<any>;
+    get(doc: { [key: string]: any }, docInfo?: { fields?: string[] }, executionOptions?: string | MappingExecutionOptions): Promise<null | T>;
```
jorgebay: I've noticed that this you are using the `null` type here. If we use it here, we should use it in `Result<T>#first()` method.

I personally can't wrap my head about `Type | null` in ts definitions. With strict null checking, this succeeds:

```typescript
function get(): Video {
  // ...
}

const o: Video = get();
// this condition compiles (!)
if (o === null) {

}
```

lib/mapping/index.d.ts
koriwi: this is the least I can do to contribute to this great library!

lib/mapping/index.d.ts
koriwi: I dont see where you have enabled strict null checking right now.
IIRC there is no tsconfig.json

koriwi: > I've noticed that this you are using the `null` type here. If we use it here, we should use it in `Result<T>#first()` method.
> 
Yes. I will change that :)

jorgebay: > I dont see where you have enabled strict null checking right now.

I mean as a general concept (not related to the patch), it feels weird for me to allow `o === null` when strictNullCheck is on (and `o` is not nullable)

jorgebay:
Landed in a37b9a7d560112ae357b548cb3904d86a56104a0, thanks!

PR #340, opened 2019-10-07
Use proxyquire, sinon and chai

jorgebay:
Replaced `rewire` (not maintained) with `proxyquire`.

Added `sinon` for fake timers, stubs and spies and `chai` to simplify assertions.


jorgebay:
Dev dependencies tree went from 138 to 51.

Production dependencies tree still at 3.

PR #341, opened 2019-10-08
Fix eachRow types to use ValueCallback

shinyaohira:
Hi,
Currently EmptyCallback is used for eachRow's callback type, but it should be ValueCallback.

from [jsdoc](https://github.com/datastax/nodejs-driver/blob/master/lib/client.js#L573)
> callback Executes callback(err, result)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -57,16 +57,16 @@ export class Client extends events.EventEmitter {
 
   execute(query: string, callback: ValueCallback<types.ResultSet>): void;
 
-  eachRow<T>(query: string,
+  eachRow(query: string,
           params: ArrayOrObject,
           options: QueryOptions,
           rowCallback: (n: number, row: types.Row) => void,
-          callback?: ValueCallback<T>): void;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -57,16 +57,16 @@ export class Client extends events.EventEmitter {
 
   execute(query: string, callback: ValueCallback<types.ResultSet>): void;
 
-  eachRow(query: string,
+  eachRow<T>(query: string,
           params: ArrayOrObject,
           options: QueryOptions,
           rowCallback: (n: number, row: types.Row) => void,
-          callback?: EmptyCallback): void;
+          callback?: ValueCallback<T>): void;
```
jorgebay: It should be `ValueCallback<types.ResultSet>`

shinyaohira:
@jorgebay Thank you!

shinyaohira:
@jorgebay when will this get published to npm?

jorgebay:
Most likely today :)

PR #344, opened 2019-10-26
Fixed MutableLong multiplication edge case that caused infinite recursion

anoek:
When performing a MutableLong multiplication, if either of the multiplier or multiplicand was `-9223372036854775808` (the smallest possible value), the multiplication method would infinitely recurse because when you negate the value you get the same (negative) value.  This patch fixes the problem by removing the recursion from the multiplication method. The overflow characteristics now match that of the Java implementation of this library, and should be slightly faster since we've eliminated a separate function call when dealing with negative values.

Tests have been added to validate the expected output of the multiplication in this boundary case. 

A test has also been added for the tokenizer hash method to validate the resulting hash on a previously problematic sequence (namely the byte sequence for `-9223372036854775808`), instead of crashing, the hash value now matches the expected hash value obtained by executing the equivalent Java tokenizer.

All tests added failed prior to this patch, and pass now.


jorgebay:
Thanks @anoek !

PR #346, opened 2019-11-05
Fix deprecated IdempotencyAwareRetryPolicy

jorgebay:
The policy, which is deprecated since 4.0, had a bug where it checked
the `options` instead `executionOptions` (introduced in 4.0).


jorgebay:
Wrong branch name 🤦‍♂

PR #349, opened 2019-12-12
[Performance] Increase unsafe buffer pool size

puzpuzpuz:
Writer objects in `nodejs-driver` seem to do quite a lot of `Buffer.allocUnsafe` calls. As you may know, `Buffer.allocUnsafe` uses a global pool (strictly speaking, it's not a "classical" object pool, but let's use this terminology), i.e. pre-allocated internal `Buffer` which is then sliced for newly allocated buffers, when possible. The size of the pool may be configured by setting `Buffer.poolSize` property (8KB by default). Once the current internal buffer fills up, a new one is allocated by the pool.

The problem with this approach is that this pool is global, so its size can't be configured by a Node.js library. Of course, a library can change `Buffer.poolSize`, but that may clash with another library or the final application which may also change it.

This PR migrates `Buffer.allocUnsafe` calls to [`nbufpool`](https://github.com/puzpuzpuz/nbufpool) library, which is just a port of the built-in pool to user-land. The pool has the same performance as the built-in pool when used with the default 8KB size. And when it's used with a more sane size (2MB) it's [significantly faster](https://github.com/puzpuzpuz/nbufpool#benchmark-results).

I'm not able to proof performance improvements in the library as I failed to find public available benchmarks. If you point me at them, I can include benchmark results.

P.S. If you don't like the idea of adding another 3-rd party library into dependencies, I have nothing against embedding it as a part of `nodejs-driver`'s code base.

P.P.S. Off-topic (as there are no GH issue in this repo, I'm writing them here):
It's better to do the following enhancements in the library:
* Add `package-lock.json` to `.gitignore`
* Freeze dependency versions in `package.json` (avoid using wildcards for minor versions - `^`)

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -19,6 +19,7 @@ const util = require('util');
 const net = require('net');
 const EventEmitter = require('events');
 const errors = require('./errors');
+const Pool = require('nbufpool');
 
 /**
  * Max int that can be accurately represented with 64-bit Number (2^53)
@@ -30,6 +31,10 @@ const maxInt = 9007199254740992;
 const maxInt32 = 0x7fffffff;
 
 const emptyObject = Object.freeze({});
+
+const unsafePoolSize = 2097152; // 2MB
+
+const unsafePool = new Pool(unsafePoolSize);
 
 function noop() {}
 
@@ -43,7 +48,7 @@ const allocBuffer = Buffer.alloc || allocBufferFillDeprecated;
  * Forward-compatible unsafe allocation of buffer.
  * @type {Function}
  */
-const allocBufferUnsafe = Buffer.allocUnsafe || allocBufferDeprecated;
+const allocBufferUnsafe = Buffer.allocUnsafe ? unsafePool.allocUnsafe.bind(unsafePool) : allocBufferDeprecated;
 
 /**
  * Forward-compatible allocation of buffer to contain a string.
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
     "long": "^2.2.0",
     "adm-zip": "^0.4.13",
     "@types/node": ">=4",
-    "@types/long": "^4.0.0"
+    "@types/long": "^4.0.0",
+    "nbufpool": "^0.1.0"
   },
   "devDependencies": {
     "mocha": "~5.2.0",


jorgebay:
Interesting!

We have a benchmarking suite of the driver at https://github.com/riptano/nodejs-driver-sut . The benchmarks should not be run on a dev laptop / computer. If you don't have the computing resources available, I can run them early next week.

I imagine the results may vary based on the workload, the type of application and the environment.

nbufpool library is designed to run against Node.js 6+ and this library runs with Node.js 4+. We are [dropping support for 4 and 6](https://datastax-oss.atlassian.net/browse/NODEJS-516) soon but not yet.

I think it can be valuable to allow users to plugin their own buffer pool.

We use JIRA for issue tracking: https://datastax-oss.atlassian.net/projects/NODEJS/issues

puzpuzpuz:
@jorgebay
Thanks for the positive feedback!

> The benchmarks should not be run on a dev laptop / computer. If you don't have the computing resources available, I can run them early next week.

Unfortunately, I only have access to my laptop and desktop machines. So, it would be great if you could run the benchmark.

> I imagine the results may vary based on the workload, the type of application and the environment.

Yes, the main gain in throughput should be seen in scenarios with lots of writes with something near 1KB payload size (it's fine if it varies by an order of magnitude). In benchmark run that I did for [Hazelcast Node.js client](https://github.com/hazelcast/hazelcast-nodejs-client) I got around +10% throughput improvement in such scenario. For really small (3B) or big (100KB) payloads it wasn't noticeable.

> nbufpool library is designed to run against Node.js 6+ and this library runs with Node.js 4+. We are dropping support for 4 and 6 soon but not yet.

It should work fine on Node v4, as there was the fallback to `new Buffer()` originally and I kept it (see https://github.com/datastax/nodejs-driver/pull/349/files#diff-50e3aa130a4f97a42ee2cf111c7b1d9dR51).

> I think it can be valuable to allow users to plugin their own buffer pool.

Under certain conditions misbehaving pool, provided by the user, may ruin correctness of `nodejs-driver` (i.e. it may return buffers pointing at the same portion of off-heap memory), so I feel like this optimization shouldn't be exposed externally. WDYT?

jorgebay:
It looks like there isn't a noticeable difference in throughput.

Here are the results: 

throughput-compare

These are the benchmark parameters: 
- baseline: master
- branch: feature/increase-unsafe-buffer-pool-size
- workload: standard
- nodejs runtime: 12

puzpuzpuz:
@jorgebay
Thanks for doing the experiment! What was the average payload size?

As for the result, I guess that `Buffer.allocUnsafe` doesn't become the bottleneck on this volume requests per second. I'll probably try to profile it on my dev machine to understand the workload.

puzpuzpuz:
Yesterday I run modified benchmark: 1KB string write was added into `standardwrite` workload with `Buffer.poolSize = 2097152;`. My goal was to understand if the `Buffer.allocUnsafe` with default pool size is a critical bottleneck for the driver library. I saw improvements in a few scenarios, but they were minor (~5%).

That's why I'm closing this PR as not valuable.

jorgebay:
Thanks for benchmarking that scenario.

In any case, nbufpool project makes a lot of sense. A dedicated buffer pool per library, based on its own needs is a common pattern in other languages and runtimes, kudos!

PR #354, opened 2020-02-12
Expose localDc param in default lbp method and improve error messages

jorgebay:
https://datastax-oss.atlassian.net/browse/NODEJS-573
https://datastax-oss.atlassian.net/browse/NODEJS-564

Expose localDc parameter in default lbp method and include available dcs in lbp init errors.

### DIFF
--- a/lib/policies/index.js
+++ b/lib/policies/index.js
@@ -40,10 +40,14 @@ exports.defaultAddressTranslator = function () {
 
 /**
  * Returns a new instance of the default load-balancing policy used by the driver.
+ * @param {string} [localDc] When provided, it sets the data center that is going to be used as local for the
+ * load-balancing policy instance.
+ * <p>When is localDc undefined, the load-balancing policy instance will use the <code>localDataCenter</code>
+ * provided in the {@link ClientOptions}.</p>
  * @returns {LoadBalancingPolicy}
  */
-exports.defaultLoadBalancingPolicy = function () {
-  return new loadBalancing.DefaultLoadBalancingPolicy();
+exports.defaultLoadBalancingPolicy = function (localDc) {
+  return new loadBalancing.DefaultLoadBalancingPolicy(localDc);
 };
 
 /**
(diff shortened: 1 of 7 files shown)


lib/policies/index.js
```diff
@@ -40,10 +40,14 @@ exports.defaultAddressTranslator = function () {
 
 /**
  * Returns a new instance of the default load-balancing policy used by the driver.
+ * @param {string} [localDc] When provided, it sets the data center that is going to be used as local for the
+ * load-balancing policy instance.
+ * <p>When is localDc undefined, the load-balancing policy instance will use the <code>localDataCenter</code>
```
adutra:
```suggestion
 * <p>When localDc is undefined, the load-balancing policy instance will use the <code>localDataCenter</code>
```

lib/policies/load-balancing.js
```diff
@@ -820,30 +820,42 @@ class DefaultLoadBalancingPolicy extends LoadBalancingPolicy {
...
   if (!(lbp instanceof LoadBalancingPolicy)) {
     throw new errors.DriverInternalError('LoadBalancingPolicy instance was not provided');
   }
 
   if (client && client.options) {
     if (lbp.localDc && !client.options.localDataCenter) {
-      client.log('info', 'Local data center \'' + lbp.localDc + '\' was provided as an argument to' +
-        ' DCAwareRoundRobinPolicy. It is more preferable to specify the local data center using \'localDataCenter\'' +
+      client.log('info', 'Local data center \'' + lbp.localDc + '\' was provided as an argument to the' +
+        ' LoadBalancingPolicy. It is more preferable to specify the local data center using \'localDataCenter\'' +
```
adutra: Suggestion: "...was provided as an argument to the load-balancing policy. It is preferable to specify..."

jorgebay:
> what is the motivation in nodejs and c# drivers to introduce a programmatic way to set the local DC

The idea is to support users that want to specify one or more local DCs based on the execution profile but don't want to change any aspect of the load balancing algorithm. We simplify the decision of having to choose a load-balancing policy class, that way the user choose to "use whatever is default but with this dc".

For example:

```javascript
const client = new Client({
  contactPoints,
  localDataCenter, // local dc for all workloads except specified otherwise
  profiles: [
    new ExecutionProfile('search', {
      consistency: consistency.one,
      loadBalancing: policies.defaultLoadBalancingPolicy("search-dc1")
    })
  ]
});
```

jorgebay:
I've applied the suggestions, I'll wait for build to finish and merge it.

PR #360, opened 2020-05-06
Fix graph Direction serialization

jorgebay:
Expose `direction` object and fix Direction serialization. Similar to #359 .

https://datastax-oss.atlassian.net/browse/NODEJS-606

I went over all deserializers in TinkerPop JavaScript GLV just to make sure we're not missing anything and this kind of issue never occurs again. There are TinkerPop serializers for the following js types: `Number`, `Date`, `Traverser`, `Vertex`, `VertexProperty`, `Property`, `Edge`, `Path`, `T`, `Direction`, `Array`, `BulkSet` and `Map` which, with this patch, are all now covered in the driver.

### DIFF
--- a/test/unit/typescript/graph-tests.ts
+++ b/test/unit/typescript/graph-tests.ts
@@ -39,4 +39,7 @@ async function myTest(client: Client): Promise<any> {
   let tokenString: string;
   tokenString = datastax.graph.t.id.toString();
   tokenString = datastax.graph.t.label.toString();
+  tokenString = datastax.graph.direction.in_.toString();
+  tokenString = datastax.graph.direction.out.toString();
+  tokenString = datastax.graph.direction.both.toString();
 }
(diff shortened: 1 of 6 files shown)


test/unit/typescript/graph-tests.ts
```diff
@@ -39,4 +39,7 @@ async function myTest(client: Client): Promise<any> {
   let tokenString: string;
   tokenString = datastax.graph.t.id.toString();
   tokenString = datastax.graph.t.label.toString();
+  tokenString = datastax.graph.direction.in_.toString();
```
adutra, line 42: What is the purpose of reassigning `tokenString` many times in a row without reading its value?

jorgebay: TypeScript transpiles to js, so execution is done in javascript using the same runtime (Node.js).

We create integration tests in JavaScript, so runtime correctness is covered. For the TypeScript part, we have to make sure the API have the expected form. These type of tests are fairly common for non-native TypeScript projects (typescript definitions), see https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0.1.450/types/cassandra-driver/cassandra-driver-tests.ts#L133-L137 for example

PR #362, opened 2020-06-30
Migrate to Jenkins pipelines

jorgebay:
https://datastax-oss.atlassian.net/browse/NODEJS-602

Includes a small refactor for spawning Apache Studio to test kerberos auth.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,19 @@
 #!groovy
 
 def initializeEnvironment() {
+  env.DRIVER_DISPLAY_NAME = 'Cassandra Node.js Driver'
+  if (env.GIT_URL.contains('riptano/nodejs-driver')) {
+    env.DRIVER_DISPLAY_NAME = 'private ' + env.DRIVER_DISPLAY_NAME
+  } else if (env.GIT_URL.contains('nodejs-dse-driver')) {
+    env.DRIVER_DISPLAY_NAME = 'DSE Node.js Driver'
+  }
+
+  env.GIT_SHA = "${env.GIT_COMMIT.take(7)}"
+  env.GITHUB_PROJECT_URL = "https://${GIT_URL.replaceFirst(/(git@|http:\/\/|https:\/\/)/, '').replace(':', '/').replace('.git', '')}"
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,630 @@
...
+      name: 'CI_SCHEDULE',
+      choices: ['DO-NOT-CHANGE-THIS-SELECTION', 'WEEKNIGHTS'],
+      description: 'CI testing schedule to execute periodically scheduled builds and tests of the driver (<strong>DO NOT CHANGE THIS SELECTION</strong>)')
+  }
+
+  triggers {
+    parameterizedCron("""
+      # Every weeknight (Monday - Friday) around 3:00 AM
+      H 3 * * 1-5 %CI_SCHEDULE=WEEKNIGHTS
+    """)
```
joao-r-reis, line 305: if you want to have nightlies on `master` only you need to add a check here (e.g. https://github.com/datastax/csharp-driver/blob/master/Jenkinsfile#L429)

note that nightly runs will occur even if no changes are pushed so if there is no check here, it will run a nightly build for every branch that the repo has if I'm not mistaken

Jenkinsfile
```diff
@@ -0,0 +1,630 @@
...
+
+pipeline {
+  agent none
+
+  // Global pipeline timeout
+  options {
+    timeout(time: 10, unit: 'HOURS')
+    buildDiscarder(logRotator(artifactNumToKeepStr: '10', // Keep only the last 10 artifacts
+                              numToKeepStr: '50'))        // Keep only the last 50 build records
+  }
```
joao-r-reis, line 209: on the old jenkins we only allowed one concurrent build per branch, if you want the same behavior you need to add `disableConcurrentBuilds()` here (e.g. https://github.com/datastax/csharp-driver/blob/master/Jenkinsfile#L396)

Jenkinsfile
jorgebay: Great catch! I only want to build `master` for weeknights .

Jenkinsfile
jorgebay: Done!

jorgebay:
> Multibranch Pipeline job configuration UI to prune branches

I couldn't find it in the UI... HALP!

PR #369, opened 2020-08-06
Support DSE auth transitional mode

jorgebay:
Introduce an internal auth provider, `NoAuthProvider` that supports transitional mode and handles auth flow when no credentials/provider where provided by the user.

https://datastax-oss.atlassian.net/browse/NODEJS-609

### DIFF
--- a/lib/auth/no-auth-provider.js
+++ b/lib/auth/no-auth-provider.js
@@ -1,0 +1,70 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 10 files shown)


lib/auth/no-auth-provider.js
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Internal authentication provider that is used when no provider has been set by the user.
+ * @ignore
+ */
+class NoAuthProvider extends AuthProvider {
+  newAuthenticator(endpoint, name) {
+    if (name === dseAuthenticator) {
+      // Try to use transitional mode
+      return new TransitionalModePlainTextAuthenticator();
```
joao-r-reis, line 33: So if the user does not set any auth configuration on the driver, it will support the transitional mode by default?

jorgebay: yes, as long as they are using `DseAuthenticator`.

PR #373, opened 2020-09-09
Error when using batching with custom mapping

leitethi:
When using batching with custom mappings the driver returns an error because mapping information is not passed to parameter getter in ModelBatchItem


jorgebay:
Nice catch @leitethi !

I will try to review it in the next days.

jorgebay (approved):
Thanks for the fix and including a unit test!

PR #374, opened 2021-02-19
Update README.md

alexleventer:
- Add instructions for token auth


absurdfarce:
Thanks for the pull request @alexleventer!  There is some new work being done in this space which should introduce a more reliable way to solve similar problems so I'm going to close this PR out in favor of that work.

PR #378, opened 2021-05-13
Re-resolve contact points after total disconnect

absurdfarce:
When we have to do a complete refresh of the control connection make sure to re-resolve contact points.  This looks to be the underlying issue for https://datastax-oss.atlassian.net/browse/NODEJS-632... it was also the subject of https://datastax-oss.atlassian.net/browse/NODEJS-577.

### DIFF
--- a/lib/control-connection.js
+++ b/lib/control-connection.js
@@ -130,18 +130,13 @@ class ControlConnection extends events.EventEmitter {
     // Use RFC 3986 for IPv4 and IPv6
     const standardEndpoint = !isIPv6 ? endpoint : `[${address}]:${portNumber}`;
 
-    const resolvedAddressedByName = this._getResolvedContactPoints(name);
-    resolvedAddressedByName.push(standardEndpoint);
-  }
-
-  _getResolvedContactPoints(name) {
-
-    let rv = this._resolvedContactPoints.get(name);
-    if (rv === undefined) {
-      rv = [];
-      this._resolvedContactPoints.set(name, rv);
+    let resolvedAddressedByName = this._resolvedContactPoints.get(name);
+    if (resolvedAddressedByName === undefined) {
+      resolvedAddressedByName = [];
+      this._resolvedContactPoints.set(name, resolvedAddressedByName);
     }
-    return rv;
+
+    resolvedAddressedByName.push(standardEndpoint);
   }
 
   async _parseContactPoint(name) {
--- a/lib/control-connection.js
+++ b/lib/control-connection.js
@@ -130,7 +130,7 @@ class ControlConnection extends events.EventEmitter {
     // Use RFC 3986 for IPv4 and IPv6
     const standardEndpoint = !isIPv6 ? endpoint : `[${address}]:${portNumber}`;
 
-    let resolvedAddressedByName = this._getResolvedContactPoints(name);
+    const resolvedAddressedByName = this._getResolvedContactPoints(name);
     resolvedAddressedByName.push(standardEndpoint);
   }
 
@@ -142,11 +142,6 @@ class ControlConnection extends events.EventEmitter {
       this._resolvedContactPoints.set(name, rv);
     }
     return rv;
-  }
-
-  async _clearResolvedContactPoints(name) {
-
-    this._getResolvedContactPoints(name).clear();
   }
 
   async _parseContactPoint(name) {
@@ -425,6 +420,26 @@ class ControlConnection extends events.EventEmitter {
     }
   }
 
+  async _refreshControlConnection(hostIterator) {
+
+    try {
+
+      this.connection = this._borrowAConnection(hostIterator);
+    }
+    catch(err) {
+
+      /* NODEJS-632: refresh nodes before getting hosts for reconnect since some hostnames may have
+       * shifted during the flight. */
+      this.log("info", "ControlConnection could not reconnect using existing connections.  Refreshing contact points and retrying");
+      this._contactPoints.clear();
+      this._resolvedContactPoints.clear();
+      await Promise.all(this.options.contactPoints.map(name => this._parseContactPoint(name)));
+      const refreshedContactPoints = Array.from(this._contactPoints).join(',');
+      this.log('info', `Refreshed contact points: ${refreshedContactPoints}`);
+      await this._initializeConnection();
+    }
+  }
+
   /**
    * Acquires a new connection and refreshes topology and keyspace metadata.
    * <p>When it fails obtaining a connection and there aren't any more hosts, it schedules reconnection.</p>
@@ -449,21 +464,7 @@ class ControlConnection extends events.EventEmitter {
         hostIterator = await promiseUtils.newQueryPlan(this._profileManager.getDefaultLoadBalancing(), null, null);
       }
 
-      try {
-
-        this.connection = this._borrowAConnection(hostIterator);
-      }
-      catch(err) {
-
-        /* NODEJS-632: refresh nodes before getting hosts for reconnect since some hostnames may have
-         * shifted during the flight. */
-        this.log("info", "ControlConnection could not reconnect using existing connections.  Refreshing contact points and retrying");
-        await Promise.all(
-          this.options.contactPoints.map(name =>
-            this._clearResolvedContactPoints(name)
-            .then(this._parseContactPoint(name)))
-        ).then(this._initializeConnection());
-      }
+      await this._refreshControlConnection(hostIterator);
     } catch (err) {
       // There was a failure obtaining a connection or during metadata retrieval
       this.log('error', 'ControlConnection failed to acquire a connection', err);
--- a/lib/control-connection.js
+++ b/lib/control-connection.js
@@ -130,13 +130,23 @@ class ControlConnection extends events.EventEmitter {
     // Use RFC 3986 for IPv4 and IPv6
     const standardEndpoint = !isIPv6 ? endpoint : `[${address}]:${portNumber}`;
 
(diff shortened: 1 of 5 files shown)


lib/control-connection.js
```diff
@@ -439,8 +449,21 @@ class ControlConnection extends events.EventEmitter {
         hostIterator = await promiseUtils.newQueryPlan(this._profileManager.getDefaultLoadBalancing(), null, null);
       }
 
-      this.connection = this._borrowAConnection(hostIterator);
+      try {
```
absurdfarce, line 452: This mirrors logic in init(), but in that function we only resolve contact point addresses if SNI is not in use.  Perhaps we should impose a similar constraint here?  It probably doesn't _hurt_ anything to do the resolution in all cases but it's not very useful in the SNI case.

lib/control-connection.js
```diff
@@ -439,8 +449,21 @@ class ControlConnection extends events.EventEmitter {
...
+      catch(err) {
+
+        /* NODEJS-632: refresh nodes before getting hosts for reconnect since some hostnames may have
+         * shifted during the flight. */
+        this.log("info", "ControlConnection could not reconnect using existing connections.  Refreshing contact points and retrying");
+        await Promise.all(
+          this.options.contactPoints.map(name =>
+            this._clearResolvedContactPoints(name)
+            .then(this._parseContactPoint(name)))
+        ).then(this._initializeConnection());
```
absurdfarce, line 465: After digging into this a bit I suspect this is doing more work than we need to.  We should be okay just getting a new connection here... a full initialize might be a bit much.

lib/control-connection.js
```diff
@@ -415,6 +415,26 @@ class ControlConnection extends events.EventEmitter {
     }
   }
 
+  async _refreshControlConnection(hostIterator) {
+
+    try {
+
+      this.connection = this._borrowAConnection(hostIterator);
+    }
+    catch(err) {
```
jdonenine: Does the type of error matter here at all?  Is there ever an error where we don't want the retry to happen?

lib/control-connection.js
```diff
@@ -415,6 +415,26 @@ class ControlConnection extends events.EventEmitter {
...
+
+      /* NODEJS-632: refresh nodes before getting hosts for reconnect since some hostnames may have
+       * shifted during the flight. */
+      this.log("info", "ControlConnection could not reconnect using existing connections.  Refreshing contact points and retrying");
+      this._contactPoints.clear();
+      this._resolvedContactPoints.clear();
+      await Promise.all(this.options.contactPoints.map(name => this._parseContactPoint(name)));
```
jdonenine: Does `_parseContactPoint()` stuff the result into `_contactPoints`?

jdonenine: Followup question, what happens if that fails? What's the end result in `_contactPoints` that we're going to use in the next line?

lib/control-connection.js
```diff
@@ -415,6 +415,28 @@ class ControlConnection extends events.EventEmitter {
...
+    else {
+      try { this.connection = this._borrowAConnection(hostIterator); }
+      catch(err) {
+
+        /* NODEJS-632: refresh nodes before getting hosts for reconnect since some hostnames may have
+         * shifted during the flight. */
+        this.log("info", "ControlConnection could not reconnect using existing connections.  Refreshing contact points and retrying");
+        this._contactPoints.clear();
+        this._resolvedContactPoints.clear();
+        await Promise.all(this.options.contactPoints.map(name => this._parseContactPoint(name)));
```
jdonenine, line 432: Does _parseContactPoint() stuff the result into _contactPoints?

Followup question, what happens if that fails? What's the end result in _contactPoints that we're going to use in the next line?

jdonenine (commented):
Just a few questions for you @absurdfarce , thanks!

lib/control-connection.js
```diff
@@ -415,6 +415,28 @@ class ControlConnection extends events.EventEmitter {
     }
   }
 
+  async _refreshControlConnection(hostIterator) {
+
+    if (this.options.sni) {
+      this.connection = this._borrowAConnection(hostIterator);
```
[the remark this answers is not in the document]
absurdfarce: SNI does endpoint resolution at connect time as part of the enhanced security it offers.  This change basically mirrors what's done when the connection is initially established... see https://github.com/datastax/nodejs-driver/blob/nodejs-632/lib/control-connection.js#L184-L200 for more on this.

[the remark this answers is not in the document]
jdonenine: Ah, so no reason to reresolve because it's going to do it all by itself in that case?

lib/control-connection.js
```diff
@@ -415,6 +415,28 @@ class ControlConnection extends events.EventEmitter {
...
+    }
+    else {
+      try { this.connection = this._borrowAConnection(hostIterator); }
+      catch(err) {
```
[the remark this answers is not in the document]
absurdfarce: There certainly could be errors here that we might not want to worry about... but it seems like that list is fairly small.  There's the obvious case of hostnames being out of date (which this PR is intended to handle) but don't forget that there's a general retry bundled into this logic.  So things like transient socket errors would also get another shot, which seems beneficial on the whole.

In short I guess I'd say more scenarios benefit from this than not so making this the default behaviour for most connection errors seems like a goodness... but there very well might be something I'm not thinking of there.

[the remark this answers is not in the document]
jdonenine: Yeah, that definitely sounds like a general benefit, I was more thinking are there scenarios where we should not expect to be able to reconnect, then we're wasting time.  But, this is just a single retry attempt, so even that is a fairly minimal downside.

lib/control-connection.js
absurdfarce: Yes, _parseContactPoint() calls _addContactPoint in various ways which winds up populating _contactPoints.

We're mirroring the logic that takes place in initialization again... see https://github.com/datastax/nodejs-driver/blob/nodejs-632/lib/control-connection.js#L186 specifically.  If we successfully resolve a hostname all the resulting IPs are added, otherwise we store the hostname along with an indicator saying resolution failed... see https://github.com/datastax/nodejs-driver/blob/nodejs-632/lib/control-connection.js#L160-L171 for the details.

lib/control-connection.js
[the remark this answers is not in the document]
absurdfarce: Right; if they're using SNI they'll need to resolve hostnames at connect-time.  That's why a resolver is built in the initialization process.  And by extension clearly any changes to DNS entries would be handled immediately in the SNI case by this mechanism.

jdonenine (approved):
Thanks for answering all my questions, 👍

test/unit/control-connection-tests.js
```diff
@@ -422,8 +422,9 @@ describe('ControlConnection', function () {
         yield this.delay;
       };
 
-      const cc = newInstance({ contactPoints: [ '::1' ], policies: { loadBalancing: lbp, reconnection: rp } },
-        getContext({ state: state, queryResults: { 'peers': [ {'rpc_address': types.InetAddress.fromString('::2') } ] }}));
+      const cc = newInstance(
+        { contactPoints: [ '::1' ], policies: { loadBalancing: lbp, reconnection: rp } },
+        getContext({ state: state, queryResults: { 'peers': [ {'rpc_address': types.InetAddress.fromString('::2') } ] }, failBorrow: [-1,1]}));
```
absurdfarce, line 427: A necessary change to fix one of the unit tests.  The old test relied on triggering an exception in borrowAConnection() which in turn would lead into the refresh reschedule logic.  The change that we make in this PR introduces another layer (the reresolve-then-retry logic) which now must _also_ fail in order to reach the refresh reschedule logic.  The test code overrides createConnection() with a custom implementation which uses the failBorrow array to determine whether or not it should "connect" successfully (in the context of the test) or throw an error.  Our addition causes an error to be thrown on the _second_ call to createConnection(), allowing the first to use the same logic it always has.

Interestingly the first call fails as well due to the custom data passed in via the LBP.

absurdfarce:
Appveyor test failures _look to be_ spurious but it's hard to tell for sure.

Both branch and PR builds are failing on https://github.com/datastax/nodejs-driver/blob/nodejs-632/test/integration/short/control-connection-tests.js#L108-L129.  The error text (as reported by the build) looks to be a ccm error:

```
Error executing ccm:Error executin
Please specify the DC this node should be added to
```

It's unclear what version of ccm is in play here; based on the appveyor log it appears to be a fixed version stored as a dep on the system running the test and the version is never printed.  The CCM_VERSION env var passed in specifies the version of C* to be used by ccm rather than the version of ccm itself.

Also worth noting that this test passes without issue in my local environment:

```
(ccm-private) [mersault@linux nodejs-driver]$ JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk CCM_VERSION=3.11.4 npm run foo                                                           

> cassandra-driver@4.6.3 foo /work/git/nodejs-driver
> mocha test/integration/short --recursive -R spec --exit -g 'should subscribe to TOPOLOGY_CHANGE'                                                                               



  ControlConnection
    #init()
      ✓ should subscribe to TOPOLOGY_CHANGE add events and refresh ring info (53413ms)
      ✓ should subscribe to TOPOLOGY_CHANGE remove events and refresh ring info (62337ms)


  2 passing (3m)
```

I _believe_ this could be fixed by changing the invocation of bootstrapNode() in the relevant test to include a "dc" prop... maybe it's worth a shot since it might be good to avoid this in the future.

absurdfarce:
Got verbal +1 from @aholmberg earlier today, calling this good

PR #379, opened 2021-05-17
Downgrade conn for C* connections with versions beyond what we support

absurdfarce:


absurdfarce:
Got verbal +1 from @aholmberg after walk through today, calling this good

PR #380, opened 2021-06-02
Refresh Astra host IDs on connection refresh

absurdfarce:


lib/control-connection.js
```diff
@@ -417,23 +418,21 @@ class ControlConnection extends events.EventEmitter {
 
   async _refreshControlConnection(hostIterator) {
 
-    if (this.options.sni) {
-      this.connection = this._borrowAConnection(hostIterator);
```
absurdfarce, line 421: This logic (an if/else case which sometimes catches errors and sometimes doesn't) was a complete mess.  In the Astra case we fell squarely into the SNI world but connection failures (which occur naturally when all nodes are killed) was causing errors to be thrown here, which in turn brought down the entire refresh job.  This has all been radically simplified in this PR.

lib/datastax/cloud/index.js
```diff
@@ -198,6 +198,17 @@ function getMetadataServiceInfo(cloudOptions, callback) {
...
 const getMetadataServiceInfoAsync = util.promisify(getMetadataServiceInfo);
 
+async function refreshMetadataServiceInfo(options) {
+
+  if (!options.cloud) {
+    return;
+  }
+  
+  const cloudOptions = new CloudOptions(options);
+  await parseZipFile(cloudOptions);
```
absurdfarce, line 208: This has to be called here (even though we've already processed the zip file) because the current impl of parseZipFile() stores extracted values on the CloudOptions instance.  Since the original version isn't persisted anywhere we have to fully populate the new one we just created before getting metadata info.

lib/datastax/cloud/index.js
```diff
@@ -198,6 +198,17 @@ function getMetadataServiceInfo(cloudOptions, callback) {
 
 const getMetadataServiceInfoAsync = util.promisify(getMetadataServiceInfo);
 
+async function refreshMetadataServiceInfo(options) {
```
absurdfarce, line 201: This function is now poorly named.  It also is equivalent to datastax.cloud.init() if we exclude the SNI special case resolution logic.  It might make more sense to just make that function more general.

lib/control-connection.js
```diff
@@ -417,23 +418,21 @@ class ControlConnection extends events.EventEmitter {
...
+
+        this.log("info", "Refreshing host IDs for cloud nodes from metadata service");
+        await cloud.init(this.options);
+        this.log("info", `Current host IDs: ${this.options.contactPoints}`);
       }
+
+      this.log("info", "Re-initializing the control connection");
+      this._contactPoints.clear();
+      this._resolvedContactPoints.clear();
+      await this.init(true);
```
absurdfarce, line 435: We still get different behaviour in the SNI case (which was the original goal of the if/else logic above)... it's just that now that logic lives in this.init() rather than inline.

absurdfarce:
Closing this out since the Astra side changes referenced in ZD72512 appear to have resolved the issue.

PR #387, opened 2022-05-24
Log negotiated SSL/TLS versions when using encrypted connections

absurdfarce:


absurdfarce:
This can be tested a couple different ways.  You can set a specific TLS version for Cassandra via cassandra.yaml:

```
client_encryption_options:
  enabled: true
...
  protocol: TLSv1.2
```

You can also configure the TLS version on the client via the "secureOptions" key of the "sslOptions" map.  An example client which (optionally) disables TLS 1.3 can be found in [nodejs647.tar.gz](https://github.com/datastax/nodejs-driver/files/8766463/nodejs647.tar.gz).  You might have to tweak the TLS versions depending on what's available on your system.

PR #388, opened 2022-06-09
NODEJS-646 Unable to add address for initially unresolved host

absurdfarce:


absurdfarce:
Test failure is from a single server timeout in an integration test with a single server/nodejs version pair.  Everything else (including this test with other server and/or nodejs versions) passed without issue.

test/unit/control-connection-tests.js
```diff
@@ -282,6 +282,58 @@ describe('ControlConnection', function () {
...
+        await cc.init();
+      } catch (e) {
+        err = e;
+      }
+
+      assert.isNull(err);
+      assert.deepStrictEqual(
+        cc.getResolvedContactPoints().get(hostname),
+        resolvedAddresses.map(a => a + ":9042"));
+    });
```
absurdfarce, line 336: This test fails _without_ the fix above with an error similar to what's described in [NODEJS-646](https://datastax-oss.atlassian.net/browse/NODEJS-646) ("TypeError: Cannot add property 0, object is not extensible").  With the fix above this test passes.

PR #392, opened 2023-02-15
Workaround to fix dse builds

joao-r-reis:


absurdfarce:
No objection in general, but I'm wondering how far we want to go with this workaround.  Seems like the Right Answer :tm: would be:

* Use the same patch versions for a given DSE major.minor across all the builds
* Change the Packer build to pre-load those specific DSE versions so that they're already available

All of that said, hopefully at some point in the future we restore the dynamic nature of DSE version selection... and then we'd just have to undo all that.  But in the time in-between (i.e. now) it seems useful to take the steps above, does it not?

joao-r-reis:
The first step Is what I'm trying to accomplish atm, these versions are the same ones that I used on the C# driver.

The second step is not necessary tbh, we never had the server versions pre-loaded in the runners before so I don't see any reason to start doing it now. Besides, that would force us to pre-load all the versions even though each runner only uses one version.

absurdfarce:
Well, kinda.  We'd definitely need to keep the DSE version specified by the Jenkinsfile in sync with whatever's installed on the runner, but if we're manually specifying those versions in the short-term that isn't a huge burden.

We wouldn't have done this before because the actual DSE version used (major.minor.patch) would be the most recent one available and we don't have that functionality available in AWS yet.  Once we get it back I agree there's probably less need for it, but while we're manually specifying everything why not?

Hmmmm, I was thinking we did this more broadly 'cause I remembered seeing it somewhere (and removing it)... but as I look back on the [big AWS PR](https://github.com/riptano/openstack-jenkins-drivers/pull/94) it looks like the only place we do this was for the really old C* versions that we don't worry about anymore.  So never mind, maybe it isn't that big a deal.  It's something we _could_ do I suppose but maybe it's not worth it.

absurdfarce (approved):

PR #403, opened 2023-05-25
Typescript generics for main Client methods

twoojoo:
When working with typescript I find very comfortable to declare the type of the requested data with generics instead of using the "as" keyword. 
So I just added generics to the main methods (**execute**, **eachRow** and **batch**) and to the types on which they are based (**Row** and **ResultSet**). I skipped graph methods and their types because I'm not really aware of how how they works in Cassandra.

Since every generic has "any" as default type, backwards compatibility should be ok (except for Row switching from being an interface to being a type, which I'm not sure if it could lead to problems in that sense).

Offtopic: it would be nice to set a string union type for hints. I'm only doubting whether hints can be used for user defined types too, which would probably make this useless.


SiyaoIsTraveling:
This would be great improvement. I created a ticket here: https://datastax-oss.atlassian.net/browse/NODEJS-562

absurdfarce:
These fixes will be handled as part of the conversion to TypeScript.

PR #405, opened 2023-08-24
Correct Client.stream return type: from EventEmitter to ResultStream

p-flock:
The client implementation of the `stream` method returns a `ResultStream`. This commit corrects the type listed in the index.ts file to line up with the implementation

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -94,7 +94,7 @@ export class Client extends events.EventEmitter {
   eachRow(query: string,
           rowCallback: (n: number, row: types.Row) => void): void;
 
-  stream(query: string, params?: ArrayOrObject, options?: QueryOptions, callback?: EmptyCallback): events.EventEmitter;
+  stream(query: string, params?: ArrayOrObject, options?: QueryOptions, callback?: EmptyCallback): types.ResultStream;
 
   batch(
     queries: Array<string|{query: string, params?: ArrayOrObject}>,


p-flock:
Just pre-emptively saying I have signed the datastax CLA

absurdfarce:
Thanks for the PR @p-flock!  And for the pre-emptive note about the CLA. ;)

With regard to the change itself, the intent of the stream() interface is to provide an easy entry point for handlers of streaming events along the lines of what's [laid out](https://github.com/datastax/nodejs-driver#row-streaming-and-pipes) in the documentation.  I could perhaps see a case for making the returned type stream.Readable in order to expose some additional context but it's not immediately clear to me that you buy much by going a step further and marking that function as returning types.ResultStream.  Is there some specific bit of functionality on types.ResultStream that you're looking to expose?  I'm not opposed to the idea necessarily.. I'm just trying to get a better handle on your use case in order to understand the proposed change.

Thanks again!

p-flock:
> Thanks for the PR @p-flock! And for the pre-emptive note about the CLA. ;)
> 
> With regard to the change itself, the intent of the stream() interface is to provide an easy entry point for handlers of streaming events along the lines of what's [laid out](https://github.com/datastax/nodejs-driver#row-streaming-and-pipes) in the documentation. I could perhaps see a case for making the returned type stream.Readable in order to expose some additional context but it's not immediately clear to me that you buy much by going a step further and marking that function as returning types.ResultStream. Is there some specific bit of functionality on types.ResultStream that you're looking to expose? I'm not opposed to the idea necessarily.. I'm just trying to get a better handle on your use case in order to understand the proposed change.
> 
> Thanks again!

sorry for the delayed response. 

The change I'm hoping to make is not actually to the implementation, just to correct the type that is specified in the `index.ts` file. 

[the client implementation of stream](https://github.com/datastax/nodejs-driver/blob/3c852c6c974da698570cecf032c34bfbd676bc6e/lib/client.js#L735C1-L742)
 has `@returns {ResultStream}` in it's docblock, and that matches the implementation below.

what I believe is incorrect is the `index.d.ts` [file](https://github.com/datastax/nodejs-driver/blob/3c852c6c974da698570cecf032c34bfbd676bc6e/index.d.ts#L97) which lists the return type of this method as `events.EventEmitter` which causes unexpected type errors for typescript libraries or applications which rely on that method. 

For example in my application I cast the result of `Client.stream`calls with `as unknown as Readable`, because that matches the runtime behavior of the the datastax driver.

It's possible I'm mistaken and there is good reason for the type specified in the definition file to be `events.EventEmitter`, but when I first came across this it seemed like a simple issue of the type definition file not being maintained along with library updates, since this library is not actually compiled with typescript it is very understandable for the mismatch to go unnoticed.

SiyaoIsHiding:
Thank you for your PR! This fix will be handled as part of the conversion to TypeScript.

PR #406, opened 2023-08-29
Initial PR to introduce new platform, i.e. Jammy as base OS and newer NodeJS versions

weideng1:
also updated C* and DSE versions to the currently supported ones

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -309,7 +309,7 @@ pipeline {
   }
 
   environment {
-    OS_VERSION = 'ubuntu/bionic64/nodejs-driver'
+    OS_VERSION = 'ubuntu/jammy64/nodejs-driver'
     JUNIT_REPORT_STACK = "${params.JUNIT_REPORT_STACK ? '1' : '0'}"
     JUNIT_REPORT_PATH = '.'
     TEST_TRACE = "${params.TEST_TRACE ? 'on' : 'off'}"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,7 +2,7 @@
 
 
 def initializeEnvironment() {
-  def nodeVersions = ['8': '8.16.2', '10': '10.17.0', '12': '12.13.0']
+  def nodeVersions = ['8': '8.16.2', '10': '10.17.0', '12': '12.13.0', '16': '16.20.2', '18': '18.17.1', '20': '20.5.1']
   env.DRIVER_DISPLAY_NAME = 'Cassandra Node.js Driver'
   env.DRIVER_METRIC_TYPE = 'oss'
   if (env.GIT_URL.contains('riptano/nodejs-driver')) {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -2,7 +2,7 @@
 
 
 def initializeEnvironment() {
-  def nodeVersions = ['8': '8.16.2', '10': '10.17.0', '12': '12.13.0']
+  def nodeVersions = ['8': '8.16.2', '10': '10.17.0', '12': '12.13.0', '16': '16.20.2', '18': '18.17.1', '20': '20.5.1']
```
absurdfarce, line 5: Along the same lines as the conversation in the runner image PR we should do away with anything < nodejs 16.x here.  So 8, 10 and 12 would go away outright.

Jenkinsfile
```diff
@@ -333,17 +319,14 @@ pipeline {
...
                    '4.0',     // Development Apache Cassandra
                    'dse-5.1.35', // Legacy DataStax Enterprise
-                   'dse-6.0.18', // Previous DataStax Enterprise
-                   'dse-6.7.17', // Current DataStax Enterprise
                    'dse-6.8.30' // Development DataStax Enterprise
           }
           axis {
             name 'NODEJS_VERSION'
-            values '8', '10', '12'
+            values '8', '10', '12', '16', '18', '20'
```
absurdfarce, line 329: Along the lines of what's discussed above 8, 10 and 12 would be removed here as well.

Jenkinsfile
```diff
@@ -361,21 +344,21 @@ pipeline {
...
           exclude {
             axis {
               name 'NODEJS_VERSION'
-              values '12'
+              values '20'
             }
             axis {
               name 'CASSANDRA_VERSION'
-              values '2.1', '3.11', 'dse-6.0.18', 'dse-6.8.30'
+              values '2.1', '3.11', 'dse-6.8.30'
```
absurdfarce, line 361: The changes in this section all appear to be excludes, specifically nodejs + Cassandra versions which should be excluded from the per-commit matrix.  I'd argue we should remove this entirely for now; if we note some kind of platform-specific issue that requires exclusions along these lines we can re-introduce them but we should start by trying to run the whole matrix.

Jenkinsfile
```diff
@@ -462,17 +445,14 @@ pipeline {
...
                    '4.0',     // Development Apache Cassandra
                    'dse-5.1.35', // Legacy DataStax Enterprise
-                   'dse-6.0.18', // Previous DataStax Enterprise
-                   'dse-6.7.17', // Current DataStax Enterprise
                    'dse-6.8.30' // Development DataStax Enterprise
           }
           axis {
             name 'NODEJS_VERSION'
-            values '8', '10', '12'
+            values '8', '10', '12', '16', '18', '20'
```
absurdfarce, line 455: Should just be 16, 18 and 20 here as well.

Jenkinsfile
```diff
@@ -569,17 +549,14 @@ pipeline {
...
                    '4.0',      // Development Apache Cassandra
                    'dse-5.1.35', // Legacy DataStax Enterprise
-                   'dse-6.0.18', // Previous DataStax Enterprise
-                   'dse-6.7.17', // Current DataStax Enterprise
                    'dse-6.8.30' // Development DataStax Enterprise
           }
           axis {
             name 'NODEJS_VERSION'
-            values '8', '10', '12'
+            values '8', '10', '12', '16', '18', '20'
```
absurdfarce, line 559: One last spot to make sure we only have 16, 18 and 20

absurdfarce (changes requested):
About what I was thinking we were going to wind up with, and the removal of the older C* and DSE versions is _very_ welcome here.  A few things to fix up but I think this is moving in a very good direction.

absurdfarce:
One thing we're missing here (and I think we missed in the Python platform upgrade as well) is the Travis config.  At some point we should update Travis so that it's only testing on the platform versions we're testing on Jenkins.  That doesn't necessarily _have_ to happen at the same time (and I wouldn't want to slow down work on the overall platform grade to address Travis issues) but it's something we shouldn't lose track of.

Jenkinsfile
weideng1: addressed in the latest commit

Jenkinsfile
weideng1: removed the exclude section entirely

absurdfarce (approved):
This also looks about like what I was expecting!

PR #407, opened 2023-09-01
NODEJS-655 Initial cut at vector support

absurdfarce:


lib/encoder.js
```diff
@@ -890,6 +907,95 @@ function defineInstanceMembers() {
     return Buffer.concat(parts, totalLength);
   };
 
+  this.decodeVector = function(buffer, params) {
+    const subtype = params["subtype"];
+    const dimensions = params["dimensions"];
+    const elemLength = 4; // TODO: figure this out based on the subtype
```
phact, line 913: I think we want to explicitly support only float subtypes since that's what cassandra does? What do you do in the other drivers @absurdfarce ?

phact (approved):

lib/encoder.js
absurdfarce: We're only supporting floats _for now_ @phact ... that's why I have some of the error messaging says what it says at the moment.  But there's no guarantee that it'll stay that way going forward and I don't really wanna build that assumption into the code.  We should be able to determine a size pretty readily based on subtype; we can then just add new subtypes as necessary in the future as they're added to Cassandra.

Re: the other drivers... the Java driver [does something really dumb](https://github.com/datastax/java-driver/blob/4.x/core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/VectorCodec.java#L115-L126) to handle this.  Fixing that is _very_ high on my list of "things I'd like to do with the driver in the future".  The Python driver does [something similar to what we're doing here](https://github.com/datastax/python-driver/blob/master/cassandra/cqltypes.py#L1444), again as a temporary expedient knowing that the current release only supports 4 byte floats.

Eventually all of these impls should be generalized so that we don't have this hidden assumption about data size built-in.  There's a _separate_ concern about the fact that we only support fixed-size subtypes so anything that has a variable size will definitely break things.  There was some conversation around this on [CASSANDRA-18504](https://issues.apache.org/jira/browse/CASSANDRA-18504) but I wouldn't say it's resolved.

synedra:
I can't "approve" the PR because I don't seem to be in the org, but I did a few hours of testing, including using Astra and OpenAI embeddings, and the driver acted as I expected in every case.

lib/encoder.js
```diff
@@ -695,12 +707,17 @@ function defineInstanceMembers() {
     }
     return value;
   };
-  this.encodeCustom = function (value, name) {
-    const handler = customEncoders[name];
+  this.encodeCustom = function (value, customTypeName) {
+
+    // Special handling for vector custom types (since they have args)
+    if (customTypeName.indexOf(customTypeNames.vector) >= 0) {
```
hhughes, line 713: why does this one use `indexOf` instread of `startsWith`? If theres a good reason please add a comment

lib/encoder.js
```diff
@@ -890,6 +907,95 @@ function defineInstanceMembers() {
...
+   * @returns {Array}
+   * @internal
+   */
+  this.parseVectorTypeArgs = function(typeName, stringToExclude, subtypeResolveFn) {
+
+    const argsStartIndex = stringToExclude.length + 1;
+    const argsLength = typeName.length - (stringToExclude.length + 2);
+    const params = parseParams(typeName, argsStartIndex, argsLength);
+    if (params.length === 1 && params[0] === "") {
+      return {};
```
hhughes, line 991: when are vector params empty? please add a comment

lib/encoder.js
```diff
@@ -890,6 +907,95 @@ function defineInstanceMembers() {
...
+    return Buffer.concat(elems);
+  };
+
+  /**
+   * Extract the (typed) arguments from a vector type
+   *
+   * @param {String} typeName
+   * @param {String} stringToExclude Leading string indicating this is a vector type (to be excluded when eval'ing args)
+   * @param {Function} subtypeResolveFn Function used to resolve subtype type; varies depending on type naming convention
+   * @returns {Array}
```
hhughes, line 982: Is this return type doc-string correct? It looks like we're returning a dictionary (unsure of the exact syntax), similar question for second param of `encodeVector`

hhughes (approved):
A few minor clarification comments, LGTM

lib/encoder.js
absurdfarce: Nope, you're absolutely right @hhughes .  Fixing.

lib/encoder.js
absurdfarce: Nope, that's out-of-date after some late code changes.  Good catch; I'll update this as well.

lib/encoder.js
absurdfarce: This actually turns out to be a thornier question than I expected.

The empty vector params case comes about when the driver has to guess the type for an object supplied for param interpolation.  If it's a Float32Array it (correctly) guesses that it's a vector but doesn't fill in the args (subtype and dimensions)... which leads to the problem.

We prolly do have enough info to fill in what's needed there.  But I don't want to have to generate a string representing vector + args only to have it re-parsed later on; it's much better to just parse it at that point and return a representation of a vector type containing objects representing subtype and dimension.  But _that_ means we'd have to change all the subsequent logic to work with that case _and_ add earlier parsing of subtypes and dimensions in cases where we aren't guessing the type based on and input param.

That entails a bit bigger refactoring of when the args are eval'd and how they're passed around.  I'm working on this now, hoping to have it wrapped up shortly enough that I can still do the release today.

absurdfarce: This turned out to be trickier than I'd expected.  In order to make the change I really wanted to make we'd need Encoder.encode() to parse the vector args and that just doesn't make sense on multiple levels.  Encoder.encode() shouldn't have knowledge of individual param structure for specific custom types AND everything is built around Encoeder.encode() returning objects with an "info" attribute representing the literal subtype string.  So we'd have to add logic from levels well below Encoder.encode() and change it's return type... all just to support this.

Fortunately, there's a much better way.  We have all the information we need to build a fully parameterized type string for the "info" attribute we return from Encoder.guessDataType().  I've resisted doing this kind of thing 'cause I didn't want to be in the business of building type strings (that's really the servers job) but here we are.

Once this functionality is added to guessDataType() I've confirmed that the empty param case above is no longer necessary.  The change also required re-working some of the test cases but the overall effect there seems to be pretty positive as well.

PR #408, opened 2023-09-07
NODEJS-657 updated travis to use nodejs 16, 18 and 20, and updated typscript ver…

weideng1:
…sion to 4.9 so that it works with nodejs 18

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,33 +9,33 @@ jobs:
     - stage: "tests"
       name: "Run eslint"
       script: "npm install -g eslint@4; npm run eslint;"
-      node_js: "10"
-    - name: "TypeScript 2.9 generation and compilation tests"
-      node_js: "10"
+      node_js: "18"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -9,33 +10,33 @@ jobs:
...
       script: "npm install -g eslint@4; npm run eslint;"
-      node_js: "10"
-    - name: "TypeScript 2.9 generation and compilation tests"
-      node_js: "10"
+      node_js: "18"
+    - name: "TypeScript 4.9 generation and compilation tests"
+      node_js: "18"
       script:
-        - npm install -g typescript@2.9;
+        - npm install -g typescript@4.9;
```
absurdfarce, line 17: I'm not _completely_ convinced yet this is where we want to land but it's unquestionably an improvement over the current state of affairs.  We can always tweak it later if we decide we want to do something different.

absurdfarce (approved):
Absolutely necessary to get us up to supporting new nodejs versions

PR #410, opened 2023-09-08
NODEJS-660 Convert platform-dependent bits of unit test to mocks

absurdfarce:


absurdfarce:
Looks like the ControlConnection test now passes.  We're still seeing some failures in unit tests due to issues addressed in NODEJS-658; we won't have a clean TravisCI build until both changes are brought together.

test/unit/control-connection-tests.js
```diff
@@ -81,44 +63,44 @@ describe('ControlConnection', function () {
...
+          throw new Error('dns.lookup() should not be used');
+        }
+      }});
 
-      cc.shutdown();
-      helper.assertInstanceOf(err, errors.NoHostAvailableError);
-      assert.deepEqual(state.connectionAttempts.sort(), [ '127.0.0.1:9042', '::1:9042' ]);
+      return testResolution(ControlConnectionMock,
+        [ '127.0.0.1:9042', '::1:9042' ],
+        [ '127.0.0.1:9042', '[::1]:9042' ]);
```
weideng1, line 84: small nit, any reason why we don't want to use the fourth parameter `'localhost:9042'` here?

absurdfarce: Yeah, this was done deliberately here.  We want to confirm that we can handle the case in which only a hostname is provided, but testResolution will default to "localhost" if no name is supplied.  So really we want to make sure the default port is added correctly in this case.

Note that the actual hostname here isn't particularly interseting; we're mocking the DNS calls so under the current implementation any hostname we pass in will return the expected values.  I'll make those mocks just a bit selective about what they return; if they only return mocked values when they actually see a request for the name "localhost" we'll get an even better test.

I'll also tweak the verbiage on the "it" block to try to make this all a bit clearer.

PR #412, opened 2023-09-14
NODEJS-659 updated dependencies to address CVEs

weideng1:


absurdfarce:
Jenkins is happy, Travis is happy and local test runs on my minimum node instance (16.x) look good.

PR #415, opened 2023-09-20
NODEJS-665: Remove Object.hasOwn() and update nodejs version specified in package.json

absurdfarce:


package.json
```diff
@@ -53,6 +53,6 @@
     "eslint": "eslint lib test"
   },
   "engines": {
-    "node": ">=8"
+    "node": ">=16"
```
absurdfarce, line 56: I'm somewhat ambivalent on this change.  The driver _might_ continue to work well with older versions (even if we don't support it) but this explicitly shuts it off.  That said, we are making a concerted effort to keep supported versions updated... so in the end I feel this is the right move.

Older versions of nodejs always have access to older versions of the driver.

We don't _officially_ support nodejs 16.x as of 4.7.0 but it only went EOL a little over a week ago... so I'm inclined to leave this at "16" in order to facilitate the transition for our users.

absurdfarce:
Easily reproduced the indicated error by attempting to run encoder tests with node.js 16.8.0.  The changes in this PR immediately solve the problem.

PR #416, opened 2023-10-18
Bun Support

Darker-Ink:
So I'm using bun for one of my projects. I'm also using cassandra-driver. These are a few issues which cause the driver to not work with bun. I've gone ahead and fixed them.

I tested this on bun version 1.0.6, With that I also tested with node v16.20.2+

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -173,7 +173,10 @@ class Connection extends events.EventEmitter {
 
     if (!this.options.sslOptions) {
       this.netClient = new net.Socket({ highWaterMark: this.options.socketOptions.coalescingThreshold });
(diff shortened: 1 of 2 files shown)


Darker-Ink:
I've also already signed the datastax CLA

asour8:
image

works, here is some example minimal code:
```ts
import cassandra, { types } from 'cassandra-driver';

const db = new cassandra.Client({
    contactPoints: [
        "172.17.0.2"
    ],
    localDataCenter: "datacenter1",
    protocolOptions: {
        maxVersion: types.protocolVersion.v4
    }
});

try {
    await db.connect();
    console.log(colors.green('Connected to ScyllaDB :3'));
} catch (error) {
    console.error(error);
}
```

PR #417, opened 2023-11-30
Remove `@types/long` dependency

ramikg:
I've removed the `@types/long` dependency, in order to avoid the following deprecation warning:
```
This is a stub types definition. long provides its own type definitions, so you do not need this installed.
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,6 @@
       "version": "4.7.2",
       "license": "Apache-2.0",
       "dependencies": {
-        "@types/long": "~5.0.0",
         "@types/node": "^18.11.18",
         "adm-zip": "~0.5.10",
         "long": "~5.2.3"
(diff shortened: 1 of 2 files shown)


nicoloboschi (approved):

Graphmaxer:
Interested into this change since it break our builds

```
error TS2688: Cannot find type definition file for 'long'.
   The file is in the program because:
   Entry point for implicit type library 'long'
```

Graphmaxer:
I created an issue in the Jira https://datastax-oss.atlassian.net/browse/NODEJS-671 Maybe it will help :/

SiyaoIsTraveling (approved):
LGTM! Thanks!
The package `long` used to be pure javascript without type declaration files (`.d.ts` files). Therefore, in the past we needed `@types/long`, a community contributed package of type declaration files for `long`.
Now `long` is shipped with type declaration files, see the `index.d.ts` [here](https://www.npmjs.com/package/long?activeTab=code). Therefore, we don't need `@types/long` anymore.
Please merge as long as our CI got green @absurdfarce .

absurdfarce:
Thanks for the investigation @SiyaoIsTraveling; your explanation makes perfect sense to me!

I've kicked off a Jenkins build for this PR; assuming there's no regression there (and I really don't expect one) we should be good to get this guy merged.

absurdfarce:
Bah, missed this in my last comment... big thanks to @ramikg as well for raising (and fixing!) the issue in the first place!

absurdfarce:
Jenkins run was indeed green, calling this good

ramikg:
Thank you!

PR #419, opened 2024-03-11
NODEJS-668: Handle empty pages in async pagination

asztal:
This fixes an issue where an empty page in the results would cause pagination to end, even if there were more results to be returned.

In practice this has happened when using `allow filtering` or when encountering deleted rows in Scylla materialised views.


absurdfarce:
Greetings @asztal and thanks for the contribution! Apologies for taking so long to get back to you.

Have you signed the Contributor License Agreement for contributions to DataStax open source projects? If not you can find it at https://cla.datastax.com/. Thanks!

PR #421, opened 2024-04-30
Support `DELETE` without providing full primary key

ramikg:
# Overview

Given a table with primary key ((_partition1_), _column1_, _column2_), Cassandra and the Node.js have a different behavior.
This PR changes the driver to conform to Cassandra's behavior.

## Cassandra behavior

Allows `DELETE` statements which specify:
- All of the partition key columns.
- Some or all of the clustering columns, under the following condition:
  - If a clustering column is specified, the all of the higher-level/preceding columns must also be specified.

## DataStax Node.js driver behavior

Requires all of the primary key columns (i.e. partition key and clustering columns) to be specified.
Not providing all primary key columns will result in the following error:
```
Error: No table matches (all PKs have to be specified)
```

# Note on code review
This PR also changes the `ModelMapper` tests to have more indicative column names.
To make the PR easier to review, this minor refactor was done in a separate commit.

### DIFF
--- a/lib/mapping/object-selector.js
+++ b/lib/mapping/object-selector.js
@@ -255,14 +255,25 @@ class ObjectSelector {
               throw new Error(`Table "${info.tables[i].name}" could not be retrieved`);
             }
 
-            // All partition and clustering keys from the table should be included in the document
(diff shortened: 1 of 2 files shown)


ShahafCohenTarica:
Are there any news about this PR?

PR #427, opened 2024-10-16
NODEJS-666: Extend driver vector support to arbitrary subtypes and fix handling of variable length types (OSS C* 5.0)

SiyaoIsHiding:
Will add more tests


lib/encoder.js
```diff
@@ -188,23 +208,38 @@ function defineInstanceMembers() {
...
-  this.decodeCustom = function (bytes, typeName) {
+
+  /**
+   * 
+   * @param {Buffer} bytes 
+   * @param {ColumnInfo | string} typeName
+   * @param {ColumnInfo | null} columnInfo 
+   * @returns 
+   */
+  this.decodeCustom = function (bytes, typeName, columnInfo = null) {
```
absurdfarce, line 219: I really dislike the idea of adding an extra param here.

The [original impl](https://github.com/datastax/nodejs-driver/blob/master/lib/encoder.js#L1621) of this function passes in "type.info" from the type info it gets as an arg.  Your [new impl](https://github.com/SiyaoIsHiding/nodejs-driver/blob/vector/lib/encoder.js#L1771) passes in "type.info" as well but also passes in... "type" as well.  One of these args seems redundant; why pass "type.info" when you can just as easily get the "info" property of the type information you're also passing.

Seems like you can do the _exact_ same thing by just passing "columnInfo" here and adding "typeName = columnInfo.info" near the top of this function.

lib/encoder.js
```diff
@@ -188,23 +208,38 @@ function defineInstanceMembers() {
...
-    if (typeName.startsWith(customTypeNames.vector)) {
-      return this.decodeVector(bytes, this.parseVectorTypeArgs(typeName, customTypeNames.vector, this.parseFqTypeName));
+    if (columnInfo != null && 'customTypeName' in columnInfo && columnInfo.customTypeName === 'vector') {
+      return this.decodeVector(bytes, columnInfo);
     }
+
+    if(typeof typeName === 'string' && typeName.startsWith(customTypeNames.vector)) {
+      const vectorColumnInfo = /** @type {VectorColumnInfo} */ (this.parseFqTypeName(typeName));
+      return this.decodeVector(bytes, vectorColumnInfo);
+    }
```
absurdfarce, line 234: Is there a reason we need _both_ cases above?  Seems like if one of these works out as true then the other does too... right?

Maybe I'm thinking with the "use two params rather than three" mindset from the comment above...

lib/encoder.js
```diff
@@ -908,70 +955,130 @@ function defineInstanceMembers() {
...
+        if (offset + size > buffer.length) {
+          throw new TypeError('Not enough bytes to decode the vector');
+        }
+        rv[i] = this.decode(buffer.subarray(offset, offset + size), subtype);
+        offset += size;
+      }else{
+        if (offset + elemLength > buffer.length) {
+          throw new TypeError('Not enough bytes to decode the vector');
+        }
+        rv[i] = this.decode(buffer.subarray(offset, offset + elemLength), subtype);
```
absurdfarce, line 985: We already have the type code of the subtype here, don't we?  So we can actually look up the decode function for the subtype using that code rather than having to go all the way back through decode() can't we?

lib/encoder.js
```diff
@@ -908,70 +955,130 @@ function defineInstanceMembers() {
...
+            return -1;
+          }
+          return subtypeSerialSize * cqlType.dimension;
+          
         }
-        return subtype;
-      }
+        return -1;
+      default:
+        throw new TypeError('Cannot calculate size');
```
absurdfarce, line 1049: Hmmmm... I'm wondering if it's better to make the default case return -1.  You wouldn't have to specify all the data types that _should_ have this behaviour (like you do in this impl)... you'd just specify size functions for the types that _are_ of a fixed size and everybody else returns -1.

Best argument I can think of _not_ to do it that way is that there may be some benefit in being explicit for every single type: is it fixed size or not?  It does make the code a bit more cumbersome, though.

lib/encoder.js
```diff
@@ -1657,32 +1807,32 @@ Encoder.prototype.encode = function (value, typeInfo) {
...
   }
   else {
     //Lets guess
-    type = Encoder.guessDataType(value);
-    if (!type) {
+    const guessedType = Encoder.guessDataType(value);
+    if (!guessedType) {
       throw new TypeError('Target data type could not be guessed, you should use prepared statements for accurate type mapping. Value: ' + util.inspect(value));
+    }else{
+      type = guessedType;
```
absurdfarce, line 1835: Why... do we have an else block on an if test which throws if it's successful?  Seems like the original impl says exactly the same thing but in a more succinct way.

absurdfarce (changes requested):
This is really good work @SiyaoIsHiding!  I love the increased test coverage we're getting here; it's good to see the same robust coverage we have for the other drivers that support vectors coming to node.js as well.  There are a few things here I'd really like to see changed, although fortunately most (all?) of them are internal things that shouldn't be super-visible to users.

lib/types/index.js
```diff
@@ -161,8 +161,17 @@ const dataTypes = {
           return this.getByName(x.trim());
         }, this)};
       }
+      const vectorMatches = /^(vector)<*(.+) *, *(\d+)>$/.exec(name);
+      if(vectorMatches){
+        return {
+          code: this.custom,
+          customTypeName: 'vector',
+          info: this.getByName(vectorMatches[2]),
+          dimension: parseInt(vectorMatches[3], 10)
```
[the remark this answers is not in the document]
SiyaoIsHiding: I like this idea and a lot of problems go away with such a design!

test/unit/encoder-vector-tests.js
```diff
@@ -0,0 +1,393 @@
...
+    //     code: types.dataTypes.custom,
+    //     info: {
+    //       code: types.dataTypes.custom,
+    //       info: 'duration'
+    //     },
+    //     customTypeName: 'vector',
+    //     dimension: 3
+    //   },
+    //   value: [new types.Duration(1, 2, 3), new types.Duration(4, 5, 6), new types.Duration(7, 8, 9)]
+    // },
```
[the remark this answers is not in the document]
SiyaoIsHiding: I couldn't make it work when I submitted this PR but I just made it work with the following test case
```js
  {
    subtypeString: 'duration',
    typeInfo: {
      code: types.dataTypes.custom,
      info: [{
        code: types.dataTypes.custom,
        info: 'org.apache.cassandra.db.marshal.DurationType'
      },3],
      customTypeName: 'vector',
    },
    value: [new types.Duration(1, 2, 3), new types.Duration(4, 5, 6), new types.Duration(7, 8, 9)]
  },
```
Note that I have to put `'org.apache.cassandra.db.marshal.DurationType'` instead of `duration`. Is it expected?

lib/types/index.js
```diff
@@ -161,8 +161,17 @@ const dataTypes = {
           return this.getByName(x.trim());
         }, this)};
       }
+      const vectorMatches = /^(vector)<*(.+) *, *(\d+)>$/.exec(name);
```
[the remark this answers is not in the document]
SiyaoIsHiding: TypeScript will complain about option 2 because a `VectorColumnInfo` type is supposed only has `vector` in `customTypeName`, not anything else. I would go with option 1 then :)

lib/encoder.js
SiyaoIsHiding: Reverted

lib/encoder.js
SiyaoIsHiding: If we don't call `decode()`. Then we need to
```js
  const decoder = this.decoders[type.code];
  return decoder.call(this, buffer, 'info' in type? type.info : null);
```
Which is basically calling `decode()`.

examples/package-lock.json
```diff
@@ -0,0 +1,50 @@
...
+      "version": "1.5.2",
+      "resolved": "https://registry.npmjs.org/async/-/async-1.5.2.tgz",
+      "integrity": "sha512-nSVgobk4rv61R9PUSDtYt7mPVB2olxNR5RWJcAsH676/ef11bUZwvu7+RGYrYauVdDPcO519v68wRhXQtxsV9w=="
+    },
+    "node_modules/cassandra-driver": {
+      "resolved": "..",
+      "link": true
+    }
+  }
+}
```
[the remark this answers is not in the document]
SiyaoIsHiding: Reverted

lib/encoder.js
SiyaoIsHiding: Are you asking why we need two `if` blocks for vector, one for `typeName` as `VectorColumnInfo`, another one for `typeName` as a string?
I added the second `if` block for when `typeName` is a string because the previous JSDoc says this `typeName` argument can be a string.
Although deleting it will still pass all the tests, I still tend to leave it there, because in JSDoc function signature we claim that this `typeName` supports a string argument.

lib/encoder.js
```diff
@@ -708,11 +742,25 @@ function defineInstanceMembers() {
...
+   * @param {string | ColumnInfo | [ColumnInfo, number]} customTypeName when it's a vector, it will be [subtype, dimension]
+   * @returns 
+   */
   this.encodeCustom = function (value, customTypeName) {
 
     // Special handling for vector custom types (since they have args)
-    if (customTypeName.startsWith(customTypeNames.vector)) {
-      return this.encodeVector(value, this.parseVectorTypeArgs(customTypeName, customTypeNames.vector, this.parseFqTypeName));
+
+    //TODO: here we cannot tell whether it's a vector or not, because we don't have the customTypeName
```
SiyaoIsHiding, line 756: My only concern is here: Assume the type of a vector is 
```json
{
      code: types.dataTypes.custom,
      customTypeName: 'vector',
      info: [{ code: types.dataTypes.int }, 3]
    }
```
the `customTypeName` argument we get here is
```json
[{ code: types.dataTypes.int }, 3]
```
We basically lost the `customTypeName: 'vector'`, and does not have a good way to tell it's a vector. Unless we say as long as we get an array and the second element is a number, then we treat it as a vector.

lib/encoder.js
SiyaoIsHiding: Will go away after we change our API from (value, type.info) to (value, type)

lib/encoder.js
absurdfarce: To be clear: this is a pretty insignificant one to me.  It's a relatively minor savings not having to go back through decode() but the difference is pretty small.

I'm perfectly fine with continuing to use decode() here.

test/unit/encoder-vector-tests.js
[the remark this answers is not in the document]
SiyaoIsHiding: Per our conversation, this does not raise any concern.

lib/encoder.js
SiyaoIsHiding: Refactored!

lib/encoder.js
SiyaoIsHiding: Ok back to this conversation, I found out that if we delete the second `if` block, integration tests fail. Because the server may pass a `type.info` of a string like `'org.apache.cassandra.db.marshal.VectorType(org.apache.cassandra.db.marshal.VectorType(org.apache.cassandra.db.marshal.TimeUUIDType , 3) , 3)'` to `Encoder.prototype.decode`. Therefore, we still need the second if block for the case where `type.info` is a string instead of a parsed info object. 
Or we can also try to parse the string into an info object earlier in the process before it passes to `Encoder.prototype.decode`. What do you think @absurdfarce ?

lib/types/index.js
```diff
@@ -161,8 +161,16 @@ const dataTypes = {
...
         }, this)};
       }
+      const vectorMatches = /^vector<\s*(.+)\s*,\s*(\d+)\s*>$/.exec(name);
+      if(vectorMatches){
+        return {
+          code: this.custom,
+          customTypeName: 'vector',
+          info: [this.getByName(vectorMatches[1]), parseInt(vectorMatches[2], 10)]
+        };
+      }
```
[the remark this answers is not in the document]
absurdfarce: Or maybe we should make this even more compact?

```
return {
   code: this.custom,
   info: ['vector', [this.getByName(vectorMatches[1]), parseInt(vectorMatches[2], 10)]]
}
```

This is (very arguably) more consistent with existing usages since it follows the idea of "code" containing the protocol-level code while "info" contains whatever additional metadata you need.

absurdfarce (commented):
This is looking really good after the last round of changes!  I _love_ how the API seems simpler and more consistent now!

One lingering question for me about how best to represent custom types generally in a way that makes sense for vectors and other non-vector custom types but that's really it.

absurdfarce:
Pausing briefly; @SiyaoIsHiding pointed out that we ought to update the docs to reflect the behavioural changes contained in this PR.  Will be merged once those edits are in place.

PR #430, opened 2024-11-19
NODEJS-681: ControlConnection Concurrent Read and Write on .host and .connection

SiyaoIsHiding:


absurdfarce:
Posted a [question](https://datastax-oss.atlassian.net/browse/NODEJS-681?focusedCommentId=55806) on the JIRA issue about this one.  I agree that this fix might be valid on it's own terms but I'm not sure I understand how this fix will address the test failures described there.

SiyaoIsHiding:
Now we see why `_refresh` is called concurrently, I think we should continue the review of this PR.

PR #431, opened 2024-12-19
NODEJS-680: Update and Automate TypeScript Support - 1. config and auth

SiyaoIsTraveling:
I created `tsconfig.json` so that If you run `tsc` command, it will generate type declaration files `.d.ts` and source map files `.d.ts.map` in the folder `dist/lib/`. Therefore, our `package.json` needs to point `main` to `lib/index.js` and `types` to `dist/lib/index.d.ts`.

I also transformed the files in `auth` folder to clear most of the typescript errors. It mostly involves refactoring es5 classes to es6 classes.

I tested this config with [GitHub private npm registry](https://adevait.com/software/publish-private-npm-packages-with-github-package-registry). I published the package privately and tested with a typescript client project. I confirm the type hints and IDE support is working fine for me.

To consider in the future:
1. License headers?
2. Bundling?


SiyaoIsHiding:
Closing because we want to migrate to TypeScript entirely: https://github.com/datastax/nodejs-driver/pull/435

PR #438, opened 2025-04-26
NODEJS-691 Fix generated timestamp on retry

luccailliau:
Hello @absurdfarce,

There is an issue with timestamps generated by timestamp generators in `DefaultExecutionOptions`.
No matter if the request is prepared or not, any retry, including speculative execution, will generate a new timestamp for the query when starting a new execution for the same request object.

As interpreted by Cassandra, the latest timestamp take precedence. Then, queries executed and marked as successful can be overwritten by past queries when retrying the execution.

The user feels that it is a retry but in fact, it is a new query.

This is done by the `WriteQueue` when calling `request.write(...)` which call `writeQueryParameters(...)` and then `getOrGenerateTimestamp()` which always generate a new timestamp for each new execution. It should generate one and use it for each consecutive calls to make a retry operation in Cassandra.

I did not write any tests for this change since I don't have a proper setup and configuration on my side which cause tests to fail. I think testing on the client side only, like this is enough:
```
const request = new ExecuteRequest(...);

const firstBuffers = request.write(...);
const secondBuffers = request.write(...);

assert.deepStrictEqual(firstBuffers, secondBuffers); // At least on the parts for the timestamp
```

Thank you very much,
Have a good day,

Luc

### DIFF
--- a/lib/requests.js
+++ b/lib/requests.js
@@ -121,6 +121,7 @@ class ExecuteRequest extends Request {
     this.params = params;
     this.meta = meta;
     this.options = execOptions || ExecutionOptions.empty();
+    this.timestamp = this.options.getOrGenerateTimestamp();
     this.consistency = this.options.getConsistency() || types.consistencies.one;
     // Only QUERY request parameters are encoded as named parameters
     // EXECUTE request parameters are always encoded as positional parameters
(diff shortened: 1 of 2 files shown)


absurdfarce:
Note: this PR is in reference to [NODEJS-691](https://datastax-oss.atlassian.net/browse/NODEJS-691)

absurdfarce:
I agree that this is an issue (and a pretty significant one) but I'm inclined to say we should address it at a different spot.  More details can be found on NODJES-691; I'd like to keep the conversation going there as much as possible since this issue isn't actually constrained to the code being modified here.

SiyaoIsHiding (approved):
Looks good to me

absurdfarce:
Agree with @SiyaoIsHiding; this looks right to me.  Test run with these changes was as expected so I think we're good to go here!

PR #439, opened 2025-04-29
NODEJS-680: Update and Automate TypeScript Support

SiyaoIsHiding:
Massive refactoring. Changes can be broken down into the following parts:
1. Rename files from `.js` to `.ts`.
2. Update import/export syntax to ESM syntax - some index files have to export both named and default to be compatible with the previous `require` behavior.
4. Refactor ES5 style classes to ES6 style classes
5. Cleaning TypeScript errors and eslint errors in `core`
6. Fix API exposure - My rationale is: if an API was exposed in previous `.d.ts`, it's supposed to be exposed. If it has documentation on how a user should use it, it's supposed to be exposed. There are some inconsistencies. Please search `//TODO` to find them.
7. Make sure tests pass and set up CI: Replace the use of proxyquire to sinon. Add rollup for bundle, api-extractor to encapusulate, and eslint for linting.

We need to discuss:
1. Please search `//TODO` to find the places where I am not sure what to do. Some are breaking changes.
2. How can we maintain backward compatibility of the dotted type names, like `cassandra.errors.DriverError`?
In the current master branch, this is the only way to use types:
```typescript
import cassandra from 'cassandra-driver';
// ...
client.execute(query, (err: cassandra.errors.DriverError, result: cassandra.types.ResultSet) => {});
```
Or, with one level down imports
```typescript
import {errors, types } from 'cassandra-driver';
 client.execute(query, (err: errors.DriverError, result: types.ResultSet) => {});
```
However, the TypeScript default way, also the way how other libraries expose the types, is to flatten all exported classes.
```typescript
import {DriverError, ResultSet } from 'cassandra-driver';
 client.execute(query, (err: DriverError, result: ResultSet) => {});
```
Now how can we maintain backward compatibility to the previous dotted ones? The only workaround I can find is to append such things at the end of our `dist/cassandra-driver-public.d.ts`
```typescript
export namespace errors{
    export type DriverError = InstanceType<typeof DriverError>;
}
```
This is the last step of the bundle script in this PR.
3. How do we maintain backward compatibility of `import x = y.z` syntax?
```typescript
import { mapping } from "../../../index";
import Mapper = mapping.Mapper;
const mapper: Mapper = new Mapper();
```
This currently will raise an error `Mapper is a type, but used as a value`. I cannot see a solution.

5. What do we want to do with our documentation? Should we update all code snippets into TypeScript?


SiyaoIsHiding:
Contact doc team and see what we should do with JSdoc type definitions v.s. TypeScript type definitions.

SiyaoIsHiding:
Optimize the pipeline to only distribute necessary files.

PR #440, opened 2025-05-05
Cancel scheduled ControlConnection refresh if shutdown was called

betalb:
Store timeout, used to schedule refresh call, and clear it on shutdown.

Even though refresh call has check if shutdown was called, timeout itself may prevent nodejs process from shutting down until timeout is reached.

I have some concerns about test itself. Not sure that all the setup steps are required. But I've copied it mostly from previous test "should reconnect when all hosts go down and back up". What I wanted from the test:
 * spin up cluster
 * open control connection and ensure that it is live
 * stop cluster
 * call control connection shutdown
 * verify that refresh was not called

So the main test is pasted below, the rest is a boilerplate copied from other test.

```javascript
// start spying on _refresh calls
cc._refresh = sinon.spy(cc._refresh);

cc.shutdown();

// wait for reconnectionDelay with 10% lag
await helper.delayAsync(parseInt(reconnectionDelay * 1.1, 10));

assert.ok(cc._refresh.notCalled);
```

### DIFF
--- a/lib/control-connection.js
+++ b/lib/control-connection.js
@@ -99,6 +99,8 @@ class ControlConnection extends events.EventEmitter {
     this._topologyChangeTimeout = null;
     // Timeout used for delayed handling of node status changes
     this._nodeStatusChangeTimeout = null;
+    // Timeout used for scheduling reconnect
+    this._reconnectTimeout = null;
 
     if (context && context.borrowHostConnection) {
       this._borrowHostConnection = context.borrowHostConnection;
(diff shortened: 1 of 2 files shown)


betalb:
Need to refresh my memory. There is one more suspicious timeout created in `_waitForReconnection`, but I remember that I've spotted it as well. But, most likely, I've ignored it because it is small (2s).

I have a code that handles this timeout as well, but need to write a test for this particular case.

betalb:
Ah, no, there was another reason why I decided not to handle timeout in `_waitForReconnection` explicitly. During shutdown `newConnection` event is emitted (with error) which triggers cleanup of this timeout. Probably small comment will make sense.

toptobes:
I believe [timeout.unref()](https://httptoolkit.com/blog/unblocking-node-with-unref/) should achieve the same goal, but cleaner:
```js
setTimeout(() => this._refresh(), delay).unref();
```

betalb:
I'm can't be 100% sure that `_refresh` calls can't be overlapped if they are called outside and from within timeout handler

SiyaoIsHiding:
`ControlConnection.shutdown()` is only called in `client.shutDown()`, and when a user wants to shut down a client, I believe we don't want the refresh to happen at all, regardless of whether it blocks the event loop or not.
So I think we should `clearTimeout` instead of `unref()` on this one, what do you think?

PR #441, opened 2025-05-20
NODEJS-692: Add Node.js v24 to CI

SiyaoIsHiding:


absurdfarce:
Jenkins build running now; assuming everything is green there with these changes I think we're ready to go.

SiyaoIsHiding:
Updating Kerberos' version because the old one is not compatible with Node.js 24.

PR #442, opened 2025-05-21
NODEJS-693: Remove broken jsdoc from 4.8.0

SiyaoIsHiding:


absurdfarce:
I can't speak to the JS file changes but the changes to docs.yaml looks like what I would expect.

PR #448, opened 2025-12-05
NODEJS-696: Migrate Travis CI to Github Action

SiyaoIsHiding:
Migrate existing Travis CI tests, also remove Node.js 16.x and add 24.x. Cuz 16.x is EOF and 24.x is the new LTS.


.github/workflows/travis-ci.yml
```diff
@@ -0,0 +1,84 @@
...
+
+
+  unit-latest-deps:
+    name: Unit tests w/ latest dependencies - Node.js 20.x
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20.x
```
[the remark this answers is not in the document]
SiyaoIsHiding: Will the next GH Action job pick up the cached latest dependencies instead of the fixed dependencies in the package-lock.json?

.github/workflows/travis-ci.yml
```diff
@@ -0,0 +1,84 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 18.x
+          cache: 'npm'
+
+      - run: npm install
+
+      - run: npm install typescript@4.9.x --no-save
```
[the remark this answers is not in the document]
SiyaoIsHiding: I just wanted to make it clear that version 4.9 overrides

absurdfarce:
@SiyaoIsHiding we can completely close this out in favor of #454 now right?

SiyaoIsHiding:
I don't think so. Travis CI runs `eslint` and the TypeScript tests, while https://github.com/apache/cassandra-nodejs-driver/pull/454 is about unit tests and integration tests.

absurdfarce:
Okay, then we should probably find some way to merge these checks into what's being run by the CI.  There really isn't any notion of what TravisCI used to provide since this project has been donated; the actual CI is entirely public so there's no real reason to have some kind of "pre-check" for PRs when we can just run the CI against them.

.github/workflows/travis-ci.yml
```diff
@@ -0,0 +1,84 @@
...
+      matrix:
+        node: [18.x, 20.x, 22.x, 24.x]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.node }}
+          cache: 'npm'
+      - run: npm install
+      - run: npm test
```
absurdfarce, line 70: Note that this only winds up running unit tests because of the [package.json config](https://github.com/apache/cassandra-nodejs-driver/blob/v4.8.0/package.json#L44)

.github/workflows/travis-ci.yml
```diff
@@ -0,0 +1,84 @@
...
+
+      - run: npm install typescript@4.9.x --no-save
+
+      - run: |
+          pushd test/unit/typescript/
+          npx tsc -p .
+          node -e "require('./api-generation-test').generate()" > generated.ts
+          npx tsc generated.ts
+
+  typescript-latest:
```
SiyaoIsHiding, line 40: We will put this into the public CI GA, with a matrix of TypeScript Transpiler (4.9, 5.9, 6.0)

.github/workflows/travis-ci.yml
```diff
@@ -0,0 +1,84 @@
...
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.node }}
+          cache: 'npm'
+      - run: npm install
+      - run: npm test
+
+
+  unit-latest-deps:
```
SiyaoIsHiding, line 73: To remove, and to discuss with Jorge

absurdfarce: For context @jorgebay: a short summary of the discussion @SiyaoIsHiding and I had about this specific point this morning.  This feels like a pretty unusual situation; the user would have to explicitly override the versions specified in package-lock.json to get to something like this.  If they do that they're testing with different dependencies than what we use so all bets are really off at that point.

Additionally, I could see something like this as an exercise during the development process, maybe something to check for newer versions of deps.  But I don't think we want it running with all (or even some) of our CI builds do we?  I can see a case for automating it (again as a means of detecting more recent versions of deps that we may want to upgrade to) but it's far from clear to me it belongs here.

jorgebay: Correct, this is not what you usually want to test.

For the record, when installing the driver as a dependency, only the package.json file is used, the package-lock file used is only the one from the parent project (not the dependency's).

TLDR: it's not "crazy" to test it like this, it's unusual. The user will still may run into deps issues when they have conflicting dependencies. In any case, the number of prod dependencies in this project is super low.

absurdfarce:
As discussed above these changes have been folded into #454

PR #453, opened 2026-02-03
CASSNODEJS-2: CONTRIBUTING md

SiyaoIsHiding:


jorgebay (approved):

README.md
```diff
@@ -34,9 +33,19 @@ $ npm install cassandra-driver
...
 - [FAQ][faq]
 
-## Getting Help
+## Contribution and Communication
 
-You can use the [project mailing list][mailinglist] or create a ticket on the [Jira issue tracker][jira].
+Please refer to the [CONTRIBUTING.md](CONTRIBUTING.md) for details on how to contribute to the project.
+
+1. **Mailing List**:
+    - https://groups.google.com/a/lists.datastax.com/g/nodejs-driver-user
```
absurdfarce: The Google Group has been deactivated (as much as I was able to) and a message has been posted there directing everybody to ASF Slack.  I'd argue we should remove this entirely and only point people at ASF Slack or the dev ML.

README.md
```diff
@@ -250,7 +259,7 @@ The current version of the driver offers support consistent with this policy for
 
 - Apache Cassandra versions 3.0 and above.
 - DataStax Enterprise versions 5.1 and 6.8.
-- Node.js versions 18.x, 20.x, and 22.x.
+- Node.js versions 20.x, 22.x, and 24.x.
```
absurdfarce, line 262: We're coming up on a scheduled release for node.js 26.x as well... is the goal here to get this out _before_ 26.x goes live (with the expectation of adding explicitly 26.x support in the next release)?

package.json
```diff
@@ -52,6 +52,6 @@
     "eslint": "eslint lib test"
   },
   "engines": {
-    "node": ">=18"
+    "node": ">=20"
```
absurdfarce, line 55: Apologies for the noob question but what's the correlation here between engines and node.js releases?

absurdfarce (changes requested):
I would like to get rid of the remaining reference to the Google Group but otherwise this looks ready to go!  A few minor questions when looking over this but overall this looks really good!

package.json
SiyaoIsHiding: Related because we changed the supported Node.js versions in README.md. I can submit this change in another PR and ask for 2 committers to review again, but we definitely should not claim the engine requirement of `>=18`, because we just do not support node.js 18 anymore.

README.md
SiyaoIsHiding: We should only add 26.x as the supported Node.js versions in the README.md when we actually add 26.x to the CI, regardless of release plans. Otherwise, users will find 26.x as supported in the README.md and get confused.

README.md
SiyaoIsHiding: Updated, also removed this obsolete google mailing list reference from several other doc

README.md
absurdfarce: Understood @SiyaoIsHiding ... I guess that's why I was wondering if we were trying to get this out before 26.x comes out.  Adding support for a new runtime version is always unpredictable and could delay things.  I would _hope_ we'd get this version out before then... just wanted to check.

absurdfarce (approved):
All right, I'm good (in fact very good) with this after your last round of changes @SiyaoIsHiding.  Thanks for seeing this through!

absurdfarce:
This PR still needs to be squashed down to a single commit so we can rebase it.  I'll leave that to you @SiyaoIsHiding ... if you need anything from me to get this into trunk just let me know and we'll make it happen!

PR #454, opened 2026-02-17
CASSNODEJS-3: Public CI after Donation

SiyaoIsHiding:
This PR will add a GitHub Action to run unit tests and integration tests against different OSS Cassandra versions and Node.js versions. Example run: https://github.com/SiyaoIsHiding/nodejs-driver/actions/runs/22090488378

The **token range test** has to be fixed because CCM [changed](https://github.com/apache/cassandra-ccm/blob/0b19c8eef37db2e11ecf5b27c42c883be88830ce/ccmlib/cmds/cluster_cmds.py#L295-L301) the default vnode token numbers for [CASSANDRA-13701](https://issues.apache.org/jira/browse/CASSANDRA-13701) since C* 4.0+. To verify whether the driver's metadata has the correct token count, it's more reliable to actually `SELECT tokens FROM system.local` instead of calculating 2*16.

The **license header** test has to be fixed due to incorrect escaping.


jorgebay (approved):
Awesome!

.github/workflows/test.yml
```diff
@@ -0,0 +1,181 @@
...
+          mkdir -p /home/runner/workspace/tools/ccm/ssl/
+          cd /home/runner/workspace/tools/ccm/ssl/
+          keytool -genkey \
+            -keyalg RSA \
+            -alias cassandra \
+            -keystore keystore.jks \
+            -storepass cassandra \
+            -keypass cassandra \
+            -validity 364635 \
+            -dname "CN=Philip Thompson, OU=TE, O=DataStax, L=Santa Clara, ST=CA, C=TE"
```
absurdfarce: Semi-nit: we should probably get Philip's name out of here... and remove references to DataStax as well

absurdfarce (approved):
Looks like a good impl for a CI process.... nice work @SiyaoIsHiding!

Had a suggestion for a minor cleanup but it's definitely not essential; I wouldn't have a problem if this went in as-is.

.github/workflows/test.yml
SiyaoIsHiding: CN=Jane He, O=ASF

SiyaoIsHiding:
We decided to merge https://github.com/apache/cassandra-nodejs-driver/pull/448 into this github action. Specifically,
1. Node.js versions and TSC versions are added as matrix instead of an old fixed version and latest version pair
2. We are considering removing unit-latest-deps. Quoting from Bret
> This feels like a pretty unusual situation; the user would have to explicitly override the versions specified in package-lock.json to get to something like this. If they do that they're testing with different dependencies than what we use so all bets are really off at that point.
Additionally, I could see something like this as an exercise during the development process, maybe something to check for newer versions of deps. But I don't think we want it running with all (or even some) of our CI builds do we? I can see a case for automating it (again as a means of detecting more recent versions of deps that we may want to upgrade to) but it's far from clear to me it belongs here.

3. For TSC 4.9 and 6.0 to work
a. I had to upgrade `@types/node` from 18.x to 20.x. `@types/node` is supposed to align with the Node.js version. As the oldest Node.js version we support is 20.x, we should upgrade `@types/node` from 18.x to 20.x.
b. TSC 6.0 has some breaking changes. 
1️⃣ is that we have to explicitly specify `"types": ["node"],` in the tsconfig.json ([Ref](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/#types-now-defaults-to-[])). 
2️⃣  is that running `tsc` specifying a file when a `tsconfig.json` is present but ignored is considered confusing, so`tsc generated.ts` will fail, we have to use `npx tsc generated.ts --ignoreConfig --types node` instead.([Ref](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/#specifying-command-line-files-when-tsconfig.json-exists-is-now-an-error)).
3️⃣ TSC 6.0 transpiles `import * as root from "../../../index";` differently, emitting an additional `root.default`, but also does not recognize `root.default` as a property, so it fails. We had to except `default` when printing objects from `root` now.

I'd really appreciate it if @toptobes can help me review whether these TSC 6.0 changes make sense.

toptobes:
@SiyaoIsHiding Regarding TSC 6.0, there shouldn't be any issue with upgrading `@types/node` and setting `"types": ["node"]` specifically. I do have a few things to say regarding points `3.2️⃣` and `3.3️⃣` though:

So firstly you are correct about needing to use `--ignoreConfig` if you're trying to transpile a specific file. However, I think you're overthinking slightly 🙂. You can just run `npx tsc` yet again which will transpile the newly generated `generated.ts` amongst the other TS files. 
- Yes, it'll take an extra maybe .5 seconds, but it saves you needing to do a brittle version check and use different options for each, and it especially saves you from having to restate options that are otherwise in the tsconfig, again in the command line (such as `--types node`)

Basically, the action can be updated to just be this:
```sh
cd test/unit/typescript/
npx tsc
node -e "require('./api-generation-test').generate()" > generated.ts
npx tsc --noEmit
```
- This works on both `4.9` and the latest `6.0`–I checked
- Also, thanks for teaching me about `pushd`! Very useful

However simplifying the action to the above requires on following the solution to the next topic, which is related to why `root.default` is a thing, and how it can be fixed.

The reason for your new `.default` friend is because `esModuleInterop` and `allowSyntheticDefaultImports` are now always `true` and you aren't even allowed to disable them without explicitly setting `"ignoreDeprecations": "6.0"` in your tsconfig (see [[1](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#combining---moduleresolution-bundler-with---module-commonjs)] [[2](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#combining---moduleresolution-bundler-with---module-commonjs)])

To fix the issue, just use `import root from "..."` instead and to compensate for the older versions, explicitly set `"allowSyntheticDefaultImports": true` and `"esModuleInterop": true` in your tsconfig.

Then, you won't even need to exclude `"default"` in `api-generation-test.ts`, because it won't be generated!

.github/workflows/test.yml
```diff
@@ -0,0 +1,222 @@
...
+  contents: read
+  actions: read
+  checks: write
+jobs:
+  eslint:
+    name: Run eslint
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        NODE_VERSION: ["20", "22", "24"]
```
toptobes: I don't think the node version will change eslint's output since it's strictly static analysis. Correct me if I'm wrong though

.github/workflows/test.yml
```diff
@@ -0,0 +1,222 @@
...
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.NODE_VERSION }}
+          cache: "npm"
+      - run: npm install -g eslint@4
```
toptobes, line 21: This is a very old version of eslint. Would there be any issues with using a newer version?

.github/workflows/test.yml
```diff
@@ -0,0 +1,222 @@
...
+        TSC_VERSION: ["4.9", "5.9", "6.0"]
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.NODE_VERSION }}
+          cache: "npm"
+
+      - run: npm install
```
toptobes: Generally `npm ci` is better than `npm install` here since it'll follwo the package-lock.json strictly rather than installing whatever's the latest from package.json

toptobes (approved):

.github/workflows/test.yml
SiyaoIsHiding: A newer version eslint will need quite some changes, as I remember from last time I tried. At least, we need to migrate to the new config file format. I don't think the efforts would be worth it now cuz we will move to TypeScript soon anyway

PR #456, opened 2026-04-09
CASSNODEJS-4: Update DRIVER_NAME after donation to ASF

SiyaoIsHiding:


jorgebay (approved):

absurdfarce:
I'm very good with the change code-wise but I think we also need a mention in the upgrade_guide to call this out to users.  Most people won't care but if somebody _was_ monitoring driver usage by looking at this string we should do what we can to make them aware that it's changed.

@jorgebay can you think of anywhere else we'd need to make mention of this _besides_ the upgrade guide?

package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "cassandra-driver",
   "version": "4.8.0",
-  "description": "DataStax Node.js Driver for Apache Cassandra",
-  "author": "DataStax",
+  "description": "Apache Cassandra Node.js Driver",
+  "author": "Apache Software Foundation",
```
absurdfarce, line 5: Is there any additional change to the "author" field here?  I agree that the ASF is a better value to have here... I just want to make sure we understand the full scope of making this change.

Does npm take any different action from what it was doing if this field is changed for instance?

jorgebay: Changing description or author has no effect on how the package works or is installed. These fields are purely metadata.
It's mainly for `npm search` and whatnot.

absurdfarce:
Okay, given @jorgebay's comments on the "author" metadata I think the only thing we need to wrap this up is a mention in the upgrade guide.  An example of what this looked like for the Python driver can be found [here]([‎docs/upgrading.rst‎](https://github.com/apache/cassandra-python-driver/pull/1284/changes#diff-b54c4c43417ac3174e0966e6ab5f8fa452b121e1e0fcb50534a43a4dc86df9e6)).

SiyaoIsHiding:
Added to the upgrade guide!

jorgebay (approved):
:shipit:

absurdfarce (approved):

PR #458, opened 2026-04-23
Changelog and version bump for 4.9.0

SiyaoIsHiding:


CHANGELOG.md
```diff
@@ -1,5 +1,30 @@
...
+### Bug fixes
+- [PR #432]  - Fix retry on socket error
+- [NODEJS-693] - Remove broken jsdoc from 4.8.0
+- [NODEJS-691] - Fix generated timestamp on retry
+
+### Dependency updates
+
+- Bump tar-fs from 2.1.2 to 2.1.4
+- Bump serialize-javascript and mocha (from 10.2.0 to 10.8.2)
+- Bump picomatch from 2.3.1 to 2.3.2
```
absurdfarce: I don't know that we need to call out dependency updates on a changelog.  They're kinda like test updates/infrastructure changes... sort of a built-in cost of doing business.

SiyaoIsHiding: Removed

package.json
```diff
@@ -54,4 +54,4 @@
   "engines": {
     "node": ">=20"
```
absurdfarce, line 55: So, this is a little bit of a weird one.

node.js 20.x is [scheduled to go end-of-life](https://github.com/nodejs/release#release-schedule) on 4/30/26.  As of this writing that's in the future but by the time we get the release staged, voted on and actually done it will _very likely_ be after 4/30.

So I guess I'm wondering... should we continue support 20.x?  By a _strict_ understanding of our "only support what's not EOL" policy we shouldn't.  That said we've already tested against 20.x... and it'll only _just_ be EOL by the time we release it.

I think we're prolly good to keep this as >= 20... but it seemed worth raising as a question.

SiyaoIsHiding: Nobody says we shouldn't support an EOL version. We had been supporting Node.js 18 which has been EOL for a long time. Java driver also has been supporting JAVA 8 forever. I don't think it's a problem that we support Node.js 20.

jorgebay: I'm +1 on supporting EOL versions for some time. It takes some time for users to upgrade :)

absurdfarce: We can discuss more but in general the drivers only support platforms/runtime that aren't EOL at the time of release.  We _don't_ do that for the Java driver but honestly that's a problem with the Java driver... we should be _way_ better about that there.

absurdfarce (approved):

PR #459, opened 2026-04-24
CASSNODEJS-6 Fix flaky client-batch-tests of cassandra.pid missing

SiyaoIsHiding:
Example all green run: https://github.com/SiyaoIsHiding/nodejs-driver/actions/runs/24861992383
I'm crying I feel like I've never seen all green run like that for years


jorgebay (approved):
Nice!

PR #462, opened 2026-05-20
NODEJS-681: ControlConnection Concurrent Read and Write on .host and .connection

toptobes:
This PR superceeds Jane He's #430 with her blessing

After some investigation, we were unable to figure out the root cause behind the NPEs, with there being multiple potential avenues where the issue may have originated from, and so we decided to fix the issue at the lowest and simplest level we could–simply adding a stronger concurrency control to `_refresh` directly via a `_refreshInProgress` flag

I personally believe the issue stemmed from `_setHealthListeners` being called multiple times on the same host/connection, causing the listeners to trigger refreshes multiple times for the same event, leading to the NPEs mentioned in the ticket. 

However the issue is quite hard to organically reproduce so the theory remains a theory.

<details>
<summary>Potential trace</summary>
<ol>
<li> _refresh() is called
<li>  _refresh() calls _refreshControlConnection()
<li> _refreshControlConnection() fails to borrow a connection so it calls _initializeConnection()
<li> _initializeConnection() calls _setHealthListeners() 
<li> _refresh() gets back in control and then also calls _setHealthListeners()
</ol>

which means that there's the potential of, sequentially:

<ol>
<li> A new host and connection being set (call them H1 and C1)
<li> Listeners being attached to the H1 and C1
<li> A newer host being set (call it H2)
<li> Listeners being attached to the H2 and C1 without the previous listeners being removed
</ol>
</details>

### DIFF
--- a/lib/control-connection.js
+++ b/lib/control-connection.js
@@ -454,7 +454,7 @@ class ControlConnection extends events.EventEmitter {
     this._refreshInProgress = true;
 
     try {
-      return await this._unsafeDoRefresh(hostIterator)
+      return await this._unsafeDoRefresh(hostIterator);
     } finally {
       this._refreshInProgress = false;
     }
--- a/lib/control-connection.js
+++ b/lib/control-connection.js
@@ -87,6 +87,7 @@ class ControlConnection extends events.EventEmitter {
     this._addressTranslator = this.options.policies.addressResolution;
     this._reconnectionPolicy = this.options.policies.reconnection;
     this._reconnectionSchedule = this._reconnectionPolicy.newSchedule();
+    this._refreshInProgress = false;
     this._isShuttingDown = false;
 
     // Reference to the encoder of the last valid connection
@@ -216,7 +217,6 @@ class ControlConnection extends events.EventEmitter {
 
   _setHealthListeners(host, connection) {
     const self = this;
-    let wasRefreshCalled = 0;
 
     function removeListeners() {
       host.removeListener('down', downOrIgnoredHandler);
@@ -225,11 +225,6 @@ class ControlConnection extends events.EventEmitter {
     }
 
     function startReconnecting(hostDown) {
-      if (wasRefreshCalled++ !== 0) {
-        // Prevent multiple calls to reconnect
-        return;
-      }
-
       removeListeners();
 
       if (self._isShuttingDown) {
@@ -446,13 +441,31 @@ class ControlConnection extends events.EventEmitter {
   }
 
   /**
-   * Acquires a new connection and refreshes topology and keyspace metadata.
+   * Acquires a new connection and refreshes topology and keyspace metadata, with protection against concurrent refreshes.
    * <p>When it fails obtaining a connection and there aren't any more hosts, it schedules reconnection.</p>
    * <p>When it fails obtaining the metadata, it marks connection and/or host unusable and retries using the same
    * iterator from query plan / host list</p>
    * @param {Iterator<Host>} [hostIterator]
    */
   async _refresh(hostIterator) {
+    if (this._refreshInProgress) {
+      return;
+    }
+    this._refreshInProgress = true;
+
+    try {
+      return await this._unsafeDoRefresh(hostIterator)
+    } finally {
+      this._refreshInProgress = false;
+    }
+  }
+
+  /**
+   * The actual implementation of the refresh logic, without protection against concurrent executions.
+   * <p>Should only be used via _refresh.</p>
+   * @param {Iterator<Host>} [hostIterator]
+   */
+  async _unsafeDoRefresh(hostIterator) {
     if (this._isShuttingDown) {
       this.log('info', 'The ControlConnection will not be refreshed as the Client is being shutdown');
       return;
@@ -499,7 +512,7 @@ class ControlConnection extends events.EventEmitter {
       }
 
       // Retry the whole thing with the same query plan
-      return await this._refresh(hostIterator);
+      return await this._unsafeDoRefresh(hostIterator);
     }
 
     this._reconnectionSchedule = this._reconnectionPolicy.newSchedule();
--- a/lib/control-connection.js
+++ b/lib/control-connection.js
@@ -87,7 +87,7 @@ class ControlConnection extends events.EventEmitter {
     this._addressTranslator = this.options.policies.addressResolution;
     this._reconnectionPolicy = this.options.policies.reconnection;
     this._reconnectionSchedule = this._reconnectionPolicy.newSchedule();
-    this._refreshInProgress = false;
+    this._refreshPromise = null;
     this._isShuttingDown = false;
 
     // Reference to the encoder of the last valid connection
@@ -227,8 +227,9 @@ class ControlConnection extends events.EventEmitter {
     function startReconnecting(hostDown) {
       removeListeners();
 
+      // Don't attempt to reconnect when the ControlConnection is being shutdown
       if (self._isShuttingDown) {
-        // Don't attempt to reconnect when the ControlConnection is being shutdown
+        this.log('info', 'The ControlConnection will not be refreshed as the Client is being shutdown');
         return;
       }
 
@@ -407,15 +408,13 @@ class ControlConnection extends events.EventEmitter {
       // To acquire metadata we need to specify the cassandra version
       this.metadata.setCassandraVersion(this.host.getCassandraVersion());
       this.metadata.buildTokens(this.hosts);
-
-      if (!this.options.isMetadataSyncEnabled) {
-        this.metadata.initialized = true;
-        return;
-      }
-
-      await this.metadata.refreshKeyspacesInternal(false);
-      this.metadata.initialized = true;
     }
+
+    if (isReconnecting && this.options.isMetadataSyncEnabled) {
+      await this.metadata.refreshKeyspacesInternal(false);
+    }
+
+    this.metadata.initialized = true;
   }
 
   async _refreshControlConnection(hostIterator) {
@@ -448,15 +447,16 @@ class ControlConnection extends events.EventEmitter {
    * @param {Iterator<Host>} [hostIterator]
    */
   async _refresh(hostIterator) {
-    if (this._refreshInProgress) {
-      return;
+    if (this._refreshPromise) {
+      return await this._refreshPromise;
     }
-    this._refreshInProgress = true;
+
+    this._refreshPromise = this._unsafeDoRefresh(hostIterator);
 
     try {
-      return await this._unsafeDoRefresh(hostIterator);
+      return await this._refreshPromise;
     } finally {
-      this._refreshInProgress = false;
+      this._refreshPromise = null;
     }
   }
 
--- a/lib/promise-utils.js
+++ b/lib/promise-utils.js
@@ -149,7 +149,7 @@ function times(count, limit, fn) {
  * @returns {undefined}
  */
 function toBackground(promise) {
-  promise.catch(() => {});
+  promise?.catch(() => {});
 }
 
 /**
--- a/lib/promise-utils.js
+++ b/lib/promise-utils.js
@@ -145,13 +145,11 @@ function times(count, limit, fn) {
 
 /**
  * Deals with unexpected rejections in order to avoid the unhandled promise rejection warning or failure.
- * @param {Promise | undefined} promise
+ * @param {Promise} promise
  * @returns {undefined}
  */
 function toBackground(promise) {
-  if (promise) {
-    promise.catch(() => {});
-  }
+  void promise.catch(() => {});
 }
 
 /**
--- a/lib/promise-utils.js
+++ b/lib/promise-utils.js
@@ -145,11 +145,13 @@ function times(count, limit, fn) {
 
 /**
  * Deals with unexpected rejections in order to avoid the unhandled promise rejection warning or failure.
- * @param {Promise} promise
+ * @param {Promise | undefined} promise
  * @returns {undefined}
  */
 function toBackground(promise) {
-  promise?.catch(() => {});
+  if (promise) {
+    promise.catch(() => {});
+  }
 }
 
 /**
--- a/lib/promise-utils.js
+++ b/lib/promise-utils.js
@@ -149,7 +149,7 @@ function times(count, limit, fn) {
  * @returns {undefined}
  */
 function toBackground(promise) {
-  void promise.catch(() => {});
+  promise.catch(() => {});
 }
 
 /**
--- a/test/integration/short/control-connection-tests.js
+++ b/test/integration/short/control-connection-tests.js
@@ -153,6 +153,25 @@ describe('ControlConnection', function () {
       assert.strictEqual(cc.hosts.length, 1);
     });
 
+    it('should not break when refreshing concurrently', async () => {
+      const cc = newInstance();
+      cc.options.policies.loadBalancing = new policies.loadBalancing.RoundRobinPolicy();
+      disposeAfter(cc);
+
+      await cc.init();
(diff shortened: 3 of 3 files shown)


lib/control-connection.js
```diff
@@ -225,11 +225,6 @@ class ControlConnection extends events.EventEmitter {
...
 
     function startReconnecting(hostDown) {
-      if (wasRefreshCalled++ !== 0) {
-        // Prevent multiple calls to reconnect
-        return;
-      }
-
       removeListeners();
 
       if (self._isShuttingDown) {
```
toptobes, line 231: this conditional in `_refresh()` logs if true... should it also log here or no?

lib/promise-utils.js
```diff
@@ -149,7 +149,7 @@ function times(count, limit, fn) {
  * @returns {undefined}
  */
 function toBackground(promise) {
-  promise.catch(() => {});
+  promise?.catch(() => {});
```
toptobes: not sure if it's this package's responsibility to deal with the fact that the promise can be null 🤷

test/integration/short/control-connection-tests.js
```diff
@@ -153,6 +153,25 @@ describe('ControlConnection', function () {
       assert.strictEqual(cc.hosts.length, 1);
     });
 
+    it('should not break when refreshing concurrently', async () => {
```
toptobes, line 156: may need a better heuristic for ensuring the refreshes are okay... not sure... I just "borrowed" this from Jane's original PR

lib/control-connection.js
```diff
@@ -446,13 +441,31 @@ class ControlConnection extends events.EventEmitter {
...
-   * Acquires a new connection and refreshes topology and keyspace metadata.
+   * Acquires a new connection and refreshes topology and keyspace metadata, with protection against concurrent refreshes.
    * <p>When it fails obtaining a connection and there aren't any more hosts, it schedules reconnection.</p>
    * <p>When it fails obtaining the metadata, it marks connection and/or host unusable and retries using the same
    * iterator from query plan / host list</p>
    * @param {Iterator<Host>} [hostIterator]
    */
   async _refresh(hostIterator) {
+    if (this._refreshInProgress) {
+      return;
```
toptobes: returning an auto-resolving promise here would defer the _refresh call to the microtask queue which may or may not open a hole for a race condition–I haven't checked.

I don't think it's necessary to return one, but if we do, it's definitely something that needs to be done with at least a little caution

lib/control-connection.js
```diff
@@ -446,13 +441,31 @@ class ControlConnection extends events.EventEmitter {
...
+    }
+    this._refreshInProgress = true;
+
+    try {
+      return await this._unsafeDoRefresh(hostIterator)
+    } finally {
+      this._refreshInProgress = false;
+    }
```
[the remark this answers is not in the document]
toptobes: `_refresh()` is never actually awaited

test/integration/short/control-connection-tests.js
```diff
@@ -153,6 +153,25 @@ describe('ControlConnection', function () {
...
+      await cc.init();
+      await new Promise(r => cc.options.policies.loadBalancing.init(null, cc.hosts, r));
+
+      const refreshPromises = [];
+      // randomly emit cc._refresh 100 times
+      for (let i = 0; i < 100; i++) {
+        refreshPromises.push(cc._refresh());
+        await helper.delayAsync(~~(Math.random() * 100));
+      }
+      await Promise.all(refreshPromises);
```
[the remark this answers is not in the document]
SiyaoIsHiding: Copilot's suggestion won't work. If the `_refresh` call all starts around the same time and ends around the same time, the null pointer exception won't actually happen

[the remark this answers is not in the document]
SiyaoIsHiding: And although my test can be unreliable, false test failures won't happen.
I agree that this test isn't ideal tho, open to ideas on how to improve it

SiyaoIsHiding (changes requested):
I think, the main reason why it broke in the past, is that `this.host` and `this.connection` are a nullable type,  (they are **supposed to be null** in certain circumstainces like refreshing), but function calls like `this._setHealthListeners(this.host, this.connection);` are not treating them as nullable. Imagine if this is TypeScript, it would complain that `_setHealthListeners` expects `Host, Connection` as argument while this call passes `Host?, Connection?`.
So, aside from the `_refreshInProgress` flag to make sure only one refresh is happening at one time, I think we should still make sure we are using `this.host` and `this.connection` as nullable. 
That means `    this._setHealthListeners(this.host, this.connection);` should be 
```
if (this.host && this.connection){
this._setHealthListeners(this.host, this.connection);
}
```
And other places that access `this.host` and `this.connection` should also has null guards, like [this](https://github.com/toptobes/cassandra-nodejs-driver/blob/37e314ff8d9d9f0974d2185e9d6b43f1349447e4/lib/control-connection.js#L408).
What do you think?

toptobes:
> I think, the main reason why it broke in the past, is that `this.host` and `this.connection` are a nullable type, (they are **supposed to be null** in certain circumstainces like refreshing), but function calls like `this._setHealthListeners(this.host, this.connection);` are not treating them as nullable. Imagine if this is TypeScript, it would complain that `_setHealthListeners` expects `Host, Connection` as argument while this call passes `Host?, Connection?`. So, aside from the `_refreshInProgress` flag to make sure only one refresh is happening at one time, I think we should still make sure we are using `this.host` and `this.connection` as nullable. That means ` this._setHealthListeners(this.host, this.connection);` should be
> 
> ```
> if (this.host && this.connection){
> this._setHealthListeners(this.host, this.connection);
> }
> ```
> 
> And other places that access `this.host` and `this.connection` should also has null guards, like [this](https://github.com/toptobes/cassandra-nodejs-driver/blob/37e314ff8d9d9f0974d2185e9d6b43f1349447e4/lib/control-connection.js#L408). What do you think?

As far as I can tell, the nullability is meant to be a transient state and trying to impose these silent null checks may just cause more subtle bugs than it solves? I wouldn't be against explicit null checks in `_setHealthListeners` as an assertion, but simply ignoring setting the listeners could be problematic

in an ideal world we may want to use an explicit state machine but that seems pretty out of scope for this PR

SiyaoIsHiding:
Why
> silent null checks may just cause more subtle bugs

?
I think they can potentially fix some bugs that we didn't discover yet.
But either way, I'm good. I think this PR can already fix the problem that we clearly know of.
If you fix that eslint error, I will give the explicit approval.

toptobes:
> Why
> 
> > silent null checks may just cause more subtle bugs
> 
> ? I think they can potentially fix some bugs that we didn't discover yet. But either way, I'm good. I think this PR can already fix the problem that we clearly know of. If you fix that eslint error, I will give the explicit approval.

Just to step back for a second we need to decide if it's even valid to get to the end of `_refresh()` without the host or the connection being set

I don't think it should be, no?

Meaning if it happens, we're just enabling the bug to propagate silently instead of catching it and yelling to the user that that's happened and needs to be patched (which it really should never happen with this refresh fix or there's a bigger issue)

SiyaoIsHiding:
No it shouldn't reach the end of _refresh without `this.host` set.
You convinced me 👍

SiyaoIsHiding:
[This](https://github.com/apache/cassandra-nodejs-driver/actions/runs/26197967397/job/77124427064?pr=462) test failure is new and it concerns me.

```
  1) ControlConnection
       #init()
         should subscribe to SCHEMA_CHANGE events and refresh keyspace information:
     Error: Condition still false after 100 attempts: () => cc.metadata.keyspaces['sample_change_1'].strategyOptions.replication_factor === '2'
      at whilstItem (test/test-helper.js:769:23)
      at Timeout.next [as _onTimeout] (lib/utils.js:1042:5)
      at listOnTimeout (node:internal/timers:585:17)
      at process.processTimers (node:internal/timers:521:7)
```
After some investigation, I think the problem is that a schema refresh triggered by an event can be permanently lost if the CC's connection is down at that moment and cannot establish new connection within 2 seconds. And this singleton refresh implementation makes it more vulnerable than before.

**What might have happened:**

In CI, a brief TCP hiccup (or Cassandra-side connection reset) closed the CC's socket during the 100ms debounce window after the ALTER KEYSPACE event:

1. ALTER KEYSPACE fires → SCHEMA_CHANGE event → EventDebouncer starts 100ms timer
2. TCP socket closes → `socketClose` →` _refresh()` acquires `_refreshInProgress`, sets `this.connection = null`
3. Debouncer fires → `metadata.refreshKeyspace` → `cc.query()` → `connection === null` →` _waitForReconnection()` (2s timeout)
4. `_refresh()` fails or does not establish new connection within 2 seconds -> `_waitForReconnection` rejects
5. Error is swallowed by toBackground() — the ALTER KEYSPACE schema update is silently dropped
6. CC eventually reconnects but no new SCHEMA_CHANGE event arrives → keyspace stays at replication_factor=3 → poll times out

In the past, when we accidentally allowed concurrent `_refresh()` calls, If one concurrent attempt succeeded while another failed, the `newConnection(null)` event would resolve the pending `_waitForReconnection` before the error could reject it. The new singleton approach eliminates that accidental rescue path.

**Fix**

Instead of allowing concurrent refresh, I think we should fix the problem of schema refresh errors being swallowed. For example, in `control-connection.js` `_nodeSchemaChangeHandler`
```js
  // Instead of: toBackground(this.handleSchemaChange(event, false))
  this.handleSchemaChange(event, false).catch(() => {
    // CC will reconnect; re-queue this event once it does
    this.once('newConnection', (err) => {
      if (!err) toBackground(this.handleSchemaChange(event, false));
    });
  });
```

SiyaoIsHiding (changes requested):
Above

SiyaoIsHiding:
Bret points out that the Java driver's behavior is that every time a new control connection is established, it refreshes schema 

https://github.com/apache/cassandra-java-driver/blob/90b09e0e34fd6fdda064f13f6acd0df7268a3dc6/core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java#L460-L469

It makes sense to me that Node.js driver can align with Java driver's behavior.
That would be an easier fix, too.
```
diff --git a/lib/control-connection.js b/lib/control-connection.js
index 5abf8b06..4199281c 100644
--- a/lib/control-connection.js
+++ b/lib/control-connection.js
@@ -415,6 +415,8 @@ class ControlConnection extends events.EventEmitter {
 
       await this.metadata.refreshKeyspacesInternal(false);
       this.metadata.initialized = true;
+    } else if (this.options.isMetadataSyncEnabled) {
+    await this.metadata.refreshKeyspacesInternal(false);
     }
   }
```

toptobes:
After some deliberation we think we can just call `this.metadata.refreshKeyspaces[Internal?]()` within `_refreshHosts` unconditionally as that function appears to invalidate all such schema caches. Need to triple check if this is the right solution thoguh

SiyaoIsHiding:
Confirming above is correct.
The question is whether `refreshKeyspacesInternal` is enough to cover any EVENTs potentially lost.
There are 3 kinds of EVENTs
1. TOPOLOGY_CHANGE, handled by `refreshHosts`
2. STATUS_CHANGE, a host UP or DOWN, is handled by `host.js`'s reconnection schedule, independent from `ControlConnection`
3. SCHEMA_CHANGE
  3a. `refreshKeyspacesInternal` redo `SELECT * FROM system_schema.keyspaces`
  3b. It also removes the cache for udt, table, function, aggregate, and views. They will be lazy-loaded when used for the first time.
So I think `refreshKeyspacesInternal` is enough.

lib/promise-utils.js
SiyaoIsHiding: This is an internal util not exposed to the users.
But why allow the argument to be null tho? Currently all usages in this package are passing non-null promises. Sounds like allowing it to be null can make bugs harder to notice.

lib/control-connection.js
SiyaoIsHiding: Why not 
```js
    if (this._refreshInProgress) {
      return this._refreshPromise;
    }
```
It makes more sense because developers would expect that after the execution of `await _refresh()`, the refresh should actually be complete.

lib/control-connection.js
SiyaoIsHiding: Makes sense, but there isn't a user demanding such log. Either makes sense to me.

lib/control-connection.js
toptobes: because no one ever actually awaits `_refresh` (I guess except for the `should not break when refreshing concurrently` test which probably should be refactored to not rely on it being given a promise`)

technically this shouldn't even be async in the first place since it's just made to be fire-and-forget anyways

toptobes: wait actually nvm we still need the `async/await` for the `finally` to work properly

SiyaoIsHiding: Although all existing usages are fire-and-forget, I for sure want `await _refresh()` to actually finish refresh unless there is a strong reason not to do so and clearly warned in JSdoc, otherwise some unfortunate driver developer in the future will have a huge headache debugging some weird failure and yanking their hair out.

lib/control-connection.js
```diff
@@ -225,15 +225,11 @@ class ControlConnection extends events.EventEmitter {
...
-        // Prevent multiple calls to reconnect
-        return;
-      }
-
       removeListeners();
 
+      // Don't attempt to reconnect when the ControlConnection is being shutdown
       if (self._isShuttingDown) {
-        // Don't attempt to reconnect when the ControlConnection is being shutdown
+        this.log('info', 'The ControlConnection will not be refreshed as the Client is being shutdown');
```
SiyaoIsHiding, line 232:
```suggestion
        self.log('info', 'The ControlConnection will not be refreshed as the Client is being shutdown');
```

lib/control-connection.js
```diff
@@ -412,15 +408,13 @@ class ControlConnection extends events.EventEmitter {
...
-        this.metadata.initialized = true;
-        return;
-      }
-
+    if (isReconnecting && this.options.isMetadataSyncEnabled) {
       await this.metadata.refreshKeyspacesInternal(false);
-      this.metadata.initialized = true;
     }
+
+    this.metadata.initialized = true;
```
SiyaoIsHiding, line 417: nit: this will have the side effect that if metadata sync is not enabled, `this.metadata.initialized = true;` will still run everytime it refreshes, not just once at initialization.

SiyaoIsHiding (commented):
This test still fails
```
1) ControlConnection
       #init()
         should subscribe to SCHEMA_CHANGE events and refresh keyspace information:
     Error: Condition still false after 100 attempts: () => !cc.metadata.keyspaces['sample_change_1']
      at whilstItem (test/test-helper.js:769:23)
      at Timeout.next [as _onTimeout] (lib/utils.js:1042:5)
      at listOnTimeout (node:internal/timers:605:17)
      at process.processTimers (node:internal/timers:541:7)
```
I think it's possible that 
1. When the driver is selecting keyspace `sample_change_1` info, that keyspace still existed in the DB, so DB sends the response of keyspace info, but it does not reach the driver yet.
2. The DROP statement executes.
3. Driver `delete metadata.keyspaces['sample_change_1']`
4. The `sample_change_1` keyspace info finally reaches the driver
5. Driver assigns `cc.metadata.keyspaces['sample_change_1']`

I need to think about how to fix this..... But I think this path is possible to fail this test.

lib/control-connection.js
toptobes: I don't think it matters, no? As far as I can tell, it's idempotent (once set to true, it's never reset to false), no? I'm sure the user can spare an extra couple CPU cycles 🙂

SiyaoIsHiding:
My above theory of how that test fails does not hold.... need to figure out other reason why it can fail

PR #471, opened 2026-07-28
CASSNODEJS-14: Multiple Node Status Updates - All but One being Dropped

toptobes:
Updates `ControlConnection._nodeStatusChangeTimeout` to be a map of hosts => timers instead so that if multiple nodes had status updates at the same time, it wouldn't lead to all but one being dropped/ignored, which should speed up recovery time in such cases now

The node status event handlers may look different in the future under an all-encompassing "events handling overhaul", but this is a pretty simple fix for the time being

### DIFF
--- a/lib/control-connection.js
+++ b/lib/control-connection.js
@@ -99,8 +99,8 @@ class ControlConnection extends events.EventEmitter {
 
     // Timeout used for delayed handling of topology changes
     this._topologyChangeTimeout = null;
-    // Timeout used for delayed handling of node status changes
-    this._nodeStatusChangeTimeout = null;
+    // Map of host endpoint strings to their pending status change timeouts
+    this._nodeStatusChangeTimers = new Map();
 
     if (context && context.borrowHostConnection) {
       this._borrowHostConnection = context.borrowHostConnection;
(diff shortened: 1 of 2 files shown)


lib/control-connection.js
```diff
@@ -598,14 +598,23 @@ class ControlConnection extends events.EventEmitter {
...
       const distance = self._profileManager.getDistance(host);
       if (event.up) {
         if (distance === types.distance.ignored) {
           return host.setUp(true);
         }
-        clearTimeout(self._nodeStatusChangeTimeout);
         // Waits a couple of seconds before marking it as UP
-        self._nodeStatusChangeTimeout = setTimeout(() => host.checkIsUp(), newNodeDelay);
+        self._nodeStatusChangeTimers.set(endPoint, setTimeout(() => {
+          self._nodeStatusChangeTimers.delete(endPoint);
```
SiyaoIsHiding, line 615: nit: do we need this line here?
> self._nodeStatusChangeTimers.delete(endPoint);

SiyaoIsHiding (commented):
Code changes make sense me. Would be great if we have tests!

lib/control-connection.js
toptobes: I mean, technically no? But there's not much point keeping dead timer ids around even if they'll be overwritten anyway, it's mostly a minor memory thing.

jorgebay (approved):
Looks great! good fix @toptobes !

